PR #3, opened 2015-06-16
Update doc for subscribing to all events

anderslime:
Just an update to the documentation. Great project btw - nice, simple and then it's a pleasure to read the source code!


andrzejkrzywda:
thanks @anderslime :)

PR #5, opened 2015-07-03
use activesupport and fix apply event methods scope

### LINKED ISSUE #4: apply event method in aggregate should be private
Any reason why the `apply_events_*` methods are public? From what I understood of aggregate roots, we need to protect their internals from the outside world.

We should probably use `send` here and declare our `apply` methods private. What do you think?

https://github.com/arkency/rails_event_store/blob/master/lib/rails_event_store/aggregate_root.rb#L19

gottfrois:
Fixes #4


spec/aggregate_root_spec.rb
```diff
@@ -3,16 +3,18 @@
 class Order
   include RailsEventStore::AggregateRoot
 
+  attr_accessor :id, :status
```
mpraglowski: This should be also private. No state outside aggregate :) And this change should go in separate PR. Would you like to move it out from this PR ? I think new PR with this change could be a good starting point for a discussion how to do it better.

gottfrois: Those fields were already public. I think it was to be able to assert status value and infer that even was properly applied. While I agree that in real context they should be private, is it that important here?

mpraglowski: This will be described in a blog post which I plan to publish soon (need to finish sample code) - you should not check (or really care) about the aggregate internal state, the only thing you care about and what you should check is what events have been produced by calling aggregate's method (executing a command). The state of an aggregate is internal - it just an implementation detail, for some operations it could be no state representation at all - only new produced events).

mpraglowski: I will merge this PR and then I will fix tests :)

gottfrois: Right I agree with you. There is no clear way yet to assert event creation. Maybe the gem should provide a custom rspec validator? If we go that path though, we need to handle more than rspec...

mpraglowski: I've pushed 2 new commits - the tests are still not perfect - I will improve them after I finish my blog post.

PR #6, opened 2015-07-06
fix: apply repository pattern correctly by decoupling event entity from any ORM

gottfrois:
Adds an explicit dependency on `activemodel`. Basically my mid-term goal is to build the `mongoid` adapter so I figured I should first start by refactoring how the repository pattern was applied here.

Thoughts and comments are welcome


mpraglowski:
thanks @gottfrois :+1: :)

gottfrois:
We need stronger specs in the futur. Current specs looks fragile and don't assert the repository :-/

mpraglowski:
Yep - noticed that. Feel free to add your specs :) I will try to find some time to update them too.

PR #7, opened 2015-07-08
add mongoid as a repository backend

gottfrois:
Breaking change.

The goal is to be able to handle multiple backend (active_record, mongoid for now). The mongoid repository is working well but we should split each classes specific to each backend in their own gem.

- `rails_event_store` gem will hold the common code to deal with event sourcing.
- `rails_event_store_active_record` gem will hold active_record repository, active_record event model and active_record migration templates and task
- `rails_event_store_mongoid` gem will hold mongoid repository and event model

This PR is just to illustrate how it would work.

Include the gems in your rails gemfile:

```ruby
gem 'rails_event_store'
```

Under the hood, the `rails_event_store_mongoid` will do:

```ruby
RailsEventStore::Repository.backend = :mongoid
```

to register itself as the backend repository to use.

Today specs are failing because it's really hard to deal with dependencies while the code is not split.

Thoughts and feedbacks are welcome.


mpraglowski:
Hmm, why not the separate mongoid_event_store gem based on ruby_event_store gem (as it is now rails_event_store) ? You basically doubled the code in RES. Have you considered to create new gem instead? No Need to configure anything, no unnecessary dependencies, you just choose what gem you want to use :)

gottfrois:
Not sure to follow. `ruby_event_store` is supposed to be agnostic to any backend right?

gottfrois:
Creating a new gem is the way to go yes. I did not do it here because I think it's better to create the gem under the Arkency repo ;)

mpraglowski:
`rails_event_store` adds a AR based persistence to persistence agnostic `ruby_event_store` - so basically it is the same as `rails_event_store_active_record` in your description

`mongoid_event_store` could add a MongoDb based persistence to persistence agnostic `ruby_event_store` - so basically it is the same as `rails_event_store_mongoid` in your description

There is not much common code in `rails_event_store` to have it as common for AR & Mongo gems.

gottfrois:
Hum i see what you mean. I still see the following classes as common code though:

- `client.rb`
- `event_entity.rb`
- `aggregate_root.rb`
- `aggregate_repository.rb`

The way I saw this is that `rails_event_store` would be the core of an event store working on rails applications. But the persistance layer is a plug & play depending on your rails application requirements. 

However, choosing one persistance layer or the other does not change the way events are created and applied. As well as how the aggregate root behaves. But those features needs to be provided by the `rails_event_store` gem.

My fear is that as we want to add more feature to the event store gem, we'll need to duplicate those features between our different gems that are only supposed to deal with persistance. After all, persistance just needs to be handle through an interface in order to be flexible.

gottfrois:
By the way my PR was inspired by how DelayedJob gem handle that issue.

gottfrois:
Any news? :)

mpraglowski:
Sorry - holiday & offline mode - I will look t this today in the evening.

mpraglowski:
@gottfrois could you check why build is failing ?

lib/rails_event_store/client.rb
```diff
@@ -1,7 +1,7 @@
 module RailsEventStore
   class Client
 
-    def initialize(repository = Repositories::EventRepository.new)
+    def initialize(repository = ::RailsEventStore::Repository.backend)
```
mpraglowski: Why not just be explicit and let developer to specify repository here? 
Then we won't need global backend configuration and meta code in `lib/rails_event_store/repository.rb`.

gottfrois: sure it could work. More work for clients thought :)

mpraglowski: Less code to support :) And client's devs are not that lazy ;)

gottfrois:
Today specs are failing because it's really hard to deal with dependencies while the code is not split. Since we now require mongoid or active_record and load mongoid specific classes without the gem being loaded.

This PR is not mergeable as it is. We need to dispatch ORM dependent classes into their own gem

mpraglowski:
Ok, let's close this PR for now. Then we should start to work on divide RES gem and moving out some parts of it to separate gems. @gottfrois ok ?

gottfrois:
yep sounds good. We need to make sure RES does not deal with any ORM code first

array42:
How close is mongoid support?

pawelpacana:
@array42 https://github.com/gottfrois/rails_event_store_mongoid has the version compatible with RES < 0.15.0. 

I don't recall any further actitivity, in the meantime repository API changed a bit for a reason (https://github.com/RailsEventStore/rails_event_store/issues/137#issuecomment-334502061).

gottfrois:
@array42 we do use the fork in production system but to be honest it's not an ideal situation as we will need to catch up with origin at some point...

PR #8, opened 2015-07-10
refactor rails event store events

### LINKED ISSUE #2: Events data
Hey guys,

What are your thoughts regarding how event classes are defined? Today we put our event data under a `data` key. Unfortunately we then have to define some helper methods to access this data as method calls instead of hash attributes:

```ruby
class SomeEvent < RailsEventStore::Event  
  def uid
    data.fetch(:uid)
  end

  def foo
    data.fetch(:foo)
  end

  def self.create(uid, foo)
    new(data: { uid: uid, foo: foo })
  end
end
```

Maybe we could drop the `data` namespace and define our events using activemodel:

```ruby
class SomeEvent < RailsEventStore::Event  
  include ActiveModel::Model

  attr_accessor :uid, :foo

  def self.create(uid, foo)
    new(uid: uid, foo: foo)
  end
end
```

The `include` could then be part of `RailsEventStore::Event` class. The downside is with attributes conflicts is the one defined by you (metadata) but this seems a small issue compare to what we can win here. Also we could use `__` methods prefix or similar for event's special attributes.

We can push this even further and define the following method in the parent class to have back a `data` like method we can use in our projections:

```ruby
def data
  attributes_except :event_id, :event_type, :metadata
end

def attributes_except(*attrs)
  as_json.reject { |k, _| attrs.include?(k) }
end
```

```ruby
# from a projection
def fooed(event)
  MyModel.find_by_uid(event.uid).update!(event.data)
end
```

Would love your thoughts on this :)

gottfrois:
Fixes #2 

This is what I had in mind. We can probably refactor that to use the event class from RubyEventStore.

### DIFF
--- a/lib/rails_event_store/client.rb
+++ b/lib/rails_event_store/client.rb
@@ -1,3 +1,5 @@
+require 'ruby_event_store'
+
 module RailsEventStore
   class Client
 
--- a/lib/rails_event_store/event.rb
+++ b/lib/rails_event_store/event.rb
@@ -1,3 +1,50 @@
-require 'ruby_event_store'
+require 'securerandom'
 
-RailsEventStore::Event = RubyEventStore::Event
+module RailsEventStore
+  class Event
+    include ActiveModel::Model
+
+    attr_reader :metadata, :event_id, :event_type
+
+    def initialize(data = {})
+      super
+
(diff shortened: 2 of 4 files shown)


mpraglowski:
Resolved in #16

PR #9, opened 2015-07-21
fix: allow event entity to call methods on data

gottfrois:
Just found out that with the new implementation I proposed we now run into `NoMethodError` exceptions when re-playing the old events against the aggregate. Nevertheless the old solution, using the `OpenStruct`, was silently failing. At least we know it now :)


mpraglowski:
Resolved in #16

PR #13, opened 2015-11-05
Use supported Rails methods for finding all events

anderslemke:
Rails doesn't seem to support `find(:all)` anymore, so I propose this fix.

How would you prefer this to be tested?

### DIFF
--- a/lib/rails_event_store/repositories/event_repository.rb
+++ b/lib/rails_event_store/repositories/event_repository.rb
@@ -23,7 +23,7 @@ module RailsEventStore
       end
 
       def get_all_events
-        adapter.find(:all, order: 'stream').map &method(:build_event_entity)
+        adapter.order('id ASC').order('stream').all.map &method(:build_event_entity)
       end
 
       def last_stream_event(stream_name)


andrzejkrzywda:
Thanks!

As for tests - we don't have any test which would cover the existing EventRepository class. This is a potential problem - do you see what would be the best way of testing it?

We could fake the ActiveRecord subset with our own code and use this in tests (instead or additionally to the current approach with InMemoryRepo) - still not really testing against the real implementation, but at least would cover this code in repo.

We could also test against the real ActiveRecord. This would require using some kind of a database. This would be more heavy and slow, but it would let us against different versions of AR.

Any thoughts?

anderslemke:
That's a good question.
I actually do not know what best practice is in this scenario.

It would probably be nice to test against specific versions of ActiveRecord, and be able to say "compatible with Rails' ActiveRecord 3.2 and up" or something similar.

I do not think it would be a good approach to fake ActiveRecord in this case, as it would only test how the `EventRepository` behaves in a fake world, not the real world.

PR #14, opened 2015-12-01
Aggregate root handle concurrent writes

mpraglowski:


mpraglowski:
@gottfrois will this solve your concurrent writes issues?

gottfrois:
Hey thanks for this fix. I've been thinking about the optimistic locking strategy in my use case (2 threads trying to write on the event store at the same time) and I think it's not the right strategy no matter what.

Optimistic locking is a great strategy in the following example:

1. Alice loads the order and wants to edit it
2. Before submit, Alice goes for a cup of coffee
3. Bob comes into play and edit the order
4. Alice comes back and submit the form, causing a concurrency issue since the order has changed

Here optimistic locking solves the issue by checking the expected version (that you would have embedded in the form maybe?). It's a long live concurrency issue.

When dealing with multiple threads trying to store events on the same event store it's different. Those are short lived concurrency issues, that happens under milliseconds or less. I don't see any other way than locking the aggregate before inserting a new event. Which is called pessimistic locking in this case.

So at least, it's not a global lock on the event store, but a lock per aggregates.

Now, regarding this fix, I think it's great for the first example I gave. Now I see a potential issue on the first event (the creation of the order) since we check against `nil` value:

https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/actions/append_event_to_stream.rb#L19

For the first event, `expected_version` will always be `nil`. What's your thought on this?

mpraglowski:
> For the first event, expected_version will always be nil. What's your thought on this?

yep, but this is out of scope of `rails_event_store` gem, update of `ruby_event_store` is needed

> So at least, it's not a global lock on the event store, but a lock per aggregates.

The same here - this is also in `ruby_event_store` - lock on stream while writing an event (or events) should do the work.

gottfrois:
Yes of course, I'm just sharing thoughts here but it's essentially `ruby_event_store`'s issues :)

mpraglowski:
Not valid anymore - code moved to separate gem - see [aggregate_root](https://github.com/arkency/aggregate_root)

mpraglowski:
This PR has been moved to https://github.com/arkency/aggregate_root/pull/1

PR #16, opened 2016-01-27
Better event

mpraglowski:
After cleanup in RubyEventStore https://github.com/arkency/ruby_event_store/pull/3 also RailsEventStore have some cleanup. The respository always returns RailsEventStore::Event. Removed EventEntity. Also applied some Ruby magic to make life with events easier.

@gottfrois this PR should solve issues we discussed in PRs:
https://github.com/arkency/rails_event_store/pull/8 & https://github.com/arkency/rails_event_store/pull/9


gottfrois:
Great work out there!

I'm not sure it will solved [#9](https://github.com/arkency/rails_event_store/pull/9) though. You need a way to delegate methods to the `data` attribute.

Am I missing something here?

Here is what my `EventEntity` class looks like in my fork:

```ruby
require 'active_model'

module EventStoreMaster
  class EventEntity
    include ::ActiveModel::Model

    attr_accessor :id, :stream, :event_type, :event_id,
                  :metadata, :data, :created_at

    def data
      @_data ||= OpenStruct.new(@data)
    end

    def aggregate_id
      data.id
    end

    def method_missing(method_name, *args, &block)
      data.public_send(method_name, *args, &block)
    end

    def respond_to_missing?(method_name, include_private = false)
      data.respond_to?(method_name)
    end

  end
end
```

mpraglowski:
@gottfrois itr is all in RailsEventStore::Event

to create new event:
```ev = Events::OrderCreated.new(order_id: id, order_number: order_number, customer_id: customer_id)```

to access its attributes:
```
customer_id = ev.customer_id
```

See https://github.com/mpraglowski/cqrs-es-sample-with-res sample - it is already using it.

mpraglowski:
@gottfrois the "magic" is here https://github.com/arkency/rails_event_store/commit/67922ceb1d932d29aff378f7c8ee9e6884f02293#diff-ecace4f786e50886cbb0ebdf62ef5312R7

gottfrois:
Ah! I missed this line :) Then it looks great. What's up with the singleton logic here. Can't use `define_method` directly?

mpraglowski:
```define_method(*args) private
Defines an instance method in the receiver. The method parameter can be a Proc, a Method or an UnboundMethod object. If a block is specified, it is used as the method body. This block is evaluated using instance_eval, a point that is tricky to demonstrate because define_method is private. (This is why we resort to the send hack in this example.)```

define_method is private ;)

gottfrois:
:+1:

PR #19, opened 2016-02-23
Decouple data storage from RailsEventStore

gottfrois:
Now is a better time since aggregate_root has been decoupled already :)

ActiveRecord repository and data storage logic as a new gem. See details here https://github.com/gottfrois/rails_event_store_active_record/pull/1/files

We would need to update https://github.com/mpraglowski/cqrs-es-sample-with-res with the following:

```ruby
gem 'rails_event_store_active_record'
```

But if you want to be explcit, though it would work by default without it:
```ruby
RailsEventStore::Client.new(RailsEventStore::Repositories::ActiveRecord::EventRepository.new)
```

We could take one step further and actually move `RailsEventStore:: EventInMemoryRepository` from spec to lib and default to it as backend strategy.

@mpraglowski let me know what you think ;)


lib/rails_event_store/event.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_support/inflector'
```
gottfrois: had to add that for `Hash#except` to work on specs. Did I miss something?

mpraglowski:
> ActiveRecord repository and data storage logic as a new gem.

I like that idea.

> We could take one step further and actually move RailsEventStore:: EventInMemoryRepository from spec to lib and default to it as backend strategy.

I've plans to have it as a separate gem - kind of "contract" for rails_event_store & http_eventstore clients and to be used in tests of course.


But I would like to keep the "rails like" nature of RailsEventStore. You should just add `gem 'rails_event_store'` to app gemfile and it should enable all features - event sourcing with rails_event_store, aggregate_root etc. (that's why I've reverted dependencies between aggregate_root & rails_event_store in 0.2.2 version).

The idea is to have rails_event_store as an "umbrella" for all of it. You get it and you have all the defaults.
That's why I would like to not have that dependency here https://github.com/gottfrois/rails_event_store_active_record/pull/1/files#diff-26887acbf263a4b7f116c50b29500435R20

mpraglowski:
My current vision of how to setup dependencies is:

```
rails_event_store
|— ruby_event_store
|     |— in_memory_event_store
|— aggregate_root
|     |— in_memory_event_store
|— in_memory_event_store
|— rails_event_store_active_record
|— rails_event_store_XXX (where XXX could be anything what is able to persist domain events)

http_eventstore
|— in_memory_event_store
```

gottfrois:
I fully agree with revert dependency changes in 0.2.2. I felt adding `gem 'aggregate_root'` in the gemfile to add _event_store_ features was quite disturbing.

Now, about dependency hierarchy, the vision you have is already what I did here with backend flexibility.

Client's gemfile would look like this:

```ruby
gem 'rails_event_store'
gem 'rails_event_store_active_record
```

Explicit `rails_event_store` dependency declared [here](https://github.com/gottfrois/rails_event_store_active_record/pull/1/files#diff-26887acbf263a4b7f116c50b29500435R20) is basically to be allowed to access common `rails_event_store` code such as [error class](https://github.com/gottfrois/rails_event_store_active_record/pull/1/files#diff-ff2967f6412acadc506ce6ceffbce992R19) and [event entity](https://github.com/gottfrois/rails_event_store_active_record/pull/1/files#diff-ff2967f6412acadc506ce6ceffbce992R49)

```
rails_event_store
|— in_memory_event_store
|— rails_event_store_active_record
|— rails_event_store_XXX
```

All the above would set, at runtime, the proper datastore to use through the `RailsEventStore::Repository.backend` class variable. Or could be explicit pass to `RailsEventStore::Client.new`.

Finally, `in_memory_event_store` could be a seperate gem but I really don't think it's worth it. It would just be used for a sain default and by specs. There is no real life usage of a in-memory store.

mpraglowski:
> Explicit rails_event_store dependency declared here is basically to be allowed to access common rails_event_store code such as error class and event entity

This should be moved to ruby_event_store and then we would avoid this dependency.

rails_event_store should be a "omasake" for those group of gems - you should just add
```ruby
gem 'rails_event_store'
```

and all should be setup with defaults - with aggregate_root, rails_event_store_active_record and soon maybe some UI to browse domain events.

If you would like custom solution you should build add separate parts in gem file and setup them as you want, i.e. like this:

```ruby
gem 'ruby_event_store'
gem 'rails_event_store_XXX'
gem 'aggregate_root'
```

> There is no real life usage of a in-memory store.

For me it might be useful in tests of all other gems and as a "contract" - all other gems should behave like  in-memory one.

gottfrois:
I see. So we would load AR by default but if I want Mongoid I would just do this right?

```
gem 'rails_event_store'
```

mpraglowski:
You should be able to do this:

```ruby
gem 'ruby_event_store'
```

`rails_event_store` gem should be "the glue" to connect all gems with defaults, when you want your custom setup you should use `ruby_event_store` (but we are not ready for this yet).

gottfrois:
That makes no sense to me. Which datastore to use should just be a specialization (or an override) of `rails_event_store`.

It makes no sense from a client point of view that if you want to use mongoid datastore you have to include `gem 'ruby_event_store'` in your gemfile. It makes much more sense from a conceptual point of view to say:

1. Bring `rails_event_store` code by installing `gem 'rails_event_store'`
2. Override `rails_event_store` default datastore by using `gem 'rails_event_store_mongoid'`

No?

mpraglowski:
Yes, but then you will include more gems than required because rails_event_store will bring also rails_event_store_active_record, aggregate_root and maybe some more soon. But this should also be possible if you want to.

gottfrois:
> you will include more gems than required because rails_event_store will bring also rails_event_store_active_record, aggregate_root

yeah that's why i wanted to even have AR not part of RES and let the client do:

```ruby
gem 'rails_event_store'
gem 'rails_event_store_active_record'
```

Now, not sure to follow why you would not want `aggregate_root` not part of RES by default?

lib/rails_event_store/event.rb
mpraglowski: yeah, missed that. but anyway I'm moving this to ruby_event_store

expect new release soon

mpraglowski:
@gottfrois take a look at https://github.com/arkency/ruby_event_store/pull/7 - this might help with extracting AR repository out of RES

gottfrois:
:+1: I will :)

mpraglowski:
@gottfrois shall we work on this? I might have some time on Friday - would like to fork https://github.com/gottfrois/rails_event_store_active_record/pull/1/files apply latest changes make sure all tests pass and then remove AR event repository from RES gem.

gottfrois:
yep or i can change the ownership of the repo if you want

mpraglowski:
ok, let's do it like that

mpraglowski:
This is done in PR "Current Event Repository & all storage related code moved to RailsEventStoreActiveRecord"  #28

PR #21, opened 2016-02-24
Fix for issue 20 - missing stream

mpraglowski:


gottfrois:
I think you also need to update ruby_event_store event model since it's used their as well https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/actions/read_all_streams.rb#L10

We should probably initialize `stream` here as well https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/event.rb#L6

mpraglowski:
@gottfrois @hodak are you using grouping of events by stream as implemented here https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/actions/read_all_streams.rb#L10 ?

I do not remember a reason for this group by but I would prefer to remove it and also remove stream attribute from event. Instead of this I would like to add partial read of all events (similar to https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/facade.rb#L17 but reading from all streams). For me stream is just a unit of organisation for events - it does not belongs to Event. With projections (someday) we could have the same event in different streams.

gottfrois:
I'm not using grouping at the moment no. I would also delegate stream filtering to the event repository.

hodak:
@mpraglowski Me neither, I've just started with this gem and went with the documentation. If there's a better way then :+1:

mpraglowski:
Ok, I will remove that group & stream soon.

PR #25, opened 2016-03-15
Make RES compatible with the newest Ruby Event Store

mlomnicki:
The newest Ruby Event Store changed how event subscription work.
Currently event types are used rather than strings

currently:

```ruby
   es.subscribe(Handler, [OrderCreated])
```

before:

```ruby
   es.subscribe(Handler, ['OrderCreated'])
```

This commit makes RES work with the new approach.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gemspec
 gem 'codeclimate-test-reporter', group: :test, require: nil
+gem 'ruby_event_store', github: 'arkency/ruby_event_store'
--- a/lib/rails_event_store/repositories/event_repository.rb
+++ b/lib/rails_event_store/repositories/event_repository.rb
@@ -89,12 +89,13 @@ module RailsEventStore
 
       def build_event_entity(record)
         return nil unless record
-        ::RailsEventStore::Event.new(record.data.merge(
-          event_type: record.event_type,
-          event_id:   record.event_id,
-          metadata:   record.metadata))
+        record.event_type.constantize.new(
+          record.data.merge(
+            event_id: record.event_id,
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gemspec
 gem 'codeclimate-test-reporter', group: :test, require: nil
+gem 'ruby_event_store', github: 'arkency/ruby_event_store'
```
mlomnicki: of course I will move it back to the gemspec once the newest Ruby Event Store has been released to rubygems

lib/rails_event_store/repositories/event_repository.rb
```diff
@@ -89,12 +89,13 @@ def read_all_streams_backward(start_event_id, count)
 
       def build_event_entity(record)
         return nil unless record
-        ::RailsEventStore::Event.new(record.data.merge(
-          event_type: record.event_type,
-          event_id:   record.event_id,
-          metadata:   record.metadata))
+        record.event_type.constantize.new(
```
mpraglowski: I would prefer move this code to `ruby_event_store`. It should not be Repository who is responsible for recreating an event, here it should be something like:

```ruby
def build_event_entity(record)
   return nil unless record
   event_factory.call(record.data.merge(
     event_id: record.event_id,
     metadata: record.metadata
   ))
end
```

mlomnicki: 100% agree. I didn't want to do that in this PR though

mpraglowski: Sure - that's why it is merged now ;)

PR #31, opened 2016-03-31
Update Readme

mlomnicki:


MartinKoerner:
Hi @mlomnicki, I also changed the README, but I found, that the correct command is `rake rails_event_store:migrate` instead of `rake rails_event_store_active_record:migration`.
The command in your pull request didn't work for me...

mlomnicki:
Hey @MartinKoerner 

Yeah, you're right. The change has been introduced [here] (https://github.com/arkency/rails_event_store_active_record/pull/3) but the newest rails_event_store_active_record hasn't been released yet.
I think that you're using RES 0.5.0. The command will work in RES 0.5.1 (or 0.6.0)

cc @mpraglowski :)

PR #35, opened 2016-05-19
Adds write lock strategy per aggregates

gottfrois:
#### Overview

This changes allow to perform a database lock per aggregates in order to avoid concurrency issues in a concurrent environment, such as, multi-threaded env or load balanced event store (multiple event store processes trying to create events on the database at the same time).

The lock only blocks the aggregate, not the entire event table. A `RubyEventStore::CannotObtainLock` exception is raised in case the lock could not be obtained. Meaning an event store was already trying to perform an `INSERT` on the given aggregate.

Client's implementation should catch this exception and perform a simple `retry` in a `begin`/`rescue` block:

``` ruby
aggregate_id = SecureRandom.uuid
client = RailsEventStore::Client.new(lock_obtainer: RailsEventStoreActiveRecord::LockObtainer.new)

begin
  event = OrderCreated.new(data: "sample", event_id: SecureRandom.uuid)
  client.publish_event(event, aggregate_id)
rescue RubyEventStore::CannotObtainLock
  retry
end
```

### DIFF
--- a/lib/rails_event_store/client.rb
+++ b/lib/rails_event_store/client.rb
@@ -3,10 +3,12 @@ module RailsEventStore
 
     def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
                    event_broker: EventBroker.new,
+                   lock_obtainer: nil,
                    page_size: PAGE_SIZE)
       @repository = repository
       @page_size = page_size
       @event_broker = event_broker
+      @lock_obtainer = lock_obtainer
     end
 
     def publish_event(event, stream_name = GLOBAL_STREAM, expected_version = :any)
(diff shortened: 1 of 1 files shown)


lib/rails_event_store/client.rb
```diff
@@ -3,10 +3,12 @@ class Client
 
     def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
                    event_broker: EventBroker.new,
+                   lock_obtainer: nil,
```
mpraglowski: This should be:

```ruby
   lock_obtainer: RubyEventStore::DefaultLockObtainer.new,
```

Do not rely on defaults from other gem. Be explicit.

gottfrois: or even better with something like this i guess?

```ruby
DefaultLockObtainer = Class.new(RubyEventStore::DefaultLockObtainer)

lock_obtainer: DefaultLockObtainer.new
```

mpraglowski:
Too much changes - need to be revised.

PR #37, opened 2016-06-21
Info about append_event

brissenden:


brissenden:
@mpraglowski Please take a look. I've added paragraph about appending events.

mpraglowski:
@brissenden +1

However it is hard to check what has changed based on commit :( we need some improvement here ;)

PR #39, opened 2016-06-28
Rails request details in event metadata.

pawelpacana:
The idea behind it is to automatically record some data not directly related to particular
event but useful in context of audit log.

What (if anything) to be recorded can be made confifgurable,I wanted to deliver working prototype first.

Possbily interesting stuff to record:
- user agent
- request id (i.e. to use as a correlation id for group of events)
- logged in user identifier (with help of some popular solutions like Devise)

### DIFF
--- a/lib/rails_event_store.rb
+++ b/lib/rails_event_store.rb
@@ -11,3 +11,4 @@ end
 require 'rails_event_store/version'
 require 'rails_event_store/client'
 require 'rails_event_store/constants'
+require 'rails_event_store/railtie' if defined?(Rails::Railtie)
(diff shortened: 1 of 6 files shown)


mlomnicki:
fantastic idea. A bit magic but the benefit is HUGE

andrzejkrzywda:
good idea, it's going to be very useful

also, the implementation looks ok. it's hard to imagine a different implementation than through the `Thread.current` which would be non-invasive.

pawelpacana:
Please check new commit with metadata custoization.

gottfrois:
Is this something we can deactivate? Correct me if I'm wrong but here we assume that commands come from http request, through rails controllers. What about other protocols? In my usecase, commands are captured by rabbitmq consumers and won't have the rails metadata the event store expect.

pawelpacana:
@gottfrois This extra metadata will be recorded only in HTTP context (via Rack middleware). In case of publishing from other sources (like external process not tied to request-response cycle) none will be added (empty `Thread.current[:rails_event_store]`).

In HTTP context, what is added is also configurable. By default we record request id and ip address. It can be made none as well.

PR #42, opened 2016-06-29
Docs upgraded to 0.9

Killavus:
* Added a section about intercepting metadata and configuring this behaviour.
* Updated documentation about subscriptions.
* Reorganized the structure of docs a little.

### DIFF
--- a/docs/request_metadata.md
+++ b/docs/request_metadata.md
@@ -1,0 +1,70 @@
+# Logging request metadata 
+
+In Rails environment, every event is enhanced with the request metadata provided by `rack` server. This can help with debugging and building an audit log from events for the future use.
+
+## Defaults
+
+With default configuration metadata is enhanced by:
+
+* `:remote_ip` - IP of the HTTP client which issued the request.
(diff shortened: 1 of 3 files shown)


docs/request_metadata.md
```diff
@@ -0,0 +1,70 @@
...
+With default configuration metadata is enhanced by:
+
+* `:remote_ip` - IP of the HTTP client which issued the request.
+* `:request_id` - An unique ID of the request.
+
+This metadata is included **only for published events**. So creating a new event instance by hand won't add metadata to it:
+
+```ruby
+event = MyEvent.new(event_data)
+event.metadata # nil, unless you provided your own data called 'metadata'.
```
mlomnicki: nit-pick, but this isn't true. The metadata is a hash and contains the timestamp

https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/event.rb#L14

Killavus: Great! I just quickly scanned the code so I just missed it. Would you rephrase this paragraph to be correct?

PR #51, opened 2016-08-08
Default args as keywords

antonpaisov:


mpraglowski:
This should be aligned with https://github.com/arkency/ruby_event_store/pull/27 but let's not wait for new ruby_event_store release. Will apply further changes in next PR.

PR #56, opened 2016-09-13
Removes ActiveRecord dependencies

gottfrois:
This is a breaking change since we will not default on the active record
event repository anymore.
Clients will need to explicitely use the proper RailsEventStore repository in
their Gemfile.

Before:

``` ruby
gem 'rails_event_store'
```

After:

``` ruby
gem 'rails_event_store'
gem 'rails_event_store_active_record'
```

Nevertheless, the gem will automatically load the proper repository.
No need to explicitely set the repository when instantiating the event store:

``` ruby
RailsEventStore::Client.new(repository: RailsEventStoreActiveRecord::EventRepository.new)
```

The goal is to prepare the code base to handle other ORM adapters, such as [Mongoid](https://github.com/gottfrois/rails_event_store_mongoid).

### DIFF
--- a/lib/rails_event_store.rb
+++ b/lib/rails_event_store.rb
@@ -1,14 +1,14 @@
 require 'ruby_event_store'
-require 'rails_event_store_active_record'
 
 module RailsEventStore
-  Event               = RubyEventStore::Event
-  InMemoryRepository  = RubyEventStore::InMemoryRepository
-  EventBroker         = RubyEventStore::PubSub::Broker
-  Projection          = RubyEventStore::Projection
+  Event           = RubyEventStore::Event
+  EventBroker     = RubyEventStore::PubSub::Broker
+  EventRepository = RubyEventStore::InMemoryRepository
+  Projection      = RubyEventStore::Projection
 end
 
 require 'rails_event_store/version'
+require 'rails_event_store/event_repository'
 require 'rails_event_store/client'
 require 'rails_event_store/constants'
 require 'rails_event_store/railtie' if defined?(Rails::Railtie)
--- a/lib/rails_event_store/client.rb
+++ b/lib/rails_event_store/client.rb
@@ -1,7 +1,7 @@
 module RailsEventStore
   class Client
 
-    def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
+    def initialize(repository: EventRepository.new,
                    event_broker: EventBroker.new,
                    page_size: PAGE_SIZE)
       @repository = repository
--- a/lib/rails_event_store/event_repository.rb
+++ b/lib/rails_event_store/event_repository.rb
@@ -1,0 +1,11 @@
+module RailsEventStore
+  if defined?(RailsEventStoreActiveRecord)
+    require 'rails_event_store_active_record'
+    EventRepository = RailsEventStoreActiveRecord::EventRepository
+  end
+
+  if defined?(RailsEventStoreMongoid)
+    require 'rails_event_store_mongoid'
+    EventRepository = RailsEventStoreMongoid::EventRepository
+  end
+end
--- a/spec/middleware_integration_spec.rb
+++ b/spec/middleware_integration_spec.rb
@@ -5,13 +5,16 @@ require 'rack/lint'
 
 module RailsEventStore
   RSpec.describe Middleware do
-    DummyEvent = Class.new(RailsEventStore::Event)
+
+    let(:dummy_event_class) do
+      Class.new(RailsEventStore::Event)
+    end
 
     specify do
       event_store = Client.new
 
       request = ::Rack::MockRequest.new(Middleware.new(
-        ->(env) { event_store.publish_event(DummyEvent.new); [200, {}, ["Hello World from #{env["SERVER_NAME"]}"]] },
+        ->(env) { event_store.publish_event(dummy_event_class.new); [200, {}, ["Hello World from #{env["SERVER_NAME"]}"]] },
         ->(env) { { server_name: env['SERVER_NAME'] }}))
       request.get('/')
 
@@ -21,5 +24,3 @@ module RailsEventStore
     end
   end
 end
-
-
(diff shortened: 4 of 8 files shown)


gottfrois:
Hey @mpraglowski, would gladly accept feedbacks on this when you have time :)

gottfrois:
Any chance to get this PR reviewed anytime soon? :)

mpraglowski:
Will try during the weekend. Sorry for long waiting time :(

lib/rails_event_store.rb
```diff
@@ -1,14 +1,14 @@
...
-require 'rails_event_store_active_record'
 
 module RailsEventStore
-  Event               = RubyEventStore::Event
-  InMemoryRepository  = RubyEventStore::InMemoryRepository
-  EventBroker         = RubyEventStore::PubSub::Broker
-  Projection          = RubyEventStore::Projection
+  Event           = RubyEventStore::Event
+  EventBroker     = RubyEventStore::PubSub::Broker
+  EventRepository = RubyEventStore::InMemoryRepository
```
mpraglowski: Why rename here? I would prefer to know what kind of event repository we use.

lib/rails_event_store/event_repository.rb
```diff
@@ -0,0 +1,11 @@
+module RailsEventStore
```
mpraglowski: Hmm I don't like this approach. It will require us to add here new code each time someone will implement new event repository. I would just remove it if we will go that way (not having RESAR in RES dependencies).

lib/rails_event_store/client.rb
```diff
@@ -1,7 +1,7 @@
 module RailsEventStore
   class Client
 
-    def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
```
mpraglowski: The idea of RES gem (and a separation a RubyEventStore gem) was to have an "umbrella" gem, one gem to install to have complete setup, with defaults. This change will break that assumption. You will have to remember to add event repository gem in the Gemfile. You will have to remember to setup proper event repository.

spec/middleware_integration_spec.rb
```diff
@@ -5,13 +5,16 @@
 
 module RailsEventStore
   RSpec.describe Middleware do
-    DummyEvent = Class.new(RailsEventStore::Event)
```
mpraglowski: What's the difference?

lib/rails_event_store.rb
gottfrois: good point

lib/rails_event_store/event_repository.rb
gottfrois: Sure, the other way to do this is to have the dependency actually load the proper repository by setting the RailsEventStore adapter at run time.

Would that work for you?

lib/rails_event_store/client.rb
gottfrois: I understand the idea. How would you handle the fact that, if i want to use RES with MongoDB, I don't want to install AR dependencies?

We could improve the following PR by having each dependencies taking care of the ORM (RES_AR, RES_MONGO, ...) setting the proper adapter, event repository and any other requirements when the gem loads.

spec/middleware_integration_spec.rb
gottfrois: Creating the class in `let` scope will not pollute the global namespace, the class only exists in this spec execution context. Polluting the global namespace might break other specs latter down the chain because a class is already defined for example.

gottfrois:
@mpraglowski I would be please to refactor that pull request to further meet your requirements but I'm really not sure how you want the ORM dependency to be handled/managed.

I understand the idea behind the concept of "install one gem and start working with defaults" but i'm not sure to see how we can mix this concept and not install unnecessary AR dependencies if we want to use another ORM.

mpraglowski:
Ok, finally I had some time to think about that and prepare my answer. Sorry for delays :(

I've created a PR in RubyEventStore project https://github.com/arkency/ruby_event_store/pull/35

With that change in mind take a look at RailsEventStore. The lib folder is only:

```
lib/
├── rails_event_store
│   ├── client.rb  => soon might be only a constructor for RubyEventStore::Client with RailsEventStoreActiveRecord repository and a proc to fill in metadata from middleware.rb
│   ├── constants.rb => to be removed, not needed anymore, duplication of things from RubyEventStore
│   ├── middleware.rb => Rails related stuff
│   ├── railtie.rb => Rails related stuff
│   └── version.rb
└── rails_event_store.rb
```

I would not push here more changes, this gem should stay just an "umbrella gem" to setup all building blocks to be easy for use with Rails, just add `gem 'rails_event_store'` and do your work.
The solution for you to have not dependency on rails_event_store_active_record is just create a similar gem: maybe `mongo_event_store` with:

```
lib/
├── mongoid_event_store
│   ├── client.rb  =>  constructor for RubyEventStore::Client with rails_event_store_mongoid repository
│   └── version.rb
└── mongoid_event_store.rb
```

Or you could just replace `rails_event_store` in your appplication with `ruby_event_store` and use it with your rails_event_store_mongoid (now the best name maybe :) ).

>  i'm not sure to see how we can mix this concept and not install unnecessary AR dependencies if we want to use another ORM.

The answer here is - just don't do it. Instead of `gem 'rails_event_store'` use `gem 'ruby_event_store'` & `gem 'rails_event_store_mongoid'` :)

Will that work for you?

gottfrois:
Thanks for the answer. The idea sounds good, but that maybe means the `rails_event_store` name is not accurate. It would make more sense to rename it into something like `rails_event_store_active_record` or `active_record_event_store`.

What do you think?

mpraglowski:
> The idea sounds good, but that maybe means the rails_event_store name is not accurate.

Maybe, but for most of Rails devs the Rails is AR - I would keep that for simplicity and not to change the gem name.

mpraglowski:
@gottfrois are you ok with my vision, would you like us to host `mongoid_event_store` repository ?

gottfrois:
Sounds good to me, yes definitely. I'll transfert the ownership of the repo. I need to find some time to change the PR to reflect what we discussed about :)

mpraglowski:
@gottfrois this is no longer valid I assume - should be closed ?

gottfrois:
Correct!

PR #59, opened 2016-10-27
Allows to specify repository adapter to use

gottfrois:
@mpraglowski would this approach be ok with your vision? It use active_record by default but allows to plug any other repository adapter.

Usage:

```
gem 'rails_event_store'
```

That's it. If we want to change the adapter, for example in `rails_event_store_mongoid`:

```
gem 'rails_event_store'
```

That's it, the `rails_event_store_mongoid` will do internaly:

```
RailsEventStore::Repository.adapter = :mongoid
```

While updating my previous PR I notice that I actually don't want to re-create a clone of RailsEventStore that would know how to deal with Mongoid. If we do that, any "nice feature to have" that we include in RailsEventStore will need to be copied over other gems. The goodies given by the railties present in RailsEventStore gem are a good example.

If you are ok with this approach, I'll write tests for it.
Please let me know

### DIFF
--- a/lib/rails_event_store.rb
+++ b/lib/rails_event_store.rb
@@ -1,25 +1,25 @@
 require 'ruby_event_store'
-require 'rails_event_store_active_record'
 
 module RailsEventStore
-  Event               = RubyEventStore::Event
-  InMemoryRepository  = RubyEventStore::InMemoryRepository
-  EventBroker         = RubyEventStore::PubSub::Broker
-  Projection          = RubyEventStore::Projection
-
-  GLOBAL_STREAM       = RubyEventStore::GLOBAL_STREAM
-  PAGE_SIZE           = RubyEventStore::PAGE_SIZE
-
-  WrongExpectedEventVersion  = RubyEventStore::WrongExpectedEventVersion
-  InvalidExpectedVersion     = RubyEventStore::InvalidExpectedVersion
-  IncorrectStreamData        = RubyEventStore::IncorrectStreamData
-  EventNotFound              = RubyEventStore::EventNotFound
-  SubscriberNotExist         = RubyEventStore::SubscriberNotExist
-  MethodNotDefined           = RubyEventStore::MethodNotDefined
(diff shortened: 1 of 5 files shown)


lib/rails_event_store.rb
```diff
@@ -1,25 +1,25 @@
 require 'ruby_event_store'
-require 'rails_event_store_active_record'
 
 module RailsEventStore
-  Event               = RubyEventStore::Event
-  InMemoryRepository  = RubyEventStore::InMemoryRepository
```
mpraglowski: I would keep it - useful in tests.

mpraglowski:
A little magic ;) Let's give it a try. Kill the mutants please ;)

gottfrois:
Not sure how to kill the last 2 mutants, can you look at it?

```
evil:RailsEventStore::Repository.adapter=:/Users/gottfrois/Code/rails_event_store/lib/rails_event_store/repository.rb:7:58c0d
@@ -1,10 +1,10 @@
 def self.adapter=(adapter)
   case adapter
   when String, Symbol
-    require("#{"rails_event_store_"}#{adapter}")
+    require("#{nil}#{adapter}")
     @@adapter = "#{"::RailsEventStore"}#{adapter.to_s.classify}#{"::EventRepository"}".constantize.new
   else
     @@adapter = adapter
   end
 end
-----------------------
evil:RailsEventStore::Repository.adapter=:/Users/gottfrois/Code/rails_event_store/lib/rails_event_store/repository.rb:7:ff5cf
@@ -1,10 +1,10 @@
 def self.adapter=(adapter)
   case adapter
   when String, Symbol
     require("#{"rails_event_store_"}#{adapter}")
     @@adapter = "#{"::RailsEventStore"}#{adapter.to_s.classify}#{"::EventRepository"}".constantize.new
   else
-    @@adapter = adapter
+    @@adapter = nil
   end
 end
```

gottfrois:
Thinking more about this, it cannot work. Let me get back to you when I found a better solution

mpraglowski:
ok - keep working - I would love to see you use cases for RES.

PR #62, opened 2016-11-24
Fixes where is initialized event repository

gottfrois:
The event repository initialization must happen inside `all.rb` since on other repo (mongoid for example) we don't require `rails_event_store.rb` but only `rails_event_store/all.rb`


mpraglowski:
@gottfrois I've released v0.14.3 with this fix

PR #64, opened 2016-12-31
RSpec matchers

pawelpacana:
I'd like to introduce RSpec matchers in order to avoid reimplementing them in every project using `rails_event_store`. So far I have following API idea:

```
# check published events
expect(stream).to have_events([...])     # exact match
expect(stream).to have_event(...)        # exact match
expect(stream).to include_events([...])  # partial match
expect(stream).to include_event(...)     # partial match

```

```
# check unpublished_events
expect(aggregate_root).to publish(an_event(CouponRedeemed).with_data(code: '123'))
expect(aggregate_root).to publish(CouponRedeemed.new(data: { code: '123' }))

```

My doubts: 
1. I've started it as a part of `rails_event_store` but maybe it should be in `ruby_event_store`. And in `aggregate_root`. Or in completely separate gem. I'm confused.
2. Does it make sense to provide matchers against event store and aggregate root? An already implemented `be_event` matcher can cover those cases (with some verbosity), i.e.:

```
expect(event_store.read_stream_events_forward(stream_name)).to include([
  an_event(CouponRedeemed).with_metadata(remote_ip: '1.2.3.4')
]) 
```
3.  Does it make sense to have "singular" versions of matchers, that is both `have_event` and `have_events`?


andrzejkrzywda:
as for where to put such code - I'd like to keep everything below `rails-event-store` magic-free and ideally framework-free (including the test framework of choice).

pkondzior:
I was thinking about implementing some of argument matching functionality so you could  match event key values through with() matchers like instance_of, hash_including, kind_of etc.

mpraglowski:
> Does it make sense to provide matchers against event store and aggregate root?

It definitely makes sense for aggregate root. In most test for aggregates I check domain events generated as a result of some action on aggregate, without storing it anywhere.

swistak35:
Other useful matchers suggestion:

```
expect(stream).to have_n_events_of_type(ProductItemSold, 3)
```

I very like @pkondzior idea about employing `rspec` power of `match_array`, `include`, it would solve hell out of cases when one would have to mock/deal with `SecureRandom.uuid`, just because we have to put exact data in event's data hash in the matcher. What I mean is the case when we have to write something like this:

```
order_uuid = SecureRandom.uuid

... (some usage of service objects / services, whatever, which create order with `order_uuid`)

expect(stream).to include_event(OrderPlaced.new({ id: order_uuid, some_relevant_data: "foo" }))
```

instead of 

```
expect(stream).to include_event(OrderPlaced.new({ id: be_present, some_relevant_data: "foo" }))
```

swistak35:
Also, one note: if you are implementing matcher like `expect(stream).to include_event(...)` don't forget to print relevant message if the stream is empty - when the array of events is empty, error message after running test can be very deceptive and you can waste long time to deduce that test output is trying to tell you that stream is empty :)

swistak35:
One more matcher request/idea: it would be useful to have `include_exactly_one_event`:
```
expect(stream).to include_exactly_one_event(OrderPlaced.new({ id: be_present, some_relevant_data: "foo" }))
```

It'd behave almost like `include_event`, but with stronger assumption (that we have exactly one event matching, not at least one. Usecase: We often *know* that we can use stronger assertion, but we don't use that. It can result in all these nasty bugs where something happened too many times (via Sebastian Sogamoso's talk "The Overnight Failure" from wroc_love.rb 2017).

andrzejsliwa:
@andrzejkrzywda @pawelpacana : this can be done in separate gem: rails_event_store_rspec, with dependency to specific rspec and rails_event_store versions. But maybe would be better to just extract all gems to separate organisation, and keep here all main and tooling gems related to event sourcing and cqrs. This will also put more understanding about all building blocks

pawelpacana:
This has been solved in https://github.com/RailsEventStore/rails_event_store-rspec. Matchers work as documented in http://railseventstore.org/docs/rspec/. 

Thanks for the feedback so far! In case of further ideas and issues - please report them in https://github.com/RailsEventStore/rails_event_store-rspec/issues

PR #67, opened 2017-03-25
Updated next DDD workshop date.

scottccoates:
I just started using rails and was surprised to find a lively DDD community, and I couldn't believe it—that's awesome. Thanks for the work you've done in this space. I felt compelled to update the next DDD conf date.


fidel:
Thank you for the kind words & this PR

PR #73, opened 2017-08-20
Add AS::Notifications instrumentation

pawelpacana:
Rails provides quite solid framework for logging/instrumentation. We
could use it to provide better insight into event handling. Popular
metric tools (Datadog, Librato) can already plug into that.
One could also imagine plugging log subscriber for debugging purpose.
http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html
http://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

This commit adds new dispatcher, only for RailsEventStore and makes it a
default.

The way a new default was baked into EventBroker was driven primarily by
existing linters/shared_examples and mutant - likely to be refactored.
Good enough for now to start discussion about relevance of this change.


pawelpacana:
And it crashed mutant with `spec/event_broker_spec.rb` 🤔

pawelpacana:
Reworked and reduced surface of changes - just another dispatcher in this PR, with little help of just released dispatcher lint.

Changing default dispatcher could be another change, leaving intact now. 

Discussion still welcome!

pawelpacana:
Note to myself: check if error in subscriber of `AS::Notification` can affect open transaction with `publish_event`

spec/client_spec.rb
```diff
@@ -13,6 +13,11 @@ module RailsEventStore
       expect(client.__send__("event_broker")).to be_a EventBroker
     end
 
+    specify 'initialize proper dispatcher type' do
+      expect(EventBroker).to receive(:new).with(dispatcher: an_instance_of(Dispatcher))
```
mlomnicki: looks a bit useless. Is it needed because mutant complains otherwise?

mlomnicki:
it will be a super useful feature :+1:
also the implementation looks good to me

spec/client_spec.rb
pawelpacana: exactly

PR #84, opened 2017-08-29
Use local dependencies

mlomnicki:
Currently rails_event_store uses released versions of ruby_event_store
and rails_event_store_active_record. This is inconvenient for
development and testing.

Thanks to this patch rails_event_store will use local dependencies from the repo.

It allows to change ruby_event_store and quickly test how the change affects
rails_event_store.

Also the CI will run the entire test suite with the most recent dependencies
rather than with the released ones.

### DIFF
--- a/aggregate_root/Gemfile
+++ b/aggregate_root/Gemfile
@@ -1,2 +1,4 @@
 source 'https://rubygems.org'
 gemspec
+
+gem 'ruby_event_store', path: '../ruby_event_store'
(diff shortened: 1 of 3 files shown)


paneq:
@mlomnicki don't forget about aggregate_root:

```
gem 'ruby_event_store', path: '../ruby_event_store'
```

mlomnicki:
@paneq great catch! 👏

PR #86, opened 2017-08-29
Optimistic Locking friendly changes

paneq:
It started here: https://github.com/RailsEventStore/aggregate_root/issues/8

# 3 possible ways of writing

## Number

* `:none` - works like `-1` assumes stream empty so far, starts adding new in stream
* given number : -1..Infinity - assumes N, starts adding N+1, N+2

Should work well for 

* non-legacy
* event sourcing scenario
* transaction around not required

## Auto

* `:auto` - assumes lock in higher layer. Will query for last N and start doing N+1, N+2... etc [New mode]
* good for legacy
* requires transaction and lock around

## Any

* `:any` - NULLable position in stream, order unknown
* For copying into highly-contagious technical streams where we don't really care about exact position 
* Best-guess order can be determined based on `EventInStream` auto-increment id.

This is just a spike that I wanted to share with you.

What do you think @mpraglowski @pawelpacana @mlomnicki ?

# Checklist

* [x] Mysql
* [x] Postgres
* [x] Re-enable mutation testing
* [ ] Verbose / non-verbose mode of running tests
* [ ] Change migration to be DB-dependant and use string for mysql UUID
* [x] Discuss 3 tables vs 2 tables
* [ ] Fix `enrich_event_metadata` to not edit `metadata` directly but rather a cloned/duped version.
* [ ] GLOBAL_STREAM is `all` instead of `__global__` and it should only write there once.
* [x] Consider global linearization with a very quick lock on write...
* [x] Make sure `AggregateRoot` can handle conflicts 
* [x] Should event be unique in a stream?


paneq:
right now it  is missing required Travis changes to run `rails_event_store_active_record` on Mysql and Postgres

rails_event_store_active_record/spec/spec_helper.rb
```diff
@@ -1,26 +1,55 @@
+if ENV['CODECLIMATE_REPO_TOKEN']
```
mlomnicki: not needed anymore. We got rid of CC

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
     end
     attr_reader :adapter
 
-    def create(event, stream_name)
-      data = event.to_h.merge!(stream: stream_name, event_type: event.class)
-      adapter.create(data)
-      event
+    def append_to_stream(events, stream_name, expected_version)
+      events = [*events]
+      expected_version   = case expected_version
```
gottfrois: this should be in its own private method. It makes the `append_to_stream` very hard to reason about otherwise, also violates single responsibility principle

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+        when nil
+          raise RubyEventStore::InvalidExpectedVersion
+        when :none
+          -1
+        when :auto
+          eis = EventInStream.where(stream: stream_name).order("position DESC").first
```
gottfrois: we should use a repository to access this

rails_event_store_active_record/spec/spec_helper.rb
```diff
@@ -1,26 +1,55 @@
...
+
+      create_table(:event_store_events_in_streams, force: true) do |t|
         t.string      :stream,      null: false
-        t.string      :event_type,  null: false
-        t.string      :event_id,    null: false
-        t.text        :metadata
-        t.text        :data,        null: false
+        t.integer     :position,    null: true
+        if ENV['DATABASE_URL'].start_with?("postgres")
+          t.references :event, null: false, type: :uuid
```
mlomnicki: Do we actually need the :uuid? Maybe :string would suffice? With :string we don't have to enable pgcrypto

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
gottfrois: it should also be in its own private method

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+          (eis && eis.position) || -1
+        else
+          expected_version
+      end
+      in_stream = events.flat_map.with_index do |event, index|
+        position = if expected_version == :any
```
gottfrois: extract in its own method

rails_event_store_active_record/spec/spec_helper.rb
paneq: And what's wrong with pgcrypto?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+          nil
+        else
+          expected_version + index + 1
+        end
+        Event.create!(
```
gottfrois: should use the `adapter`

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+          event_id: event.event_id
+        ),EventInStream.new(
+          stream: "__global__",
+          position: nil,
+          event_id: event.event_id
+        )]
+      end
+      EventInStream.import(in_stream)
+      self
+    rescue ActiveRecord::RecordNotUnique
```
gottfrois: this class should know nothing about active record

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+        )
+        [EventInStream.new(
+          stream:   stream_name,
+          position: position,
+          event_id: event.event_id
+        ),EventInStream.new(
+          stream: "__global__",
+          position: nil,
+          event_id: event.event_id
+        )]
```
gottfrois: better indentation technique would be easier to read IMO

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -1,14 +1,51 @@
...
+        else
+          expected_version + index + 1
+        end
+        Event.create!(
+          id: event.event_id,
+          data: event.data,
+          metadata: event.metadata,
+          event_type: event.class,
+        )
+        [EventInStream.new(
```
gottfrois: should be extracted into its own method as well

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -17,75 +54,75 @@ def delete_stream(stream_name)
     end
 
     def has_event?(event_id)
-      adapter.exists?(event_id: event_id)
+      Event.exists?(id: event_id)
```
gottfrois: should use the adapter

mlomnicki:
Excellent job @paneq! This is quite a big PR. Could we maybe split it into smaller chunks?

The checklist already contains some bits that could be extracted to separate PRs (mysql/postrgres, fix enrich_event_metadata, etc)

I would also open separate PRs for the following:
* ability to publish multiple events
* Event#hash
* Discuss - do we have to depend on activerecord-import?

Makes sense?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -17,75 +54,75 @@ def delete_stream(stream_name)
...
     end
 
     def last_stream_event(stream_name)
-      build_event_entity(adapter.where(stream: stream_name).last)
+      build_event_entity(
+        EventInStream.preload(:event).where(stream: stream_name).order('position DESC, id DESC').first
```
gottfrois: should use the event in stream repository like the other comment earlier

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -17,75 +54,75 @@ def delete_stream(stream_name)
...
+      )
     end
 
-    def read_events_forward(stream_name, start_event_id, count)
-      stream = adapter.where(stream: stream_name)
-      unless start_event_id.equal?(:head)
-        starting_event = adapter.find_by(event_id: start_event_id)
-        stream = stream.where('id > ?', starting_event)
+    def read_events_forward(stream_name, after_event_id, count)
```
gottfrois: same comments about adapters

rails_event_store_active_record/spec/spec_helper.rb
mlomnicki: Nothing wrong. It's not enabled by default. Are we going to force users to enable it even though it's not really needed?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: The code we are writing here is the repository :)

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: I don't see the point of this `adapter` anymore now that the repository needs to work with 2-3 ActiveRecord classes instead of 1.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: Why would `rails_event_store_active_record/event_repository.rb` need to know nothing about active record when it actively uses active record to implement the needed features.

rails_event_store_active_record/spec/spec_helper.rb
paneq: Yep :)

paneq:
> The checklist already contains some bits that could be extracted to separate PRs

I agree. It's more so that I don't forget about those things than to actually include in this PR.

rails_event_store_active_record/spec/spec_helper.rb
mlomnicki: 🤔

paneq:
> ability to publish multiple events

well... We could add such functionality in separate PR without releasing a new version because publishing multiple events using the old code is still going to be completely broken from optimistic locking perspective but maybe a separate PR for extending API is a good idea? Dunno

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
gottfrois: it's the event repository, not EventInStream repository. I would make a clear distinction

paneq:
> Event#hash

No brainer. Feel free to extract and merge. Very simple. Related mutant discussion https://twitter.com/pankowecki/status/902521997573455877

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
gottfrois: nvm, didn't see we are in active record namespace :)

paneq:
> Discuss - do we have to depend on activerecord-import?

Ideally not, but I didn't want to write myself all the code for doing a single INSERT with multiple records. It might not be relevant if we migrate to transactions and 3 tables approach.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
gottfrois: i think it makes it way harder to re-write this repository for another ORM if they are a mixed of multiple active record classes instead of a single responsibility repository which maps to one active record class

paneq:
@gottfrois Thanks for the comments about extracting methods etc. This is right now however totally Work In Progress in which I am trying to find out if what we are trying to achieve is possible within the constraints that I described. I am still not exactly sure and there are bigger decisions to be made (3vs2 tables), handling concurrency etc. So once I am happy with the Big decisions, I am gonna focus on refactoring the smaller parts.

paneq:
@gottfrois Also, I am trying to reach a state in which this can be merged to master, unreleased. And then everyone can easily improve the codebase with proper tests until we are satisfied with the results. And the we could do an official release. So I am gonna dismiss the review right now. But does not mean I don't agree. I just want to handle those kind of things later :)

gottfrois:
no hard feelings :)

mlomnicki:
> > Event#hash

> No brainer.

Apparently mere developers such as me don't understand it :) I.e. how is this change related to optimistic locking? Why the big number is written in binary notation? Why do we even need to overwrite Event#hash, etc

paneq:
> Apparently mere developers such as me don't understand it :) I.e. how is this change related to optimistic locking? 

In one race condition test I wanted to guarantee that all events are returned but the order is not known so I used `Array#to_set` and check if two sets are equal. That failed. Which brings as to your next question:

> Why do we even need to overwrite Event#hash, etc

Because our Event has `==` operator which makes it behave like a Value Object (ignoring metadata). But value objects with are equal in `==` sense should also return the same `hash` which is used when you put such object into `Hash` or `Set`.

```
       expect({
         klass.new(event_id: "doh") => :YAY
       }[ klass.new(event_id: "doh") ]).to eq(:YAY)
```

```
       expect(Set.new([
         klass.new(event_id: "doh")
       ])).to eq(Set.new([klass.new(event_id: "doh")]))
```

These two are failing without implementing `hash`

> Why the big number is written in binary notation?

It's easier that way to see how many bits it has.

mlomnicki:
Wow that's a proper explanation. Much appreciated, thanks!!

paneq:
Regarding mutations in `rails_event_store_active_record`... I don't think it makes sense to run them on postgres or mysql. Probably I should bring back Sqlite into testing suite and mutate only on that.

paneq:
Current status:

selection_560

paneq:
Today I realized that ideally I would want to come out with a DB schema that allows for https://github.com/RailsEventStore/rails_event_store/issues/106 and since I don't want to keep the DB schema constantly I kind of coupled both problems together.

paneq:
@pawelpacana @mlomnicki before we merge this I have a few questions. Let's start with most important one:

* `add_index :event_store_events_in_streams, [:stream, :event_uuid], unique: true` - Do you think we need it? I think it makes sense. Given event should appear in a stream once and only once. `read_events_forward` and other similar methods basically rely on that fact because they find position based on event id and if there were two events with same id in a stream then we could have random results and failures.

mlomnicki:
@paneq yep, defo makes sense :+1:

paneq:
@mlomnicki @pawelpacana I want to merge today, please review holistically :)

The idea behind `__global__` stream is that it contained all events but only once. It might not be needed right now but when we add the feature to copy event between streams it could be useful to have `__global__` without duplicates.

So publishing in stream `a` would add event in stream `a` and `__global__` and linking an event into stream `b` would only add it in stream `b`, without adding again to `__global__`. Does that make sense?

paneq:
So far I was not sure how we could introduce a 3rd table `streams` with `name` and `position` and unique index on `name` that could work in `:any` mode and not have a race condition on creating the record. This is an actual problem we had in another system where 2 background jobs tried to write first event to the same stream.

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -87,7 +97,7 @@ def enrich_event_metadata(event)
       metadata[:timestamp] ||= clock.()
       metadata.merge!(metadata_proc.call || {}) if metadata_proc
 
-      event.class.new(event_id: event.event_id, metadata: metadata, data: event.data)
+      # event.class.new(event_id: event.event_id, metadata: metadata, data: event.data)
```
pawelpacana: what's up with this comment?

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
```diff
@@ -6,108 +6,370 @@
...
-    expect(repository.read_all_streams_backward('5', 4)).to eq ['4','3','2','1'].map{|x| TestDomainEvent.new(event_id: x)}
-    expect(repository.read_all_streams_backward('5', 100)).to eq ['4','3','2','1'].map{|x| TestDomainEvent.new(event_id: x)}
+    expect(repository.read_all_streams_backward(:head, 3)).to eq(events.last(3).reverse)
+    expect(repository.read_all_streams_backward(:head, 100)).to eq(events.reverse)
+    expect(repository.read_all_streams_backward(events[4].event_id, 4)).to eq(events.first(4).reverse)
+    expect(repository.read_all_streams_backward(events[4].event_id, 100)).to eq(events.first(4).reverse)
   end
-end
+
+  it 'reads events different uuid object but same content' do
```
pawelpacana: I don't quite get the description about _different uuid object_

ruby_event_store/lib/ruby_event_store/client.rb
paneq: Changing metatadata inside the instance vs new instance with new metadata. Keeping the old behavior was not possible during refactorings. A separate task would be to bring back the old behavior if still wanted, but test it properly. It was not.

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
paneq: I wasn't clear at all here. we provide different string instance containing the same UUID. This makes it possible to establish whether we need `==` or `eql?` or `equal?` comparison logic. So instead of using the same variable, we use different string variable but with identical content.

PR #89, opened 2017-08-30
Release management replacement for monorepo

pawelpacana:
Previously we've used `rake release` on each of the gems that handled:
- setting git tag
- building gem pkg
- pushing pkg to rubygems

That was automated with Bundler and
https://github.com/bundler/bundler/blob/46b4b2f23c3d50314bed694fc634380798ba0721/lib/bundler/gem_helper.rb.

We've also had to bump gem version and update dependencies (if neeeded).

Now that we have monorepo we don't have this out of the box. This
commit brings back release automation with `make release`:
- version is read from `RES_VERSION` file and set in all gems and gemspecs
  (dependencies version)
- git tag is set and pushed
- pkg build and rubygems push executed for each gem


Makefile
```diff
@@ -39,8 +39,45 @@ test-rails-event-store-active-record:
...
+release-aggregate-root:
+	@make -C aggregate_root release
+
 install: install-aggregate-root install-ruby-event-store install-rails-event-store install-rails-event-store-active-record ## Install all deps
 
 test: test-aggregate-root test-ruby-event-store test-rails-event-store test-rails-event-store-active-record ## Run all specs
 
 mutate: mutate-aggregate-root mutate-ruby-event-store mutate-rails-event-store mutate-rails-event-store-active-record ## Run all mutation tests
+
+release: git-check-clean git-check-committed set-version git-tag release-rails-event-store release-ruby-event-store release-rails-event-store-active-record release-aggregate-root
```
mlomnicki: if you add a comment at the end of the line then this target will be displayed in `make help`

```
release: git-check-clean .... release-aggregate-root ## Make a new release and push to rubygems
```

```
$ make help
install    Install all deps
test       Run all tests
release    Make a new release and push to rubygems

pawelpacana:
Fixed comment, task now appears in listing. Merge it and we're ready for `0.15.0` release :)

PR #90, opened 2017-08-30
Add AS::Notifications instrumentation

pawelpacana:
Reopened (previously as #73) after force-push to master ;)

Rails provides quite solid framework for logging/instrumentation. We
could use it to provide better insight into event handling. Popular
metric tools (Datadog, Librato) can already plug into that.
One could also imagine plugging log subscriber for debugging purpose.
http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html
http://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

### DIFF
--- a/rails_event_store/spec/instrumentation_spec.rb
+++ b/rails_event_store/spec/instrumentation_spec.rb
@@ -26,4 +26,24 @@ module RailsEventStore
       expect(notifications.size).to eq(1)
     end
   end
+
+  RSpec.describe Client do
+    specify do
+      client = Client.new
+      event  = DummyEvent.new
+
+      notifications = []
+      callback = ->(name, started, finished, unique_id, data) do
+        notifications << unique_id
+
+        expect(data[:event]).to eq(event)
+      end
+
+      ActiveSupport::Notifications.subscribed(callback, "publish_event.rails_event_store") do
+        client.publish_event(event)
+      end
+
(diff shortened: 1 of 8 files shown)


rails_event_store/spec/instrumentation_spec.rb
```diff
@@ -46,4 +46,18 @@ module RailsEventStore
...
+      subscriber = ActiveSupport::Notifications.subscribe(/rails_event_store/, ->(*) { raise })
+
+      ActiveRecord::Base.transaction do
+        client.publish_event(DummyEvent.new)
+      end
+
+      ActiveSupport::Notifications.unsubscribe(subscriber)
+      expect(client.read_all_streams_backward).not_to be_empty
+    end
+  end
```
paneq: That's ok. ActiveSupport::Notifications subscriber is like a sync handler on a different layer and same rules apply. Devs needs to catch exceptions in sync handlers if they don't want them to bubble up.

paneq:
@pawelpacana Looks good, however considering `ActiveJobDispatcher` and the idea of pipelines, I would implement it using a decorator pattern so we can have:

```
RailsEventStore::Client.new(
  dispatcher: RailsEventStore::InstrumentDispatcher.new(
    RailsEventStore::ActiveJobDispatcher.new
  )
```

Issue #83, opened 2017-08-29
Run mutant on one Ruby version only


pawelpacana:
Testing with mutant takes time (a lot of it). Yet mutating on each Ruby version in matrix doesn't uncover new facts about missing code coverage. Unit tests should be enough to verify that functionality works on given Ruby verison.

Proposal: run `make mutate` on latest Ruby only

paneq:
I will work on this as I need it anyway for making `locking_friendly` pass.

mlomnicki:
I already started working on this here https://github.com/mlomnicki/rails_event_store/blob/20d0e2e6db0949a1025d58be6340a73b1ceb126d/.travis.yml

Feel free to use this config if you find it valuable

PR #97, opened 2017-08-31
Single mutation per ruby + DB connection string

paneq:
* Run mutant on one (newest) Ruby version only
  * https://github.com/RailsEventStore/rails_event_store/issues/83
* DB connection strings that we should test our code with
  * https://github.com/RailsEventStore/aggregate_root/issues/8
  * https://github.com/RailsEventStore/rails_event_store/pull/86

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,43 @@
 language: ruby
-rvm:
-  - 2.2.7
-  - 2.3.4
-  - 2.4.1
+matrix:
+  include:
+  - rvm: 2.4.1
+    env: TARGET=aggregate_root SCRIPT=mutate
+  - rvm: 2.4.1
+    env: TARGET=ruby_event_store SCRIPT=mutate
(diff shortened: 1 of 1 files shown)


pawelpacana:
I'm curious what's the rationale for postgres in:
```
rvm: 2.2.7
  env: TARGET=rails_event_store_active_record SCRIPT=test DATABASE_URL=postgres://localhost/rails_event_store_active_record?pool=5
```
and

```
rvm: 2.2.7
  env: TARGET=rails_event_store_active_record SCRIPT=test DATABASE_URL=postgres://localhost/rails_event_store_active_record?pool=5
```

paneq:
@pawelpacana good question. I simply assumed postgres as default.

pawelpacana:
if sqlite proves to be significantly faster then I'd change it for non-current Ruby runs, let's see in next PR

PR #101, opened 2017-08-31
Use a correct link to the docs

### LINKED ISSUE #71: README has no Installation section
While it is referenced in https://github.com/arkency/aggregate_root, README has no installation section

mlomnicki:
Fixes #71


pawelpacana:
btw. I like how Rails exposes minimal getting started section and then points to docs https://github.com/rails/rails#getting-started

PR #103, opened 2017-09-01
Problem: only sync handlers are possible

mpraglowski:
Solution: ActiveJob dispatcher and a new way of subscribing to domain events

`ActiveJobDispatcher` is a new dispatcher ready to use with RES as a replacement of `DefaultDispatcher`.
The main improvement is that if a handler class is an ancestor of `ActiveJob::Base` it will be handled asynchronously.
Also it defines a set of `ActiveJob` adapters for which enqueue of tasks to perform (handlers) need to be done only after transaction is committed.

The new way of subscribing to domain events:
- currently we have `subscribe(lambda_or_callable_class_instance, [domain_event], unsubscribe_proc)`, this will not allow us to use async jobs as we need to pass instance to RES client
- adding 2 new methods:
   - `on([domain_event], class_name, unsubscribe_proc)` and `on_all_events(class_name, unsubscribe_proc)`
   - for a klass given in `on` method a callable proxy will be created and will be passed to "old" `subscribe` method
   - dispatcher is responsible for creating a proxy - that allow customising the proxy based on dispatcher capabilities

### DIFF
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -21,32 +21,16 @@ module RailsEventStore
     end
 
     def async_proxy(klass)
-      raise InvalidHandler.new(klass) unless klass.respond_to?(:perform_later)
-      enqueue_now?(klass) ? enqueue_now_proxy(klass) : enqueue_after_commit_proxy(klass)
-    end
-
-    def enqueue_now_proxy(klass)
-      ->(e) { klass.perform_later(YAML.dump(e)) }
-    end
-
-    def enqueue_after_commit_proxy(klass)
       ->(e) {
         if ActiveRecord::Base.connection.transaction_open?
           ActiveRecord::Base.
             connection.
             current_transaction.
-            add_record( AsyncRecord.new(klass, e) )
+            add_record(AsyncRecord.new(klass, e))
         else
           klass.perform_later(YAML.dump(e))
         end
       }
-    end
-
-    def enqueue_now?(klass)
-      %w(
-        ActiveJob::QueueAdapters::InlineAdapter
-        ActiveJob::QueueAdapters::TestAdapter
-      ).include?(klass.queue_adapter.class.to_s)
     end
 
     class AsyncRecord
@@ -55,11 +39,7 @@ module RailsEventStore
         @event = event
       end
 
-      def has_transactional_callbacks?
-        true
-      end
-
-      def committed!(*_, **__)
+      def committed!
         @klass.perform_later(YAML.dump(@event))
       end
     end
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -1,7 +1,44 @@
 require 'active_job'
 
 module RailsEventStore
+  module AsyncProxyStrategy
+    class AfterCommit
+      def call(klass, event)
+        if ActiveRecord::Base.connection.transaction_open?
+          ActiveRecord::Base.
+            connection.
+            current_transaction.
+            add_record(AsyncRecord.new(klass, event))
+        else
+          klass.perform_later(YAML.dump(event))
+        end
+      end
+
+      private
+      class AsyncRecord
+        def initialize(klass, event)
+          @klass = klass
+          @event = event
+        end
+
+        def committed!
+          @klass.perform_later(YAML.dump(@event))
+        end
+      end
+    end
+
+    class Inline
+      def call(klass, event)
+        klass.perform_later(YAML.dump(event))
+      end
+    end
+  end
+
   class ActiveJobDispatcher
+    def initialize(proxy_strategy: AsyncProxyStrategy::Inline.new)
+      @async_proxy_strategy = proxy_strategy
+    end
+
     def call(subscriber, event)
       subscriber.call(event)
     end
@@ -21,27 +58,7 @@ module RailsEventStore
     end
 
     def async_proxy(klass)
-      ->(e) {
-        if ActiveRecord::Base.connection.transaction_open?
-          ActiveRecord::Base.
-            connection.
-            current_transaction.
-            add_record(AsyncRecord.new(klass, e))
-        else
-          klass.perform_later(YAML.dump(e))
-        end
-      }
-    end
-
-    class AsyncRecord
-      def initialize(klass, event)
-        @klass = klass
-        @event = event
-      end
-
-      def committed!
-        @klass.perform_later(YAML.dump(@event))
-      end
+      ->(e) { @async_proxy_strategy.call(klass, e) }
     end
   end
 end
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -54,7 +54,7 @@ module RailsEventStore
 
     def sync_proxy(klass)
       raise InvalidHandler.new(klass) unless klass.method_defined?(:call)
-      ->(e) { klass.new.call(e) }
+      klass.new
     end
 
     def async_proxy(klass)
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -1,0 +1,67 @@
+require 'active_job'
+
+module RailsEventStore
+  class ActiveJobDispatcher
+    def call(subscriber, event)
+      subscriber.call(event)
+    end
+
+    def proxy_for(klass)
+      async_handler?(klass) ? async_proxy(klass) : sync_proxy(klass)
+    end
+
+    private
+    def async_handler?(klass)
+      klass < ActiveJob::Base
+    end
+
+    def sync_proxy(klass)
+      raise InvalidHandler.new(klass) unless klass.method_defined?(:call)
+      ->(e) { klass.new.call(e) }
+    end
+
+    def async_proxy(klass)
+      raise InvalidHandler.new(klass) unless klass.respond_to?(:perform_later)
+      enqueue_now?(klass) ? enqueue_now_proxy(klass) : enqueue_after_commit_proxy(klass)
+    end
+
+    def enqueue_now_proxy(klass)
+      ->(e) { klass.perform_later(YAML.dump(e)) }
+    end
+
+    def enqueue_after_commit_proxy(klass)
+      ->(e) {
+        if ActiveRecord::Base.connection.transaction_open?
+          ActiveRecord::Base.
+            connection.
+            current_transaction.
+            add_record( AsyncRecord.new(klass, e) )
+        else
+          klass.perform_later(YAML.dump(e))
+        end
+      }
+    end
+
+    def enqueue_now?(klass)
+      %w(
+        ActiveJob::QueueAdapters::InlineAdapter
+        ActiveJob::QueueAdapters::TestAdapter
+      ).include?(klass.queue_adapter.class.to_s)
+    end
+
+    class AsyncRecord
+      def initialize(klass, event)
+        @klass = klass
+        @event = event
+      end
+
+      def has_transactional_callbacks?
+        true
+      end
+
+      def committed!(*_, **__)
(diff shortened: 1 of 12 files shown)


mlomnicki:
Good stuff :+1:

I'm wondering whether we actually need the sync proxy? We could let ActiveJob handle that

Sync
```ruby
class MySyncJob < ApplicationJob
   self.queue_adapter = :inline

   def perform(...)
   end
end
```

Async
```ruby
class MyAsyncJob < ApplicationJob
    # uses default adapter

   def perform(...)
   end
end
```

This would simply the implementation a little bit but on the other hand passing `self.queue_adapter = :inline` to every sync job seems a bit tedious. WDYT?

mlomnicki:
Question: does it deserialize events or is it up to the developer to do that

```ruby
class MyJob < ApplicationJob
   def perform(event)
     # what's the event here? YAML, JSON or Res::Event?
   end
end
```

mpraglowski:
Compatibility reasons :) I do not want to force others to rewrite existing handlers and force all handlers to be ActiveJob::Base ancestors.
However your solution is nice - it allows simple change between sync & async handlers (of course ignoring all issues that might be in the handler code) by just changing job adapter 👍

mpraglowski:
BTW I want to avoid code like this:

`es.subscribe(instance_of(Denormalizers::OrderSubmitted), [Events::OrderSubmitted])` source from https://github.com/mpraglowski/cqrs-es-sample-with-res/blob/master/config/initializers/event_store.rb

It could be replaced by simple:

`es.on([Events::OrderSubmitted], Denormalizers::OrderSubmitted))`

mpraglowski:
ehh wrong button - GH UX sucks sometimes ;)

mpraglowski:
> Question: does it deserialize events or is it up to the developer to do that

Currently it is YAML, however I want to work on that - it should be a RES::Event here.

pawelpacana:
> currently we have subscribe(lambda_or_callable_class_instance, [domain_event], unsubscribe_proc), this will not allow us to use async jobs as we need to pass instance to RES client

Reminder that lambda can only be sync handler - async would require handler to de defined as a class that can be loaded in sidekiq (or whatever else in different process).  

https://stackoverflow.com/questions/23970/how-do-i-marshal-a-lambda-proc-in-ruby

mlomnicki:
@mpraglowski  The code itself looks good to me :+1:

I have some doubts when it comes to the API though. 

Why this way?

```ruby
def  on(event_types, subscriber_klass, &proc)
```

rather than

```ruby
def  on(subscriber_klass, event_types, &proc)
```

I mean compare these two

```ruby
es.on(OrderDenormalizer, [
  OrderSubmitted,
  OrderCancelled,
  OrderAccepted,
  OrderVerified,
  OrderCompleted
])

# vs

es.on([
  OrderSubmitted,
  OrderCancelled,
  OrderAccepted,
  OrderVerified,
  OrderCompleted
], OrderDenormalizer)
```

IMO the former reads much better. I realize this opinion is highly subjective though...

Next, if we agree that the former is better then do we actually need the `#on` API if we already have `#subscribe`?

I think `subscribe` is preffered because it nicely matches with `publish` and emphasizes the PubSub pattern.

I realize it will make the subscribe method slightly more complex to keep backwards compatibility but I would choose it over the new API method anyway. WDYT?

mpraglowski:
> Why this way?

because `def  on(event_types, subscriber_klass, &proc)` reads better :) 

`on` some `event_types` do call that `subscriber_klass`

> Next, if we agree that the former is better then do we actually need the #on API if we already have #subscribe?

`#subscribe` have different meaning - it always expects callable instance (not class) (could be Proc). It is for me currently more "low level" API. `#on` uses `#subscribe` internally to subscribe the generated proxy for the events.

> I think subscribe is preffered because it nicely matches with publish and emphasizes the PubSub pattern.

Yeah - would be nice to have methods overloading ;) And I would like to keep current API and do not change `#subscribe` behaviour.

> I realize it will make the subscribe method slightly more complex to keep backwards compatibility but I would choose it over the new API method anyway. WDYT?

To break the API or not to break :) That is the question ;)

mlomnicki:
> because def on(event_types, subscriber_klass, &proc) reads better :)

I expected you'll say that :) Apparently we need more input. Anyone?

> To break the API or not to break :) That is the question ;)

I'd say don't break unless absolutely necessary. How about
```ruby
def subscribe(handler, events)
  if handler.responds_to?(:call)
    event_broker.add_subscriber(....)
  else
    proxy = get_proxy_for(handler)
    event_broker.add_subscriber(proxy, ...)
  end
end
```

but still we'd have to agree on whether (handler, events) is better than (events, handler)

mpraglowski:
hmm I like the conditional `subscribe` - that will preserve current API and allow to pass the class as a handler

> but still we'd have to agree on whether (handler, events) is better than (events, handler)

in that case we do not have that problem - we will keep existing API so `(handler, events)` is the winner ;)

mlomnicki:
> (handler, events) is the winner ;)

🎉 🍾

pawelpacana:
Btw. regarding https://github.com/RailsEventStore/rails_event_store/pull/103#issuecomment-326519988, that code:

```
def instance_of(klass, *args)
  ->(event) { klass.new(*args).call(event) }
end

Rails.application.config.event_store.tap do |es|
  es.subscribe(instance_of(Denormalizers::OrderSubmitted), [Events::OrderSubmitted])
end
```

can be written as
```
Rails.application.config.event_store.tap do |es|
  es.subscribe(Denormalizers::OrderSubmitted.new, [Events::OrderSubmitted])
end
```

as both result of `instance_of` function and `Denormalizers::OrderSubmitted` instance respond to `#call` with an argument on an `event`.

mlomnicki:
> can be written as

> Rails.application.config.event_store.tap do |es|
>    es.subscribe(Denormalizers::OrderSubmitted.new, [Events::OrderSubmitted])
> end

Right. Also the title is a bit misleading because Async handlers are possible even without this patch

```ruby
es.subscribe(->(e) { MyJob.perform_async(e.to_hash) }, [Events::OrderSubmitted])
```
or
```ruby
es.subscribe(MyJob, [Events::OrderSubmitted])

class MyJob
   include Sidekiq::Worker

   def self.call(event)
      self.class.perform_async(EventSerializer.serialize(event))
   end

   def perform(serialized_event)
      event = EventSerializer.deserialize(serialized_event)
      # do something
  end
end
```

Regardless of that, the PR is a step in the right direction so thumbs up

paneq:
@mpraglowski Why is it by default using `after_commit` strategy?

It makes sense for non-SQL adapters but it makes things worse for SQL-based queues.

I strongly believe we need to support two approaches

* async
* async after commit

At the beginning it would be better to support them manually and maybe later make it smarter with some detection.

Also:

```
 +        if ActiveRecord::Base.connection.transaction_open?
 +          ActiveRecord::Base.
 +            connection.
 +            current_transaction.
 +            add_record(AsyncRecord.new(klass, e))
```

From my experience this can get more complicated in rails testing environment

mpraglowski:
Good point - not all RES instances must use RESAR as event repository.

paneq:
> not all RES instances must use RESAR as event repository.

Actually the usecase where _async after commit_ is not optimal is exactly when they use RESAR and SQL based queue as well.

mpraglowski:
Yep - understood. _async after commit_ does not make sense when you are in the same transaction anyway - in that case it is better to enqueue inside transaction.

paneq:
@mpraglowski by reading the new commits I don't see easily what's the API for the developers to choose between async and async_after_commit

mpraglowski:
@paneq it is very explicit - here is a RES initialization with that setup:

```ruby
config.event_store = RailsEventStore::Client.new(
  event_broker: RailsEventStore::EventBroker.new(
    dispatcher: RailsEventStore::ActiveJobDispatcher.new(
      proxy_strategy: RailsEventStore::AsyncProxyStrategy::AfterCommit.new
    )
  )
)
```

paneq:
@mpraglowski so it's a global strategy? I always though we need this feature per handler. But maybe we don't...

mpraglowski:
Dispatcher is global. Handler is the place when you decide if it is sync (a regular class with `call` method implemented) or async (class inherited from ActiveJob).

paneq:
@mpraglowski I understand, I was wondering if you sometimes need both, async for some handlers and async after commit for others but my guess is that you don't so that's OK :)

mpraglowski:
@paneq we could improve it later and choose proxy strategy based on ActiveJob queue adapter but I would like to cut the scope for now and have it merged soon. BTW here https://github.com/mpraglowski/cqrs-es-sample-with-res I've updated my sample app to have async handlers.

rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    private
+    def async_handler?(klass)
+      klass < ActiveJob::Base
+    end
+
+    def sync_proxy(klass)
+      raise InvalidHandler.new(klass) unless klass.method_defined?(:call)
+      ->(e) { klass.new.call(e) }
```
pawelpacana: Returning `klass.new` is enough -- we're sure it responds to `#call`

rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
```diff
@@ -0,0 +1,64 @@
+require 'active_job'
+
+module RailsEventStore
+  module AsyncProxyStrategy
+    class AfterCommit
+      def call(klass, event)
+        if ActiveRecord::Base.connection.transaction_open?
```
pawelpacana: That woudln't work with different adapter, right?

mpraglowski: yep - but this strategy is valid only if you are using ActiveRecord & RESAR event repository

rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
```diff
@@ -0,0 +1,64 @@
...
+      @async_proxy_strategy = proxy_strategy
+    end
+
+    def call(subscriber, event)
+      subscriber.call(event)
+    end
+
+    def proxy_for(klass)
+      async_handler?(klass) ? async_proxy(klass) : sync_proxy(klass)
+    end
```
pawelpacana: Wouldn't it be a simpler (in terms of code and understanding) to decide in `Dispatcher#call` whether `subscriber` needs to wrapped in proxy?

pawelpacana: Right now it happens in event broker. Broker checks if proxy is needed and delegates to dispatcher to wrap. Dispatcher returnes wrapped subscriber and then receives it in `#call`.

I could see that current implementation results in fewer calls but it took me a while to grasp it. Is performance on this code path significant? (benchmark neeeded)

pawelpacana:
> I strongly believe we need to support two approaches
> 
> async
> async after commit

It would be greate to explain this in docs:
- _ async after commit_ for Sidekig and Resque like
- _ async_ for DelayedJob like

rails_event_store/spec/active_job_dispatcher_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    it "builds sync proxy for callable class" do
+      handler = ActiveJobDispatcher.new.proxy_for(CallableHandler)
+      expect(handler.respond_to?(:call)).to be_truthy
+      expect_any_instance_of(CallableHandler).to receive(:call).and_call_original
+      expect_no_enqueued_job do
+        handler.call(event)
+      end
+    end
+
+    it "error when ActiveJob::Base is used as handler class" do
```
pawelpacana: This should make no difference for any class without `#call`, right?

ruby_event_store/lib/ruby_event_store/pub_sub/dispatcher.rb
```diff
@@ -4,6 +4,11 @@ class Dispatcher
       def call(subscriber, event)
         subscriber.call(event)
       end
+
+      def proxy_for(klass)
+        raise InvalidHandler.new(klass) unless klass.method_defined?(:call)
+        ->(e) { klass.new.call(e) }
```
pawelpacana: same thing as previously

pawelpacana:
Merging -- technically speaking it is green.  Follow-up refactor will happen based on master branch.

Issue #109, opened 2017-09-14
Decide what should be returned by AggregateRoot#apply


paneq:
Current code

```
module AggregateRoot
  def apply(event)
    apply_strategy.(self, event)
    unpublished_events << event
  end
```

However there is no test checking for it and manual mutation testing still passed all tests:

```
module AggregateRoot
  def apply(event)
    apply_strategy.(self, event)
    unpublished_events << event
    nil
  end
```

Returns all `unpublished_events` might sound good but if someone wants to test events created by calling a command and set up a state using other commands, it's not easy to extract domain events created only by last action. Example:

```
class Product
  def supply(quantity)
    raise NotRegistered unless @store_id
    apply(ProductSupplied.new(data: {
      store_id: @store_id,
      sku: @sku,
      quantity: quantity,
    }))
  end
end

  specify "product supplied" do
    product.register(store_id: 1, sku: "BOOK")
    product.supply(30)
```

Since `product.unpublished_events` is private (#108) and `product.supply` returns all `unpublished_events` there is no easy way to test only events generated by `product.supply`.

My proposal:

```
module AggregateRoot
  def apply(*events)
    events.each do |event|
      apply_strategy.(self, event)
      unpublished_events << event
    end
  end
```

* allow applying multiple events (useful sometimes)
* return only applied events instead of all `unpublished_events`

PR #110, opened 2017-09-14
apply returns applied events

paneq:
that makes testing much easier when commands are used to build state.
Developers can easily opt-out by doing

```ruby
def apply(*events)
  super
  nil
end
```

in their class.

or by explicitly returning `nil` or `self`

```ruby
  def supply(quantity)
    raise NotRegistered unless @store_id
    apply(ProductSupplied.new(data: {
      store_id: @store_id,
      sku: @sku,
      quantity: quantity,
    }))
    nil
  end
```

### DIFF
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -4,9 +4,11 @@ require 'aggregate_root/configuration'
 require 'aggregate_root/default_apply_strategy'
 
 module AggregateRoot
-  def apply(event)
-    apply_strategy.(self, event)
-    unpublished_events << event
+  def apply(*events)
+    events.each do |event|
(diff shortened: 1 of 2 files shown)


paneq:
Related to https://github.com/RailsEventStore/rails_event_store/issues/109

pawelpacana:
Seems like a step in good direction without causing much trouble.

mlomnicki:
I think `apply` should be private although that's a separate issue.

Overall this idea is an improvement to what we have currently so +1

andrzejkrzywda:
A good change. Should the README/release_notes change be part of the PR, given that we change the public API?

paneq:
@andrzejkrzywda yep! Will do when we release.

PR #114, opened 2017-09-18
Make unpublished_events public and read-only

paneq:
Instead of https://github.com/RailsEventStore/rails_event_store/pull/108

### DIFF
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -7,7 +7,7 @@ module AggregateRoot
   def apply(*events)
     events.each do |event|
       apply_strategy.(self, event)
-      unpublished_events << event
+      unpublished << event
     end
   end
 
@@ -28,10 +28,14 @@ module AggregateRoot
     @unpublished_events = nil
   end
 
+  def unpublished_events
+    unpublished.each
+  end
+
   private
   attr_reader :loaded_from_stream_name
 
-  def unpublished_events
+  def unpublished
     @unpublished_events ||= []
   end
 
--- a/aggregate_root/spec/aggregate_root_spec.rb
+++ b/aggregate_root/spec/aggregate_root_spec.rb
@@ -10,12 +10,12 @@ describe AggregateRoot do
     expect(order).to receive(:apply_order_created).with(order_created).and_call_original
     order.apply(order_created)
     expect(order.status).to eq :created
-    expect(order.expected_events).to eq([order_created])
+    expect(order.unpublished_events.to_a).to eq([order_created])
   end
 
   it "brand new aggregate does not have any unpublished events" do
     order = Order.new
-    expect(order.expected_events).to be_empty
+    expect(order.unpublished_events.to_a).to be_empty
   end
 
   it "should have no unpublished events when loaded" do
@@ -25,7 +25,7 @@ describe AggregateRoot do
 
     order = Order.new.load(stream, event_store: event_store)
     expect(order.status).to eq :created
-    expect(order.expected_events).to be_empty
+    expect(order.unpublished_events.to_a).to be_empty
   end
 
   it "should publish all unpublished events on store" do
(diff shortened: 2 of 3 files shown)


paneq:
Review please @andrzejkrzywda @mpraglowski @andrzejsliwa

aggregate_root/lib/aggregate_root.rb
```diff
@@ -7,7 +7,7 @@ module AggregateRoot
   def apply(*events)
     events.each do |event|
       apply_strategy.(self, event)
-      unpublished_events << event
+      unpublished << event
```
andrzejkrzywda: totally minor, but the previous name was more explicit. might be useful for new contributors

aggregate_root/spec/aggregate_root_spec.rb
```diff
@@ -10,12 +10,12 @@
     expect(order).to receive(:apply_order_created).with(order_created).and_call_original
     order.apply(order_created)
     expect(order.status).to eq :created
-    expect(order.expected_events).to eq([order_created])
+    expect(order.unpublished_events.to_a).to eq([order_created])
```
andrzejkrzywda: it seems to be annoying that (tests show that here), that we need to add `to_a` always with the solution suggested.

andrzejkrzywda:
At first, I like the idea of returning the iterator, but now I see the downsides with the `to_a`. I think the solution with returning a list clone would be simpler here. Did it have any drawbacks?

aggregate_root/lib/aggregate_root.rb
paneq: We now have public api `unpublished_events` which is read_only and private API `unpublished` which does `||=` when needed. I don't know what previous name you have in mind. We can't use `unpublished_events` anymore because it is read only

aggregate_root/spec/aggregate_root_spec.rb
paneq: You can use `each`, `map` without adding `to_a`. You only need `to_a` when you want an Array which makes sense to me.

paneq:
>  a list clone would be simpler here

how is `array.clone` exactly simpler than `array.each` ?

aggregate_root/spec/aggregate_root_spec.rb
andrzejsliwa: I think that name is messing up with expectation, looking on plural form you are expecting to have arrays or array like ;)

aggregate_root/lib/aggregate_root.rb
andrzejkrzywda: Can we use `@unpublished_events` for internal use, and `unpublished_events` for the public accessor? I think that would be idiomatic Ruby to solve this kind of a problem.

andrzejkrzywda:
The simplicity comes from not having to call `to_a` by the clients of this code in case of list-based assertions. I happen to use the old version of AggregateRoot, which still had the `unpublished_events` public and it is convenient that it returns a list, so we can compare with `[MyEvent]`.

pawelpacana:
> I happen to use the old version of AggregateRoot, which still had the unpublished_events public and it is convenient that it returns a list, so we can compare with [MyEvent]

Maybe `assert_unpublished_events` is the test assertion you're looking for?

andrzejkrzywda:
Yeah, `assert_unpublished_events` would be a possible solution.

aggregate_root/lib/aggregate_root.rb
paneq: We could be we would need to repeat the `||=` logic. I don't see anything un-idiomatic with current way.

aggregate_root/spec/aggregate_root_spec.rb
paneq: And an iterator is array-like.

paneq:
It was deemed good enough for merge.

paneq:
Draft of release updated.

PR #125, opened 2017-09-27
Proxy for refactoring

paneq:


paneq:
Was:

 of `Client` figuring out whether a subscriber is a `Class` or not and based on that calling `proxy_for` to Broker which delegates back to dispatcher, which wraps the logic inside a proc, returns it back higher, so that the client can subscribe this proxy proc in a broker... 

After changes:

* allow subscribing anything
* as long as the dispatcher knows how to handle it

Right now I used inheritance but perhaps dispatchers should be a pipeline/middlewares pattern in the future. If one dispatcher does not know how to handle dispatching to a certain subscriber maybe next in line knows and if not maybe another one would know until the last one.

I am not convinced static verification during a subscription is useful. I would consider removing #verify as well. This is ruby, I would let it fail dynamically. I kept it because we had basic verification so far.

Additional benefit of that approach is that we have the original subscriber in the list of subscribers and we could get their names etc and display in a rake task.

I know the tests are failing right now, I just wanted to make sure first that we agree that this direction makes sense.

Please review @mpraglowski @mlomnicki @andrzejkrzywda

mpraglowski:
> Right now I used inheritance but perhaps dispatchers should be a pipeline/middlewares pattern in the future. If one dispatcher does not know how to handle dispatching to a certain subscriber maybe next in line knows and if not maybe another one would know until the last one.

I like that approach. Maybe it would be nice to have it not only for dispatchers but as overall guideline for adding new plugins i.e. serialization of events.

mpraglowski:
> I am not convinced static verification during a subscription is useful. I would consider removing #verify as well. I kept it because we had basic verifications so far.

That's just a definition of moment when we want to report failure:
* verify on subscribe - let dev now ASAP before any event is published - usually on app initialisation
* verify on call - just when some event is published - at runtime

When on `#call` then we need to define how to handle error here https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/pub_sub/broker.rb#L24 - because some subscribers have been notified already and some might be still waiting to be notified. Maybe that's not the issue as we have the same with failing sync handlers.

paneq:
```
   when :auto
-    eis = EventInStream.where(stream: stream_name).order("position DESC").first
+    eis = EventInStream.where(stream: stream_name).order(nil).first
```

Shit, how did it start failing now like that...

paneq:
selection_622

a) It seems we have a random.
b) Why are we building everything twice?

Issue #132, opened 2017-10-04
AsyncRecord in AsyncProxyStrategy has incomplete API


jakubkosinski:
In order to make `AsyncRecord` work with AR, it should respond to more methods:
- `rolledback!`
- `before_committed!`

They can just return `nil` but should be defined.

* [x] `v0.18.2`
* [x] `master`

paneq:
@jakubkosinski which rails version are we talking about?

paneq:
I was able to reproduce on Rails 5.1.4

```
(fix-missing-methods) > RAILS_VERSION=5.1.4 bundle exec rspec
5
1
...before_committed!
committed!
.rolledback!
........................
```

the reason it was failing for you is that we were not testing on this rails version so far. The tests in general were going through this path in code.

PR #133, opened 2017-10-04
Add missing methods to AsyncRecord

mpraglowski:
When transaction is rolled back AR calls `rollback!` on all records
added to transaction. The methods, even withou aby body, need to be
defined to avoid missing method error.

Fix for https://github.com/RailsEventStore/rails_event_store/issues/132


paneq:
Interesting consideration.

It's hard to mutation test methods which are only needed for some Rails versions:

```
        def rolledback!(**_)
        end

        def before_committed!
        end
```

So I am gonna exclude them from mutation testing.

It's ok because if someone removes them. The tests should fail on certain Rails versions.

Issue #131, opened 2017-10-04
`all` vs `__global__`


paneq:
- [x] Change `__global__` to `all` in RES_active_record
- [x] Make publishing to `all` work properly (1 record instead of 2 records)
- [x] Keep `all` as higher level detail
- [x] Change the semantic that reading from `all` stream directly will read data written into other streams.
- [x] Change migration to use `all` #129

TLDR: Expose magic `all` stream.

mpraglowski:
Why at all we need to have `__global__` or `all` stream? Isn't the "global" stream a list of all events stored in `event_store_events` with `position`  attribute and without any entry in `event_store_events_in_streams`. Any read/write to `RailsEventStore::GLOBAL_STREAM` should not care about `event_store_events_in_streams`.

> Change the semantic that reading from all stream directly will read data written into other streams.

I think I do not understand this :(

paneq:
@mpraglowski 

> Why at all we need to have `__global__` or `all` stream?

Maybe we don't. That's who I approached it.

>  Isn't the "global" stream a list of all events stored in event_store_events with position attribute and without any entry in event_store_events_in_streams

Perhaps it could be. That would require wrapping saving event and event in streams in a single transaction but maybe that's an interesting design.

> Any read/write to RailsEventStore::GLOBAL_STREAM should not care about event_store_events_in_streams.

I see your vision. My vision was to keep `__global__` implemented in an identical way as all other streams. Thanks for sharing your vision.

mpraglowski:
> I see your vision. My vision was to keep __global__ implemented in an identical way as all other streams. Thanks for sharing your vision.

I like that part about having identical way as all other streams. This is my biggest concern with the way I described it here, it will need some `if` in each method - probably not worth it.

pawelpacana:
Just a thought:

1. let's assume we don't have `__global__` anymore and there is `all` stream with changed semantic (that all events belong to it regardless if an event was published in another named stream or not)
2. we already have an API for reading `read_all_streams_forward` and `read_all_streams_backward`; there is theoretically no need to expose the fact that we have a stream called `all` (i.e. to use it as an argument in `read_events_forward` and `read_events_backward`)
3. all events in `all` stream have no particular position, we determine the order by autoincremented `id` of `event_store_events_in_streams` table

What if:

1. we remove any reference to `GLOBAL_STREAM` in code, resulting in no exposure of `all` stream anymore
[...]

Issue #129, opened 2017-10-02
Migration to new DB schema


paneq:
* [ ] Verify data migrated
  * [x] positions
  * [ ] created_at
* [x] Check on all DBs
  * [x] postgres
  * [x] mysql
  * [x] sqlite
* [x] Verify proper schema at the end
* [x] v1-v2 migration generator by calling `rake res:v1tov2schema_migration`
* [x] Fix random with SQL

paneq:
Should we have a new gem for upgrades that would just have this migration?

I think that ideal steps to verify correctness would include

* filling DB with data via older gem write API
* install new gem version
* run migration
* run queries against API using newer read API and make sure they give the same results
* write new data and make sure everything works correctly

Sounds pretty much complicated

paneq:
@mlomnicki any ideas?

paneq:
Alternatively, we ship rails_event_store_active_record_single_table adapter which still handles race conditions incorrectly but does not require migrating schema of events.

paneq:
It's not writing to `all` or `__global__` ... Damn.

pawelpacana:
> Alternatively, we ship rails_event_store_active_record_single_table adapter which still handles race conditions incorrectly but does not require migrating schema of events.

I think that would be interesting option -- it would give some time to prepare to migration. We could issue deprecation warning that you still use old schema and that there's new one.

pawelpacana:
```ruby
# This migration is not intended for live migration
# It assumes no data is added at that time.
# Make sure you have a backup before running on production

# 10_000_000 distinct stream names
# can cause around 2GB of usage
# make sure you can run this migration on your production system
streams = {}

RailsEventStoreActiveRecord::Event.find_each do |ev|
  position = nil
  if preserve_positions?(ev.stream)
    streams[ev.stream] ||= -1
    position = streams[ev.stream] += 1
  end

  RailsEventStoreActiveRecord::EventInStream.create!(
    stream: "all",
    event_id: ev.id,
    created_at: ev.created_at,
  )

  unless ev.stream == "all"
    RailsEventStoreActiveRecord::EventInStream.create!(
      stream: ev.stream,
      position: position,
      event_id: ev.id,
      created_at: ev.created_at,
    ) 
  end
end

add_index :event_store_events_in_streams, [:stream, :position], unique: true
add_index :event_store_events_in_streams, [:stream, :event_id], unique: true
add_index :event_store_events_in_streams, [:created_at]

[...]

PR #138, opened 2017-10-11
Remove redundant "__global__" stream, slightly changing "all" semantics

pawelpacana:
- addresses https://github.com/RailsEventStore/rails_event_store/issues/131
- no changes to migration in https://github.com/RailsEventStore/rails_event_store/issues/129, it might be slightly different if we decide on https://github.com/RailsEventStore/rails_event_store/issues/131#issuecomment-335634991


paneq:
@pawelpacana 

Since the global stream is unordered (on writes we can't expect certain position) we should raise an exception if:

```ruby
if stream_name == RubyEventStore::GLOBAL_STREAM && expected_version != :any
```

and have a test for it.

pawelpacana:
@paneq added 3 test cases, each with `expected_version` different that `:any`. Mergin and moving forward :)

PR #149, opened 2017-10-24
Allow to control default apply strategy strict mode via configuration

mpraglowski:
Because sometimes you don't care about missing handlers and you do not
want to add apply strategy to each aggregate. No breaking changes.
Current behaviour is preserved.


mpraglowski:
This is not valid approach. Whole `AggregateRoot.configuration` should be removed.

Issue #148, opened 2017-10-19
Do not persist event when EventInStream persisting failed


paneq:
At the beginning I thought it won't matter but it could matter for _change data capture_ softwares. It's better not to persist events if they don't land in any stream.

pawelpacana:
PoC in https://github.com/RailsEventStore/rails_event_store/pull/141

PR #156, opened 2017-10-30
Events not persisted if stream append failed

paneq:
Fixes: #148


paneq:
Already reviewed by @pawelpacana :)

PR #165, opened 2017-11-16
initialize rails engine for our views and clone browser

antonpaisov:


antonpaisov:
we did it in master already.

Issue #154, opened 2017-10-27
Custom serializers/deserializers aka mappers


paneq:
I think with custom serialization/deserialization strategy we could solve a few problems.

a) performance - if you want faster RES just serialize using different format
b) event's interface. If you want to use structs or hashes that should be ok, as long as we know how to deserialize it and serialize it.

I believe we need:

* event_id (string)
* event_type (string)
* data (string/binary)
* metadata (string/binary)

Related: #27 #50

paneq:
Good point:

> ruby_event_store injects a timestamp into the metadata hash 

Response:

>This is a timestamp of event publication. It is similar to you go to post some message at the post office, you handle your message (letter in envelope) that is "frozen" - but the post office clerk will hit the envelope with stamp adding a time when it has been handled. This is exactly what is RubyES doing.
> And it could not be set earlier because the moment of event creation is not the same as the moment of publication.

I don't think we can assume `metadata` is a `Hash`

If we want flexibility in defining events then the client could use a closed struct which does not allow populating it with any kind of data...

On the other hand we also have request metadata that we populate... We must somehow do it before the event is serialized. So it seems there are 3 stages

1. an event class is instantiated and filled with data by a developer
2. the event's metadata is enriched by RES
3. the event is serialized by a mapper (YAML by default, other can be provided)
4. the event is stored by RES

This feature is about introducing an explicit 3rd stage of the process. I am not sure yet how we can do it without introducing an interface for the 2nd stage as well.

BTW. I think timestamp should be a separate attribute, and not in metadata. We only know the timestamp once the event is actually persisted `RailsEventStoreActiveRecord::Event#created_at`. Also... An event in the future could be persisted in different streams at different times.

paneq:
> I believe we need:
> event_id (string)
> event_type (string)
> data (string/binary)
> metadata (string/binary)

This is in terms of what needs to be returned to us by a serializer. Not what an event needs to define.

The point of the serializer/deserializer (aka mapper) is to decouple the event format from what needs to be persisted somehow.

paneq:
I think that I come to an understanding:

* the event's metadata is enriched by RES
  * right now this is implemented in RubyEventStore. I believe this should be moved up to RailsEventStore

    ```ruby
    metadata[:timestamp] ||= clock.()
    metadata.merge!(metadata_proc.call || {}) if metadata_proc
    ```
[...]

PR #168, opened 2017-11-21
Events mappers / serialisers proposal

andrzejsliwa:
Related issue: #154

Goal:
- Lets customise the serialisation and factorisation of the events.
- Switch from ActiveRecord declarative serialisation to explicit, configured version.
- Keep backward compatibility.
 
Extra:
As proof of concept `WithEventClassRemapping` mapper was added to handle example use case of events classes "translation" on the fly (To keep store immutable, in case of renaming/moving events classes)


rails_event_store_active_record/lib/rails_event_store_active_record.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_event_store_active_record/generators/migration_generator'
 require 'rails_event_store_active_record/generators/v1_v2_migration_generator'
 require 'rails_event_store_active_record/event'
+require 'rails_event_store_active_record/mappers/yaml'
+require 'rails_event_store_active_record/mappers/yaml_with_event_class_remapping'
```
andrzejsliwa: it's added just as example, we can maintain it on private code base side, but was neccessary to show reasoning of implementation (to keep flexibility for future use cases)

andrzejsliwa: If you would like to merge it, I will provide necessary specs to cover it

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize(serializer: ::YAML)
+        @serializer = serializer
+      end
+
+      def create_event(event)
+        Event.create!(
```
andrzejsliwa: create!/persistance can be extracted here, back to repository (to return hash)... but this will reduce flexibility

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize(serializer: nil)
+        @serializer = serializer || ::YAML
+      end
```
andrzejsliwa: I keep DI here, to separate testing of serialiser and mapper (with YAML we are making assumption that its well tested, for mapper we will verify if he "mapping" it correctly with using serialiser)

andrzejsliwa:
I made change on level of RailsEventStoreActiveRecord::EventRepository, because its best place in my opinion, for InMemoryRepository on RubyEventStore there is no need of customising of serialization.

I still have problem about the naming of it (mappers, serializers, etc... we have already serializer, so I didn't make serialiser for serialiser)

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      def build_event_payload(event)
```
paneq: Things I would consider changing:

* method name
* Using an object (instead of open Hash) or ClosedStruct
  * or using 4 methods such as `event_id(event)`, `binary_data(event)`, etc

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
...
+        {
+          id:         event.event_id,
+          data:       serializer.dump(event.data),
+          metadata:   serializer.dump(event.metadata),
+          event_type: event.class,
+        }
+      end
+
+      def build_event_entity(record)
+        return nil unless record
```
paneq: when would that happen? why would we prefer `nil` over exception?

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
...
+      def build_event_payload(event)
+        {
+          id:         event.event_id,
+          data:       serializer.dump(event.data),
+          metadata:   serializer.dump(event.metadata),
+          event_type: event.class,
+        }
+      end
+
+      def build_event_entity(record)
```
paneq: `build_event_entity` - an event is not an entity

andrzejsliwa: name and responsibility its extracted from repository

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: its extracted from repository

andrzejsliwa: I didn't change it because I also don't know the answer

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
paneq: I don't think that's necessary. A YAML mapper should use YAML internally, I see no benefit in passing `YAML` as dependency.

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: any suggestion about method names are welcome,

about object creation, which way you prefer? I'm open for suggestion

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: 👍 , I will remove it

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: I'm open for name proposal

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
paneq: let's drop it :)

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,31 @@
...
+        @serializer = ::YAML
+      end
+
+      def build_event_payload(event)
+        {
+          id:         event.event_id,
+          data:       serializer.dump(event.data),
+          metadata:   serializer.dump(event.metadata),
+          event_type: event.class,
+        }
```
andrzejsliwa: I would keep here hash, because it what `Event.create!` is accepting, even if I will make here close struct or object, it will be casted to_h. On begining I created event record here, but moved it back to repository

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: we can't

this is the use case you asked for
```ruby
  1) RailsEventStoreActiveRecord::EventRepository behaves like event_repository knows last event in stream
     Failure/Error:
       record.event.event_type.constantize.new(
         event_id: record.event.id,
         metadata: serializer.load(record.event.metadata),
         data:     serializer.load(record.event.data)
       )
     
     NoMethodError:
       undefined method `event' for nil:NilClass
     Shared Example Group: :event_repository called from ./spec/event_repository_spec.rb:24
```

andrzejsliwa: eventually I can move this check up, back to repository

andrzejsliwa: done

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
paneq: @andrzejsliwa nah, you can assign one-by-one to ActiveRecord `Event`:

```ruby
def event_to_db(event)
  db_event = Event.new
  db_event.id = mapper.get_id(event)
  ...
end

or one by one to a hash in this persistence layer, but I would prefer keeping `Hash` out of this mapper layer.

**Comment by andrzejsliwa** [comment]
ok, then I will build new Event instance

File: rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb (hunk 7 above)
Comment (andrzejsliwa) [reply rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb]: @paneq take a look again

File: rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb [hunk 8]
```diff
@@ -134,6 +134,7 @@ def detect_pkey_index_violated(e)
     end
 
     def build_event_entity(record)
+      return nil unless record
```
paneq: I would just drop this check at all.

andrzejsliwa:
thens spec is failling like I pasted in previous comments

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -137,11 +135,7 @@ def detect_pkey_index_violated(e)
 
     def build_event_entity(record)
       return nil unless record
```
andrzejsliwa: removing of this check is resulting:

Failures:

  1) RailsEventStoreActiveRecord::EventRepository behaves like event_repository knows last event in stream
     Failure/Error:
       record.event.event_type.constantize.new(
         event_id: record.event.id,
         metadata: serializer.load(record.event.metadata),
         data:     serializer.load(record.event.data)
       )
     
     NoMethodError:
       undefined method `event' for nil:NilClass
     Shared Example Group: :event_repository called from ./spec/event_repository_spec.rb:24

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,30 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize
+        @serializer = ::YAML
+      end
+
+      def build_event_record(event)
```
andrzejsliwa: @paneq would you rename it to `event_to_record`, and `record_to_event`

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: Then we need to fix

```
    def last_stream_event(stream_name)
      build_event_entity(
        EventInStream.where(stream: stream_name).order('position DESC, id DESC').first
      )
    end
```

to handle `EventInStream.where(stream: stream_name).order('position DESC, id DESC').first` being `nil` to not call `build_event_entity`

andrzejsliwa: this will be just moving condition again UP :) right?

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
paneq: Better. I think I am in favor of 4 smaller methods to serialize and one (with 4 args) to deserialize, or we should introduce a closed object that is not Hash :) which would serve as transport layer between those 2 parts.

paneq: I can't decide which one is better.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
andrzejsliwa: done

paneq: Yep, but it is only needed in one place (I believe) so let's keep it close to that one place only.

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: I think with creating insteance of Event the contract is more cleaner now with `event_to_record`, and `record_to_event`

andrzejsliwa: @paneq please look again on code base, now looks cleaner and explaining behaviour in symmetric way

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,30 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize
+        @serializer = ::YAML
+      end
+
+      def event_to_record(event)
+        Event.new(
```
paneq: @andrzejsliwa nah, the mapper cannot assume the persistence layer. It might get stored in mongo. That's why I said we shold go with 4 separate attributes or 1 intermediary

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,30 @@
...
+        )
+      end
+
+      def record_to_event(record)
+        record.event.event_type.constantize.new(
+          event_id: record.event.id,
+          metadata: serializer.load(record.event.metadata),
+          data:     serializer.load(record.event.data)
+        )
+      end
```
paneq: similarly here, why is it doing `record.event` ? Because it is coupled to persistence layer.

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
andrzejsliwa: c'mon its implemented on specific repository, so can assume it for sure

paneq: Ah.... I did not notice. It should be part of `ruby_event_store`, not `raills_event_store_active_record`.

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,30 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize
+        @serializer = ::YAML
+      end
+
+      def event_to_serialized_record(event)
```
paneq: I would rename `event` to `domain_event`

rails_event_store_active_record/lib/rails_event_store_active_record/mappers/yaml.rb
```diff
@@ -0,0 +1,30 @@
+module RailsEventStoreActiveRecord
+  module Mappers
+    class YAML
+      def initialize
+        @serializer = ::YAML
```
paneq: I would just use `::YAML` constant in code instead of `@serializer`

ruby_event_store/lib/ruby_event_store/serialized_record.rb
```diff
@@ -0,0 +1,12 @@
+module RubyEventStore
+  class SerializedRecord
+    def initialize(id:, data:, metadata:, event_type:)
+      @id         = id
+      @data       = data
+      @metadata   = metadata
+      @event_type = event_type
+    end
```
paneq: what do you think about doing `freeze` here?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -69,9 +65,9 @@ def has_event?(event_id)
     end
 
     def last_stream_event(stream_name)
-      build_event_entity(
-        EventInStream.where(stream: stream_name).order('position DESC, id DESC').first
-      )
+      record = EventInStream.where(stream: stream_name).order('position DESC, id DESC').first
+      return nil unless record
+      build_event_instance(record)
```
paneq:
```
return nil unless record
build_event_instance(record)
```

I think that could be `record && build_event_instance(record)`

:) but that's a minor

ruby_event_store/ruby_event_store.gemspec
```diff
@@ -19,6 +19,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
+
+  spec.add_development_dependency 'activesupport'
```
paneq: @andrzejsliwa it's a runtime dependency, not a development.

ruby_event_store/lib/ruby_event_store/serialized_record.rb
```diff
@@ -0,0 +1,13 @@
+module RubyEventStore
+  class SerializedRecord
+    def initialize(id:, data:, metadata:, event_type:)
+      @id         = id
```
paneq: Do you think it would be better to rename it to `event_id` as higher layer use (or stay with `id` as lower layer uses?) I think `event_id` could be better.

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    RSpec.describe Default do
+      let(:data)   {{some_attribute: 5}}
+      let(:metadata)     {{some_meta: 1}}
+      let(:domain_event) { SomethingHappened.new(data: data, metadata: metadata) }
+
+      specify '#event_to_serialized_record returns YAML serialized record' do
+        record = subject.event_to_serialized_record(domain_event)
+        expect(record).to            be_a SerializedRecord
+        expect(record.id).to         eq domain_event.event_id
+        expect(record.metadata).to   eq YAML.dump(metadata)
```
paneq: This test is not checking anything, you are comparing 2 identical code path which do `YAML.dump(metadata)`. Please make it a static string, provided as it is.

ruby_event_store/lib/ruby_event_store/mappers/default.rb
```diff
@@ -0,0 +1,35 @@
...
+        @serializer = serializer
+        @events_class_remapping = events_class_remapping
+      end
+
+      def event_to_serialized_record(domain_event)
+        SerializedRecord.new(
+          id:         domain_event.event_id,
+          metadata:   serializer.dump(domain_event.metadata),
+          data:       serializer.dump(domain_event.data),
+          event_type: domain_event.class
```
paneq: That should be a `String`. Very important!

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
+SomethingHappened = Class.new(RubyEventStore::Event)
+
+module RubyEventStore
+  module Mappers
+    RSpec.describe Default do
+      let(:data)   {{some_attribute: 5}}
+      let(:metadata)     {{some_meta: 1}}
+      let(:domain_event) { SomethingHappened.new(data: data, metadata: metadata) }
```
paneq: please provide `event_id` manually.

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      specify '#event_to_serialized_record returns YAML serialized record' do
+        record = subject.event_to_serialized_record(domain_event)
+        expect(record).to            be_a SerializedRecord
+        expect(record.id).to         eq domain_event.event_id
```
paneq: please verify that it equals that manually provided UUID that I asked for above :)

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        expect(record.metadata).to   eq YAML.dump(metadata)
+        expect(record.data).to       eq YAML.dump(data)
+        expect(record.event_type).to eq SomethingHappened
```
paneq: You can see an incompatible interface and I asked for `String`. Here it is a `Class` but in next test it is a `String`. It should be the same thing always.

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      specify '#serialized_record_to_event returns event instance' do
+        record = SerializedRecord.new(
+          id:         domain_event.event_id,
+          data:       YAML.dump(data),
+          metadata:   YAML.dump(metadata),
+          event_type: SomethingHappened.name
```
paneq: `String` here, `Class` in previous test.

ruby_event_store/lib/ruby_event_store/serialized_record.rb
paneq: Do you think we should check that `id` is a string?

ruby_event_store/lib/ruby_event_store/serialized_record.rb
```diff
@@ -0,0 +1,13 @@
...
+      @data       = data
+      @metadata   = metadata
+      @event_type = event_type
+      freeze
+    end
```
paneq: Do you think we should verify the interface that `id, data, metadata, event_type` are strings?

paneq: That could possibly help avoid some errors.

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      specify '#serialized_record_to_event its using events class remapping' do
+        subject = described_class.new(events_class_remapping: {'EventNameBeforeRefactor' => 'SomethingHappened'})
+        record = SerializedRecord.new(
+          id:         domain_event.event_id,
+          data:       YAML.dump(data),
+          metadata:   YAML.dump(metadata),
+          event_type: "EventNameBeforeRefactor"
+        )
+        event = subject.serialized_record_to_event(record)
+        expect(event).to be_a SomethingHappened
```
paneq: `expect(event).to eq(domain_event)` maybe?

ruby_event_store/spec/serialized_record_spec.rb
```diff
@@ -0,0 +1,19 @@
+module RubyEventStore
+  module Mappers
+    RSpec.describe SerializedRecord do
+      let(:id)         { double(:id) }
+      let(:data)       { double(:data) }
+      let(:metadata)   { double(:metadata) }
+      let(:event_type) { double(:event_type) }
+
+      specify '#event_to_serialized_record returns YAML serialized record' do
```
paneq: `'#event_to_serialized_record returns YAML serialized record'` nope, not really what the test is about.

ruby_event_store/spec/serialized_record_spec.rb
```diff
@@ -0,0 +1,19 @@
+module RubyEventStore
+  module Mappers
+    RSpec.describe SerializedRecord do
```
paneq: I don't feel like we are testing anything here. Is this really necessary? If anything like `id, data, metadata, event_type` was not what was passed into it, then other tests would fail. The only thing which needs to be tested is `frozen?`.

It might make more sense if we introduce `Strings` verifications that I proposed.

ruby_event_store/spec/mappers/default_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+  end
+end
```
paneq: It would be nice to test with another serializer to document that it is possible to use it in: `def initialize(serializer: YAML` 

I guess mutant is missing a mutation similar to:

```ruby
def initializer(serializer: Object.new)
  serializer = YAML
end
```

that would show that you never use anything other than the default so why is it a method parameter :)

ruby_event_store/ruby_event_store.gemspec
andrzejsliwa: ups, thx

ruby_event_store/lib/ruby_event_store/serialized_record.rb
andrzejsliwa: event_id makes sense

ruby_event_store/spec/mappers/default_spec.rb
andrzejsliwa: 👍

ruby_event_store/spec/mappers/default_spec.rb
andrzejsliwa: 👍

ruby_event_store/lib/ruby_event_store/mappers/default.rb
andrzejsliwa: Thats why I like contracts.ruby, this can be detected on first run of code (in spec)

ruby_event_store/lib/ruby_event_store/serialized_record.rb
andrzejsliwa: done

PR #170, opened 2017-11-26
Add RubyEventStore#when_any_of method

vbyno:
There are the cases when it is needed to use one handler for many events
within RubyEventStore::Projection.
Passing array of events to #when method we can rewrite this:
```
  Projection.
  from_stream(stream_name).
  init( -> { { total: 0 } }).
  when(MoneyDeposited, ->(state, event) { state[:total] += event.data[:amount] }).
  when(MoneyWithdrawn, ->(state, event) { state[:total] += event.data[:amount] }).
  run(event_store)
```
to this:
```
  Projection.
  from_stream(stream_name).
  init( -> { { total: 0 } }).
  when([MoneyDeposited, MoneyWithdrawn], ->(state, event) { state[:total] += event.data[:amount] }).
  run(event_store)
```

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/projection.rb
+++ b/ruby_event_store/lib/ruby_event_store/projection.rb
@@ -25,8 +25,11 @@ module RubyEventStore
       self
     end
 
-    def when(event, handler)
-      @handlers[event] = handler
+    def when(events, handler)
+      Array(events).each do |event|
+        @handlers[event] = handler
+      end
+
       self
     end
 
--- a/ruby_event_store/spec/projection_spec.rb
+++ b/ruby_event_store/spec/projection_spec.rb
@@ -135,6 +135,18 @@ module RubyEventStore
       expect(deposits).to eq(total: 10)
     end
 
+    specify "subsrcibe one handler to many events" do
+      stream_name = "Customer$123"
+      event_store.publish_event(MoneyDeposited.new(data: { amount: 10 }), stream_name: stream_name)
+      event_store.publish_event(MoneyWithdrawn.new(data: { amount: 2 }), stream_name: stream_name)
+      cashflow = Projection.
+        from_stream(stream_name).
+        init( -> { { total: 0 } }).
+        when([MoneyDeposited, MoneyWithdrawn], ->(state, event) { state[:total] += event.data[:amount] }).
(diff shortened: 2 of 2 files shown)


paneq:
Why not `when([MoneyDeposited, MoneyWithdrawn], ...` ?

vbyno:
@paneq , I may change my pull request if you think this functionality is needed. Are you OK with `when([...]..`?

paneq:
Right now I am asking what was your motivation for introducing new API when the change for me seems to naturally fit an already existing one. Why `when_any_of` over `when` ? What's benefit of this approach? I am not familiar with this code, I am asking rather as someone who tries to provide developers with friendly APIs.

vbyno:
@paneq , I just like methods to be as straitforward as it is possible. Method which can be invoked with an array of events or with a single event looks too diverse to me. From the other point, it is easier to read the code. To me code like this
```
.when(Event1, handler1)
.when_any_of(Event1, Event2, handler2)
.when(Event3, handler3)
```
looks cleaner than
```
.when(Event1, handler1)
.when([Event1, Event2], handler2)
.when(Event3, handler3)
```

vbyno:
We may consider `.when(Event1, Event2, handler2)` too. (without an array, just not strict arity of `when` method

paneq:
for the record: https://eventstore.org/docs/projections/4.0.2/user-defined-projections/

I will let others raise their voices on the API of projections @mpraglowski @pawelpacana

mpraglowski:
We already have similar case with `subscribe` method.

When we subscribing to several events we use:

```ruby
event_store.subscribe(subscriber, [InvoiceCreated, InvoiceUpdated])
```

not something like:

```ruby
event_store.subscribe_to_all_events(subscriber, InvoiceCreated, InvoiceUpdated)
```

That's why I would prefer to have consistent API in projections:

```ruby
.when([Event1, Event2], handler)
```

vbyno:
@mpraglowski , got it

vbyno:
@mpraglowski , @paneq , I've updated the PR. Please, take a look

mpraglowski:
@vbyno could you update documentation ?

vbyno:
@mpraglowski , sure

PR #171, opened 2017-11-27
support for declarative event handlers methods for aggregate root

andrzejsliwa:
I would like to introduce second way of declaring event handler methods on aggregate:
```ruby
  on OrderCreated do |_event|
    @status = :created
  end
```
which will generate standard private handler method:
```ruby
  def apply_order_created(_event)
    @status = :created
  end
```

Full example of usage (include of Events modules, and declaration):
```ruby
class Order
  include AggregateRoot
  include Orders::Events

  def initialize
    @status = :draft
  end

  attr_accessor :status
  private

  on OrderCreated do |_event|
    @status = :created
  end

  def apply_order_expired(_event)
    @status = :expired
  end
end
```

usage of proposed declarative `on ` will be optional and will coexist with standard convention (after all its generating exactly same method like required by standard convention)

I see few pros of declarative approach:
-   I can see direct correlation between callback and event name (with standard approach, conversion is hidden, and method looks like not used. 
- I case of missing Event class I will see error immediately without running specific use case specs
- I can easily navigate to related event by using standard editors/IDE features
- I can easily find usages of Event by using standard editors/IDE features, which will improve the flow in case of refactoring

### DIFF
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -8,9 +8,7 @@ module AggregateRoot
     def on(*event_klasses, apply_strategy:  DefaultApplyStrategy.new, &block)
       event_klasses.each do |event_klass|
         handler_name = apply_strategy.handler_name_by_class(event_klass)
-        define_method handler_name do |event|
-          instance_exec event, &block
-        end
+        define_method handler_name, &block
         private handler_name
       end
     end
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -4,6 +4,22 @@ require 'aggregate_root/configuration'
 require 'aggregate_root/default_apply_strategy'
 
 module AggregateRoot
+  module ClassMethods
+    def on(*event_klasses, apply_strategy:  DefaultApplyStrategy.new, &block)
+      event_klasses.each do |event_klass|
+        handler_name = apply_strategy.handler_name_by_class(event_klass)
+        define_method handler_name do |event|
+          instance_exec event, &block
+        end
+        private handler_name
+      end
+    end
+  end
+
+  def self.included(host_class)
+    host_class.extend(ClassMethods)
+  end
+
   def apply(*events)
     events.each do |event|
       apply_strategy.(self, event)
--- a/aggregate_root/lib/aggregate_root/default_apply_strategy.rb
+++ b/aggregate_root/lib/aggregate_root/default_apply_strategy.rb
@@ -15,11 +15,15 @@ module AggregateRoot
       end
     end
 
-    private
-    attr_reader :strict
+    def handler_name_by_class(event_class)
+      "apply_#{event_class.name.demodulize.underscore}"
+    end
 
     def handler_name(event)
-      "apply_#{event.class.name.demodulize.underscore}"
+      handler_name_by_class(event.class)
     end
+
+    private
+    attr_reader :strict
   end
 end
--- a/aggregate_root/spec/aggregate_root_spec.rb
+++ b/aggregate_root/spec/aggregate_root_spec.rb
@@ -189,4 +189,28 @@ describe AggregateRoot do
     expect(order.unpublished_events.respond_to?(:pop)).to eq(false)
     expect(order.unpublished_events.respond_to?(:unshift)).to eq(false)
   end
+
+  describe ".on" do
+    it "generates private apply handler method" do
+      check = double
+      expect(check).to receive(:call).with(instance_of(Orders::Events::SpanishInquisition))
+
+      Order.on(Orders::Events::SpanishInquisition) do |event|
+        check.call(event)
+        @status = :spanish_inqusition
+      end
+
+      order = Order.new
+      order.apply(Orders::Events::SpanishInquisition.new)
+      expect(order.status).to          eq :spanish_inqusition
+      expect(order.private_methods).to include :apply_spanish_inquisition
+    end
+  end
+
+  describe '.include' do
+    it 'extend class with AggregateRoot::ClassMethods' do
+      expect(Order).to receive(:extend).with(AggregateRoot::ClassMethods)
+      Order.include(AggregateRoot)
+    end
+  end
 end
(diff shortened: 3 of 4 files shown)


riethmayer:
The navigation argument for me really is the killer feature of this pull request. 👍

aggregate_root/lib/aggregate_root.rb
```diff
@@ -4,6 +4,22 @@
 require 'aggregate_root/default_apply_strategy'
 
 module AggregateRoot
+  module ClassMethods
+    def on(*event_klasses, apply_strategy:  DefaultApplyStrategy.new, &block)
+      event_klasses.each do |event_klass|
+        handler_name = apply_strategy.handler_name_by_class(event_klass)
+        define_method handler_name do |event|
+          instance_exec event, &block
```
paneq: Why do we need `instance_exec` ? Why not:

```ruby
define_method(handler_name, &block)
```

?

aggregate_root/lib/aggregate_root.rb
```diff
@@ -4,6 +4,22 @@
...
+        end
+        private handler_name
+      end
+    end
+  end
+
+  def self.included(host_class)
+    host_class.extend(ClassMethods)
```
paneq: could we be explicit and ask the client to use:

```
class Order
  include AggregateRoot
  extend AggregateRoot::On
```

There might be a better name  then `AggregateRoot::On` maybe `AggregateRoot::OnHandlers` or `AggregateRoot::EasyHandlers` ?

andrzejsliwa: why? having it enabled by default making no difference, its up to developer to decide to use it or not

aggregate_root/lib/aggregate_root.rb
```diff
@@ -4,6 +4,22 @@
 require 'aggregate_root/default_apply_strategy'
 
 module AggregateRoot
+  module ClassMethods
+    def on(*event_klasses, apply_strategy:  DefaultApplyStrategy.new, &block)
+      event_klasses.each do |event_klass|
+        handler_name = apply_strategy.handler_name_by_class(event_klass)
```
paneq: this seems _weird_ to me... Could we not `tell` the strategy that we defined a method (with any name, any convention) and we want it to be used as a handler for `*event_klasses` instead of asking the `apply_strategy` about this the method name we should define?

aggregate_root/lib/aggregate_root.rb
```diff
@@ -4,6 +4,22 @@
 require 'aggregate_root/default_apply_strategy'
 
 module AggregateRoot
+  module ClassMethods
+    def on(*event_klasses, apply_strategy:  DefaultApplyStrategy.new, &block)
```
paneq: The fact that it gets `apply_strategy` as an argument is a big smell for me. After all https://github.com/RailsEventStore/rails_event_store/blob/0f7ee713a08d9b834c28cbfe25c00d995e3d8b64/aggregate_root/lib/aggregate_root.rb#L47 the strategy can be overwritten on an instance level, not class level so there is a mismatch here. These two features don't place nicely together.

andrzejsliwa: to be honest, I found this issue as well, for me stragegy can be class configuration, not instance method

aggregate_root/lib/aggregate_root.rb
paneq: > why?

less magic, more explicit. when you look into the class you can easily see where the additional class methods are coming from.

aggregate_root/lib/aggregate_root.rb
andrzejsliwa: but I can't refactor it, without breaking backwad compatibility

paneq:
My general feeling is that while it can easily work with the default strategy being used right now, i am not sure how it can easily work with non-default strategies. I am not sure if that should be a blocker but we should discuss it imho.

aggregate_root/lib/aggregate_root.rb
andrzejsliwa: then you will need as well ;)
```ruby
include Aggregate::Apply
```

andrzejsliwa: after all those are part of main functionality

andrzejsliwa:
For me its improving usage of default strategy enough to not consider need of building custom strategy

paneq:
> The navigation argument for me really is the killer feature of this pull request.

@riethmayer I use this convention for that:

```ruby
  def apply_strategy
    ->(_me, event) {
      {
        OrderItemAdded => method(:apply_item_added),
        OrderSubmitted => method(:apply_submitted),
        OrderCancelled => method(:apply_cancelled),
        OrderExpired   => method(:apply_expired),
        OrderShipped   => method(:apply_shipped),
      }.fetch(event.class).call(event)
    }
  end
```

That way I can in RubyMine easily navigate to event classes such as `OrderItemAdded` and handlers such as `apply_item_added` with one click. Just FYI, not really related to the PR.

The additional benefit is that if you grep/look for the classes and methods whether they are used anywhere in your code (especially when trying to safely remove code) you can easily see them being used.

andrzejsliwa:
@paneq @riethmayer sure, we can achieve "navigation" this way, but in my opinion such example have more boilerplate and adding "technical noise" to aggregate

aggregate_root/lib/aggregate_root.rb
paneq: One month has passed and we haven't figure out what to do about it.

andrzejsliwa:
@paneq @pawelpacana what we should do with this PR, do you have anything again of merging it?

paneq:
I wish I knew. I am _pro_ this feature, but against `apply_strategy:  DefaultApplyStrategy.new`. 

But I don't see right now in which direction we could refactor the whole solution to make it better in overall. The fact that `apply_strategy` is defined on instance level but it is needed in class level makes the design troublesome. I think we should stick to one place only and move it in the same direction. I voiced my concerns but nothing was proposed by anybody to resolve the matter so far.

paneq:
Perhaps `on` could store somehow on class level the list of defined methods (with any name really) and which events they handle, and `default_apply_strategy` could read that list and add to an internal map?

```ruby
   def apply_strategy
     DefaultApplyStrategy.new(class.methods_defined_by_on)
   end
```

That way, non default strategies could also have access to that data and make use of it, if they wish... ? Does that make sense to you? :)

paneq:
#190 went in instead.

Issue #134, opened 2017-10-04
link_events_to_stream API


paneq:
It would copy/link events to new streams but without triggering handlers again.

So for the first time you would publish_event which would save it in `all` stream, provided stream and then later (maybe in a handler) you could use link_events_to_stream to append to another stream.

I believe 1 stream at a time should be allowed as different stream can use different strategies for appending.

vbyno:
@paneq , could you please specify an API you expect to see here?
Right now in our project we use something like
```
publish(UserAssignedToCompany.strict(metadata: { streams_to_append: ["company-#{ company.id }", 'other_stream'], data: {...} }))
```

paneq:
@vbyno I imagine

```ruby
link_events_to_stream(event_id, stream_name: expected_version:)
```

so it is similar to `publish_event`.

* We need `event_id` instead of a full `event` because the `data` and `metadata` is going to be the same.
* We need one `stream_name` because the `expected_version` can apply to one `stream_name` at a time.
* And we need `expected_version` to handle optimistic locking http://railseventstore.org/docs/expected_version/

paneq:
This similarity to `publish_events` would make the implementation in repositories much easier IMHO and similar to what they already do when the event is published the 1st time.

PR #177, opened 2017-12-31
link_to_stream

paneq:
Implements API for linking events to streams #134 

* does not trigger handlers
* verifies uniqueness of linked event in a stream
* verifies existence of linked event

### DIFF
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -165,6 +165,7 @@ module RailsEventStoreActiveRecord
       e.message.include?("event_store_events_pkey") ||  # Postgresql
       e.message.include?("event_store_events.id")   ||  # Sqlite3
 
+      e.message.include?("Key (stream, event_id)")  ||  # Postgresql
       e.message.include?("event_store_events_in_streams.stream, event_store_events_in_streams.event_id")  ## Sqlite3
     end
 
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -12,48 +12,16 @@ module RailsEventStoreActiveRecord
     end
 
     def append_to_stream(events, stream_name, expected_version)
-      raise RubyEventStore::InvalidExpectedVersion if stream_name.eql?(RubyEventStore::GLOBAL_STREAM) && !expected_version.equal?(:any)
-
-      events = normalize_to_array(events)
-      expected_version =
-        case expected_version
-        when Integer, :any
-          expected_version
-        when :none
-          -1
-        when :auto
-          eis = EventInStream.where(stream: stream_name).order("position DESC").first
-          (eis && eis.position) || -1
-        else
-          raise RubyEventStore::InvalidExpectedVersion
-        end
-
-      ActiveRecord::Base.transaction(requires_new: true) do
-        in_stream = events.flat_map.with_index do |event, index|
-          position = unless expected_version.equal?(:any)
-            expected_version + index + POSITION_SHIFT
-          end
-          build_event_record(event).save!
-          events = [{
-            stream: RubyEventStore::GLOBAL_STREAM,
-            position: nil,
-            event_id: event.event_id
-          }]
-          events.unshift({
-            stream:   stream_name,
-            position: position,
-            event_id: event.event_id
-          }) unless stream_name.eql?(RubyEventStore::GLOBAL_STREAM)
-          events
-        end
-        EventInStream.import(in_stream)
+      add_to_stream(events, stream_name, expected_version, true) do |event|
+        build_event_record(event).save!
+        event.event_id
       end
-      self
-    rescue ActiveRecord::RecordNotUnique => e
-      if detect_pkey_index_violated(e)
-        raise RubyEventStore::EventDuplicatedInStream
+    end
+
+    def link_to_stream(event_ids, stream_name, expected_version)
+      add_to_stream(event_ids, stream_name, expected_version, false) do |event_id|
+        event_id
       end
-      raise RubyEventStore::WrongExpectedEventVersion
     end
 
     def delete_stream(stream_name)
@@ -145,6 +113,52 @@ module RailsEventStoreActiveRecord
     private
 
     attr_reader :mapper
+
+    def add_to_stream(collection, stream_name, expected_version, include_global, &to_event_id)
+      raise RubyEventStore::InvalidExpectedVersion if stream_name.eql?(RubyEventStore::GLOBAL_STREAM) && !expected_version.equal?(:any)
+
+      collection = normalize_to_array(collection)
+      expected_version =
+        case expected_version
+        when Integer, :any
+          expected_version
+        when :none
+          -1
+        when :auto
+          eis = EventInStream.where(stream: stream_name).order("position DESC").first
+          (eis && eis.position) || -1
+        else
+          raise RubyEventStore::InvalidExpectedVersion
+        end
+
+      ActiveRecord::Base.transaction(requires_new: true) do
+        in_stream = collection.flat_map.with_index do |element, index|
+          position = unless expected_version.equal?(:any)
+            expected_version + index + POSITION_SHIFT
+          end
+          event_id = to_event_id.call(element)
+          collection = []
+          collection.unshift({
+            stream: RubyEventStore::GLOBAL_STREAM,
+            position: nil,
+            event_id: event_id
+          }) if include_global
+          collection.unshift({
+            stream:   stream_name,
+            position: position,
+            event_id: event_id
+          }) unless stream_name.eql?(RubyEventStore::GLOBAL_STREAM)
+          collection
+        end
+        EventInStream.import(in_stream)
+      end
+      self
+    rescue ActiveRecord::RecordNotUnique => e
+      if detect_pkey_index_violated(e)
+        raise RubyEventStore::EventDuplicatedInStream
+      end
+      raise RubyEventStore::WrongExpectedEventVersion
+    end
 
     def detect_pkey_index_violated(e)
       e.message.include?("for key 'PRIMARY'")       ||  # MySQL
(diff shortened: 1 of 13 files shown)


rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -146,10 +117,61 @@ def get_all_streams
 
     attr_reader :mapper
 
+    def add_to_stream(collection, stream_name, expected_version, include_global, &to_event_id)
```
andrzejsliwa: for me this methods still doing to much, it's difficult to reason about it, please reconsider extract few well named methods from it, to show on this level only top view of the process

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -146,10 +117,61 @@ def get_all_streams
...
+      end
+      raise RubyEventStore::WrongExpectedEventVersion
+    end
+
     def detect_pkey_index_violated(e)
-      e.message.include?("for key 'PRIMARY'")       ||  # MySQL
-      e.message.include?("event_store_events_pkey") ||  # Postgresql
-      e.message.include?("event_store_events.id")       # Sqlite3
+      m = e.message
+      m.include?("for key 'PRIMARY'")       ||  # MySQL
```
andrzejsliwa: please consider here naming matching strings as:
```ruby
MYSQL_PRIMARY_KEY_MESSAGE = "for key 'PRIMARY'"
POSTGRES_PRIMARY_KEY_MESSAGE  = "event_store_events_pkey"
...
m.include?(MYSQL_PRIMARY_KEY) ||
m.include?(POSTGRES_PRIMARY_KEY_MESSAGE) 
```

comments will be not needed if constant name will show meaning and intent.

ruby_event_store/spec/in_memory_repository_spec.rb
```diff
@@ -8,9 +8,32 @@ module RubyEventStore
     let(:test_race_conditions_any)   { false }
     let(:test_race_conditions_auto)  { true }
     let(:test_expected_version_auto) { true }
+    let(:test_link_events_to_stream) { true }
```
andrzejsliwa: I don't see where it used.

ruby_event_store/lib/ruby_event_store/in_memory_repository.rb
```diff
@@ -80,7 +71,24 @@ def normalize_to_array(events)
       return *events
     end
 
-    def append_with_synchronize(events, expected_version, stream, stream_name)
+    def add_to_stream(events, expected_version, stream_name, include_global)
+      raise InvalidExpectedVersion if !expected_version.equal?(:any) && stream_name.eql?(GLOBAL_STREAM)
+      events = normalize_to_array(events)
+      stream = read_stream_events_forward(stream_name)
+      expected_version = case expected_version
```
andrzejsliwa: why this code (with expected version) is repeated in another place?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: I found 3 methods which definitely make sense extracting. There is potential for more but I could not find good scope and names.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: Makes sense

ruby_event_store/lib/ruby_event_store/in_memory_repository.rb
paneq: Where?

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -146,10 +117,61 @@ def get_all_streams
 
     attr_reader :mapper
 
+    def add_to_stream(collection, stream_name, expected_version, include_global, &to_event_id)
+      raise RubyEventStore::InvalidExpectedVersion if stream_name.eql?(RubyEventStore::GLOBAL_STREAM) && !expected_version.equal?(:any)
+
+      collection = normalize_to_array(collection)
+      expected_version =
```
andrzejsliwa: @paneq here ;)

ruby_event_store/spec/in_memory_repository_spec.rb
paneq: > Large diffs are not rendered by default.

Press "load diff" and check hundreds of new lines of tests in `ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb` :)

ruby_event_store/lib/ruby_event_store/in_memory_repository.rb
paneq: One is InMemoryRepository and the other is ActiveRecordRepository.

andrzejsliwa: sure but is exactly same logic, so why to not share it from ruby_event_store? Maybe extract to separate class all versioning code as strategy pattern, and use it in both places

paneq: It differs in 1 line and it is only repeated in 2 places. Once the number of places reaches 3 or 4 I will consider sharing it between various implementations. So far I was not inclined to do it.

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
```diff
@@ -106,6 +222,25 @@
...
+    ], 'another', :auto)
+    repository.append_to_stream([
+      event0 = TestDomainEvent.new(event_id: SecureRandom.uuid),
+      event1 = TestDomainEvent.new(event_id: SecureRandom.uuid),
+    ], 'stream', :auto)
+    repository.link_to_stream([
+      eventA.event_id,
+      eventB.event_id,
+      eventC.event_id,
+    ], 'stream', 1)
```
pawelpacana: what's the expectation here?

pawelpacana: missing expectation here?

ruby_event_store/lib/ruby_event_store/mappers/default.rb
```diff
@@ -18,6 +18,10 @@ def event_to_serialized_record(domain_event)
         )
       end
 
+      def fetch
+
+      end
+
```
pawelpacana: What's this empty method for?

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
paneq: Inspired by `specify ':auto queries for last position in given stream' do` probably that you can mix `:auto` and `Integer` in that order. Although I can't say the name reveals that...

paneq: The expectation is no exception.

paneq: Making a query would probably be a better idea.

pawelpacana: expect{ ... }.not_to raise_error would be a clear indication of that

ruby_event_store/spec/in_memory_repository_spec.rb
```diff
@@ -12,6 +12,28 @@ module RubyEventStore
...
+        TestDomainEvent.new(event_id: eid = "fbce0b3d-40e3-4d1d-90a1-901f1ded5a4a"),
+        'other',
+        -1
+      )
+      repository.append_to_stream(
+        TestDomainEvent.new(event_id: "a1b49edb-7636-416f-874a-88f94b859bef"),
+        'stream',
+        -1
+      )
+      expect(eid).not_to receive(:eql?)
```
pawelpacana: Is this expectation necessary? Looks like a detail of the implementation. What was the intent here?

pawelpacana:
Regarding `append_to_stream`, `link_to_stream` and `add_to_stream`.

Let's hypothetically assume we have following operations available in repository:
- `commit` which adds new event record to `event_store_events`
- `link` which adds existing event record to `event_store_events_in_streams`

Then `append_to_stream` would consist of `commit` + `link(all_stream)` + `link(given_stream)` and some appropriate atomicity guarantee around it.

Following, `link_to_stream` would consist of `link(given_stream)` and some appropriate atomicity guarantee around it.

It is kind-of similar to how `add_to_stream` handles it but without passing boolean flag and maybe making block call unnecessary, 

Wdyt?

ruby_event_store/spec/in_memory_repository_spec.rb
paneq: Let's look at the bigger picture here:

```
    def link_to_stream(event_ids, stream_name, expected_version)
      events = normalize_to_array(event_ids).map{|eid| read_event(eid) }
      add_to_stream(events, expected_version, stream_name, nil)
    end

    def append(events, expected_version, stream, stream_name, include_global)
      raise WrongExpectedEventVersion unless (stream.size - 1).equal?(expected_version)
      events.each do |event|
        raise EventDuplicatedInStream if stream.any?{|ev| ev.event_id.eql?(event.event_id) }
        if include_global
          raise EventDuplicatedInStream if all.any?{|ev| ev.event_id.eql?(event.event_id) }
          all.push(event)
        end
        stream.push(event)
      end
      streams[stream_name] = stream
      self
    end
```

* mutant told me that I could change the `raise EventDuplicatedInStream if stream.any?{|ev| ev.event_id.eql?(event.event_id) }` into `equal?`.
* I was surprised by that :-) and I wondered why. I turned out that the reason is that because when we are linking events we first read them from in memory so `equal?` is good enough (same in memory string is compared). When we append it's different string so `equal?` is not good enough but then  `include_global` is `true` and the next line checks it so everything is ok.
* I had 2 choices.
    * change `eql` to `equal` which I did not like
    * write a test showing why `eql` a not `equal`. I chose this option and this is the test.

It is indeed an implementation detail and that's why it is in a repository test and not in a linter.

The intention of this test is to prove that we check for duplicates in events of a single stream before checking all events. One might call it a performance test (shorter collection is checked before longer collection is checked). The way it is done is by ensuring that we don't compare equality with the other event from the other stream.

Alternative solution:

* disable mutation testing in a single method
* use `equal` instead of `eql` - even though I would prefer not to.
* moving checks into higher methods maybe...
* something else I haven't thought about

That's the story more or less. What's your take on it?

PR #189, opened 2018-02-17
Protobuf mapper

paneq:
### Goal

* Provide reference implementation for another mapper
* Make it possible to not use `RubyEventStore::Event` for defining events

### Checklist

* [x] - Test if it works with ActiveRecord (sqlite3) repository (on Rails event store level)
* [x] - Consider supporting metadata with those protobuf messages that define it
  * [x] - If `metadata_proc` generates structure `{remote_ip: "10.10.10.2", request_id: "111"}` and message only supports `metadata.remote_ip` then it should not try to fill out `request_id`.
    *  However it might be problematic on a few levels. Some classes can expect `metadata[:key]=value` (current RES::Event), some can expect `metadata['key']=value` (protobuf), and some `metadata.key=(value)`
    * Maybe that's not an issue because `metadata_proc` is configurable already? Is it usable enough if other mappers require separate config?
* [ ] - Update documentation
  * [ ] - Mappers - show how to write your own
  * [ ] - Document how this one can be used with protobuf

### Discussion

* Should it be a separate gem? I added `require 'google/protobuf'` in an initializer to avoid RES having a dependency on `google-protobuf`.


paneq:
Related: #154

If we ship it then I can say we have a proof that having custom serializers is possible.

paneq:
What do you think @andrzejsliwa @mlomnicki @mpraglowski @gottfrois @pawelpacana @fidel @joelvh @sevos @mentero @siemakuba @vbyno

paneq:
weird.

browser - res_active_record failed with 

> Could not find gem 'rails (= 5.1.4)' in any of the gem sources listed in your Gemfile.

ruby_event_store/spec/client_spec.rb
```diff
@@ -214,5 +215,17 @@ module RubyEventStore
       expect(subscriber.handled_events).to be_empty
     end
 
+    specify 'can handle protobuf event class instead of RubyEventStore::Event' do
+      client = RubyEventStore::Client.new(repository: InMemoryRepository.new)
+      event = ResTesting::OrderCreated.new(
+        event_id: "f90b8848-e478-47fe-9b4a-9f2a1d53622b",
```
vbyno: Is that ok that `event_id` here is a part of `data`?  It probably shouldn't be serialized, WDYT?

paneq: @vbyno why not?

I think assuming this message may be serialized using protobuf and sent to a different system, it needs to contain the event id imho.

It was the assumption of the whole solution that the event message must have an id.

In protobuf message you don't distinguish between id, data, metadata. Developer can define the structure anyway they want.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Mappers
+    class Protobuf
+      def initialize(event_id_getter: :event_id, events_class_remapping: {})
+        @event_id_getter = event_id_getter
+        @events_class_remapping = events_class_remapping
+      end
+
+      def event_to_serialized_record(domain_event)
+        SerializedRecord.new(
+          event_id:   domain_event.public_send(event_id_getter),
```
siemakuba: If there is an assumption that the Protobuf message would self-contain all the event data, along with the event_id (possibly) and metadata (possibly) is there a real value in trying to extract the ID from the message? The moment you agree to accept an event which structure is beyond your control you should not make any assumptions about this structure. 
What if the message would not really have any kind of event_id? One may try to pass one of the message attributes as an ID just to fulfil the mapper's requirement. And what if the attribute is not unique?
Moreover, the event_id is skipped when deserializing the record back to the protobuf message (it's obvious - the message itself is the event in this case).

Having said that, I'd choose ensuring the id is unique over trying extracting the id from the message data:
```ruby
event_id:  SecureRandom.uuid
```

siemakuba: Hm, the moment I submitted the comment I realised it still must be possible to retrieve the event from the stream by ID. So, maybe some kind of a checksum would serve as a better event id than a random string?

paneq: @siemakuba If the structure if your messages does not fit the assumptions of our mapper and repository standard then you can substitute the mapper and repository to something that you implement on your own.

> is there a real value in trying to extract the ID from the message?

yes, we can have a reference to it outside the message, in a DB or in-memory.

> What if the message would not really have any kind of event_id

Then you cannot persist this message with our default repository anyway, so this mapper is useless to you anyway :)

> I'd choose ensuring the id is unique over trying extracting

https://github.com/RailsEventStore/rails_event_store/blob/f866e07c046f2b67443b89913edc485318a538eb/rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/migration_template.rb#L19

Doing that for our repository requires having access to it.

I was thinking about using random uuid which would not be inside the message. But then whoever got that message would not know the ID because it would not be there so it would not know what message it is processing and would not be capable of doing exactly once processing so I realized _meh, the messages must contain the event_id anyway_. It's harder with _metadata_ but that's a bigger take.

Imho we should in the future switch types of messages for writing and for reading. Read messages probably can contain more data than just what's been sent. It could contain its position in a stream, stream_name, whether that's the original stream it was written first or not etc. But that's not where we are right now. So in the future I see this as some kind of Container+Original message.

> maybe some kind of a checksum would serve as a better event id than a random string

how can you compute a checksum of a message which must contain the value itself?

paneq:
```
Failures:
  1) no warnings should eq []
     Failure/Error: expect(ruby_event_store_warnings).to eq([])
     
       expected: []
            got: ["/home/travis/build/RailsEventStore/rails_event_store/ruby_event_store/lib/ruby_event_store/mappers/...nt_store/ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb:25: warning: private attribute?"]
     
       (compared using ==)
```

https://bugs.ruby-lang.org/issues/10967

Doh...

paneq:
> Maybe that's not an issue because metadata_proc is configurable already? Is it usable enough if other mappers require separate config?

The proc is configurable, but now how the returned value gets applied to metadata.

> However it might be problematic on a few levels. Some classes can expect metadata[:key]=value (current RES::Event), some can expect metadata['key']=value (protobuf), and some metadata.key=(value)

Perhaps we need a configurable metadata setter with symbol,string,accessor values available or a custom proc?

paneq:
Revelation: This a method that could be implemented in a mapper

```
def add_metadata(event, key, value)
```

especially considering that you change your mapper when you change what kind of events you use.

Issue #188, opened 2018-02-16
InMemoryAdapter runs easily into race conditions


mentero:
Hi, we ran into a problem with using `InMemoryAdapter` today. I prepared an example repository that reproduces the problem. 

https://github.com/mentero/ruby_event_store_test

Long story short. We have daily workers running at night and we wanted to publish some events from them. We have similar setup to [this one](https://github.com/mentero/ruby_event_store_test/commit/cec709764dcda4e805a1b7831ed94a63f86ed57a). The problem is there are many sidekiq workers running at the same time and we got flooded by `RubyEventStore::WrongExpectedEventVersion` quickly.

We have investigated the problem and it seems like a classic race condition caused by `expected_version` being calculated outside of the mutex. It is very easy to run into a scenario where two workers calculate the same version number before entering synchronized block. 

I don't know why `InMemoryAdapter` was chosen over `ActiveRecord` one, but nevertheless this behaviour should probably be considered a bug as it makes `InMemoryAdapter` very brittle.

paneq:
@mentero Thanks for your ticket!

a) `InMemoryAdapter` is not documented right now
b) We've never expected it to be used in production, I believe.
c) what RES version are we talking about?

We will definitely need to think about, but I cannot promise solving this soon.

mentero:
@paneq 

We have `0.20` in our app but I reproduced that on the `0.25` in the repository linked above

paneq:
Related: * https://github.com/RailsEventStore/rails_event_store/blob/5baff280fc5c03a5b6a66da576c3e9a1e12334e1/ruby_event_store/lib/ruby_event_store/in_memory_repository.rb#L99
* https://github.com/RailsEventStore/rails_event_store/commit/d1f52b8ea4bba53c45fb4e980daabef112fde26d#diff-ef17629f8f1d7edd404efaa4b47a6fb2R9

paneq:
Related:

```
module RubyEventStore
  RSpec.describe InMemoryRepository do
    # There is no way to use in-memory adapter in a
    # lock-free, unlimited concurrency way
    let(:test_race_conditions_any)   { false }
    let(:test_race_conditions_auto)  { true }
    let(:test_expected_version_auto) { true }
    let(:test_link_events_to_stream) { true }
```

https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/spec/in_memory_repository_spec.rb

paneq:
[...]

Issue #194, opened 2018-02-19
master branch is failing with v1_v2_schema_migration_spec for rails_event_store_active_record


paneq:
```
Failures:
  1) v1_v2_migration should eq 0
     Failure/Error: expect(process.exit_code).to eq(0)
     
       expected: 0
            got: 7
     
       (compared using ==)
     # ./spec/v1_v2_schema_migration_spec.rb:68:in `fill_data_using_older_gem'
     # ./spec/v1_v2_schema_migration_spec.rb:33:in `block (2 levels) in <top (required)>'
Finished in 9.92 seconds (files took 0.63602 seconds to load)
146 examples, 1 failure, 35 pending
Failed examples:
rspec ./spec/v1_v2_schema_migration_spec.rb:26 # v1_v2_migration should eq 0
```

most likely due to new Rails version (5.1.5) being released

Issue #192, opened 2018-02-19
bunded_context not passing on master


paneq:
```
/d/r/bounded_context (protobuf_mapper|✔) $ rm Gemfile.lock 
~/d/r/bounded_context (protobuf_mapper|✔) $ gem uninstall rails

Select gem to uninstall:
 1. rails-4.2.9
 2. rails-5.0.6
 3. rails-5.1.4
 4. rails-5.1.5
 5. All versions
> 3
Successfully uninstalled rails-5.1.4
~/d/r/bounded_context (protobuf_mapper|✔) $ gem uninstall rails

Select gem to uninstall:
 1. rails-4.2.9
 2. rails-5.0.6
 3. rails-5.1.5
 4. All versions
> 3
Successfully uninstalled rails-5.1.5
~/d/r/bounded_context (protobuf_mapper|✔) $ bundle install
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 10.5.0
Using abstract_type 0.0.7
Using concurrent-ruby 1.0.5
Using i18n 0.9.5
Using minitest 5.11.3
Using thread_safe 0.3.6
Using tzinfo 1.2.5
Using activesupport 5.1.5
Using builder 3.2.3
Using erubi 1.7.0
Using mini_portile2 2.3.0
Using nokogiri 1.8.2
Using rails-dom-testing 2.0.3
Using crass 1.0.3
Using loofah 2.2.0
Using rails-html-sanitizer 1.0.3
Using actionview 5.1.5
Using rack 2.0.4
Using rack-test 0.8.2
Using actionpack 5.1.5
Using nio4r 2.2.0
Using websocket-extensions 0.1.3
Using websocket-driver 0.6.5
Using actioncable 5.1.5
Using globalid 0.4.1
Using activejob 5.1.5
Using mini_mime 1.0.0
Using mail 2.7.0
Using actionmailer 5.1.5
Using activemodel 5.1.5
Using arel 8.0.0
Using activerecord 5.1.5
Using ice_nine 0.11.2
Using memoizable 0.4.2
Using adamantium 0.2.0
Using equalizer 0.0.11
Using anima 0.3.0
Using ast 2.4.0
Using bounded_context 0.23.0 from source at `.`
Using bundler 1.16.1
Using concord 0.1.5
Using diff-lcs 1.3
Using method_source 0.9.0
Using procto 0.0.3
[...]

PR #195, opened 2018-02-19
Null repository object

siemakuba:
This is a kind of solution for issue #188. 
It does not fix the race condition issue inside InMemoryRepository, but it adds a NullRepository.
It could be used in cases where RubyEventStore is used as a PubSub broker.


paneq:
@siemakuba Thank you!, I will have a look once we handle #192 and #194 because it breaks our build currently and affects all PRs.

But just by looking at commit messages. We don't remove constants without deprecating them so moving `InMemoryRepository` to `Repositories::InMemory` is a breaking change for those who use it. My 2cents in the morning.

siemakuba:
> We don't remove constants without deprecating them so moving InMemoryRepository to Repositories::InMemory is a breaking change for those who use it. 

Right, this is reasonable. I will adjust my PR accordingly. Thank you.

ruby_event_store/lib/ruby_event_store/repositories/null.rb
```diff
@@ -0,0 +1,17 @@
...
+      def method_missing(method_name, *_)
+        super unless respond_to_missing?(method_name)
+      end
+
+      def respond_to_missing?(method_name, include_private = false)
+        @repository_to_mimic.respond_to?(method_name, include_private)
+      end
+    end
+  end
+end
```
paneq: I had to check in IRB how would that work.

```
irb(main):007:0* module RubyEventStore
irb(main):008:1>   module Repositories
irb(main):009:2>     class Null
irb(main):010:3>       def initialize(repository_to_mimic = InMemory.new)
irb(main):011:4>         @repository_to_mimic = repository_to_mimic
irb(main):012:4>       end
irb(main):013:3> 
irb(main):014:3*       def method_missing(method_name, *_)
irb(main):015:4>         super unless respond_to_missing?(method_name)
irb(main):016:4>       end
irb(main):017:3> 
irb(main):018:3*       def respond_to_missing?(method_name, include_private = false)
irb(main):019:4>         @repository_to_mimic.respond_to?(method_name, include_private)
irb(main):020:4>       end
irb(main):021:3>     end
irb(main):022:2>   end
irb(main):023:1> end
=> :respond_to_missing?
irb(main):024:0> RubyEvetStore::Repositories::Null
NameError: uninitialized constant RubyEvetStore
	from (irb):24
	from /Users/rupert/.rbenv/versions/2.2.8/bin/irb:11:in `<main>'
irb(main):025:0> RubyEventStore::Repositories::Null
=> RubyEventStore::Repositories::Null
irb(main):026:0> RubyEventStore::Repositories::Null.new
=> #<RubyEventStore::Repositories::Null:0x007f90741ebf48 @repository_to_mimic=#<InMemory:0x007f90741ebf20>>
irb(main):027:0> RubyEventStore::Repositories::Null.new.asd
=> nil
irb(main):028:0> RubyEventStore::Repositories::Null.new.def
NoMethodError: undefined method `def' for #<RubyEventStore::Repositories::Null:0x007f9073832970>
	from (irb):15:in `method_missing'
	from (irb):28
	from /Users/rupert/.rbenv/versions/2.2.8/bin/irb:11:in `<main>'
```

paneq: I must say I am not comfortable with this solution. It's cleaver but here is what I would expect instead

* write methods (append_to_stream, link_to_stream, delete_stream) pass and return nil
* read methods (has_event?, read_events_forward, read_all_streams_backward, etc) raise custom exception

although that does not sound quite right as well

paneq: Technically we only need `append_to_stream` to work (do nothing, but exception-free) and everything else could rise an exception (which i prefer much more than `nil`).

paneq:
@siemakuba I commented on some parts but I have deeper thoughts right now.

paneq:
```ruby
module RubyEventStore
  class Client
    def initialize(repository:,
                   event_broker:  PubSub::Broker.new,
                   page_size: PAGE_SIZE,
                   metadata_proc: nil,
                   clock: ->{ Time.now.utc })
      @repository     = repository
      @event_broker   = event_broker
      @page_size      = page_size
      @metadata_proc  = metadata_proc
      @clock          = clock
    end


    def publish_events(events, stream_name: GLOBAL_STREAM, expected_version: :any)
      append_to_stream(events, stream_name: stream_name, expected_version: expected_version)
      events.each do |ev|
        @event_broker.notify_subscribers(ev)
      end
      :ok
    end

    def publish_event(event, stream_name: GLOBAL_STREAM, expected_version: :any)
      publish_events([event], stream_name: stream_name, expected_version: expected_version)
    end

    def append_to_stream(events, stream_name: GLOBAL_STREAM, expected_version: :any)
      events = normalize_to_array(events)
      events.each{|event| enrich_event_metadata(event) }
      @repository.append_to_stream(events, stream_name, expected_version)
      :ok
    end

    def subscribe(subscriber, event_types, &proc)
      @event_broker.add_subscriber(subscriber, event_types).tap do |unsub|
        handle_subscribe(unsub, &proc)
      end
    end

    def subscribe_to_all_events(subscriber, &proc)
      @event_broker.add_global_subscriber(subscriber).tap do |unsub|
        handle_subscribe(unsub, &proc)
      end
    end

    private

      return unless event.respond_to?(:metadata)
      metadata = event.metadata
      metadata[:timestamp] ||= @clock.()
      metadata.merge!(@metadata_proc.call || {}) if @metadata_proc
    end

    def normalize_to_array(events)
      return *events
    end

    def handle_subscribe(unsub, &proc)
      begin
        proc.call
      ensure
        unsub.()
      end if proc
    end
```

Regarding #188

@andrzejkrzywda had an interesting thought that maybe if you are not using the repository this should tell us something deeper. Here is a thought: _you just want to use `PubSub::Broker.new`_ and not `RailsEventStore::Client`.  I guess you don't need `enrich_event_metadata` and probably we could move `handle_subscribe` into the `Broker`...

The API is different (subscribe -> add_subscriber, subscribe_to_all_events -> add_global_subscriber and most importantly publish_events -> notify_subscribers) but maybe we could make it similar.

So when I think about it...

Maybe it is better to create a custom wrapper on top of `PubSub::Broker.new` that would just do:

```
class PublishOnlyClient
    def initialize(event_broker:  PubSub::Broker.new)
      @event_broker   = event_broker
    end

    def publish_events(events, stream_name: nil, expected_version: nil)
      events.each do |ev|
        @event_broker.notify_subscribers(ev)
      end
      :ok
    end

    def publish_event(event, stream_name: nil, expected_version: nil)
      publish_events([event])
    end

    def subscribe(subscriber, event_types, &proc)
      @event_broker.add_subscriber(subscriber, event_types).tap do |unsub|
        handle_subscribe(unsub, &proc)
      end
    end

    def subscribe_to_all_events(subscriber, &proc)
      @event_broker.add_global_subscriber(subscriber).tap do |unsub|
        handle_subscribe(unsub, &proc)
      end
    end

    private

    def handle_subscribe(unsub, &proc)
      begin
        proc.call
      ensure
        unsub.()
      end if proc
    end
end
```

or make PubSub::Broker respond to (publish_event, subscribe, subscribe_to_all_events).

Either solution would make it possible to substitute it instead of RES::Client and use with the same API.

What do you think @andrzejsliwa @mlomnicki @mpraglowski @pawelpacana @fidel @sevos @siemakuba @vbyno ?

siemakuba:
Sorry for being quiet that long ;)
Using the Broker directly was my first thought, but as you mentioned - the subscription handling logic would still enforce having the Client as well. The wrapper idea is interesting but in my opinion, it introduces quite a lot of overhead as compared to having a client configured with a simple NullRepository object as a repository. 
We implemented it exactly this way, and it solved our issues perfectly. 

I've seen that the issue we reported (#188) has been resolved and is ready to release - we will look into that, but probably would stick with the null object repo for a while.

Saying all that, there's nothing but closing the PR that is left ;) 
Thank you for your comments and support!

PR #208, opened 2018-03-07
use the current version of rails for the migration generation

midnight-wonderer:
Hi guys,  
Don't you think this is kinda weird?
```ruby
def migration_version
  return nil if Gem::Version.new(rails_version) < Gem::Version.new("5.0.0")
  "[4.2]"
end
```
(If rails 5 then select rails 4 migration class)  
It is also leave some database options on schema dump.  
```ruby
  create_table "event_store_events", id: :string, limit: 36, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
    t.string "event_type", null: false
    t.text "metadata"
    t.text "data", null: false
    t.datetime "created_at", null: false
    t.index ["created_at"], name: "index_event_store_events_on_created_at"
  end

  create_table "event_store_events_in_streams", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
    t.string "stream", null: false
    t.integer "position"
    t.string "event_id", null: false
    t.datetime "created_at", null: false
    t.index ["created_at"], name: "index_event_store_events_in_streams_on_created_at"
    t.index ["stream", "event_id"], name: "index_event_store_events_in_streams_on_stream_and_event_id", unique: true
    t.index ["stream", "position"], name: "index_event_store_events_in_streams_on_stream_and_position", unique: true
  end
```
  
Can we match the migration version?  
P.S. I don't mind nitpicking comment if you guys don't satisfied with this PR in anyway I will change it for you.

### DIFF
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/generators/migration_generator.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/generators/migration_generator.rb
@@ -10,13 +10,13 @@ module RailsEventStoreActiveRecord
 
     private
 
-    def rails_version
-      Rails::VERSION::STRING
+    def active_record_version
+      Gem::Version.new(ActiveRecord::VERSION::STRING)
     end
 
     def migration_version
-      return nil if Gem::Version.new(rails_version) < Gem::Version.new("5.0.0")
-      "[4.2]"
+      return nil if active_record_version < Gem::Version.new("5.0.0")
+      "[#{active_record_version.to_s.sub(/\A(\d+\.\d+)\.\d+\z/, '\1')}]"
     end
 
     def timestamp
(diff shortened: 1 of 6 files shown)


paneq:
> Don't you think this is kinda weird?

Nope

> Can we use matched version for the migration?

The migration was written based on what Rails 4.2 does when seeing this code. If Rails 5.X changes defaults for those methods then they won't be doing the same thing anymore.

Does that make sense?

MidnightWonderer:
That's why we have test no?

MidnightWonderer:
@paneq I see what you mean.  
  
Consider the pull request as a feature pull request then.  
  
As of Rails 5.1 the default primary keys are `bigint`.  
Since the event keep on generating new records.  
Exceeding the number is definitely possible.  
As for the `position` field the sequence seems to be scoped down to `stream` so I guess keeping an `int` as the default type still seems to be a safe choice.

pawelpacana:
Thanks for the input. We might reconsider the strategy around versions at some point in the future.

Issue #126, opened 2017-09-28
Enable ActiveJob handlers by default in rails_event_store


paneq:
Right now we have this code ready but it is not enabled as default dispatcher:

```
module RailsEventStore
  class Client < RubyEventStore::Client
    def initialize(repository: RailsEventStore.event_repository,
                   event_broker: EventBroker.new,
                   page_size: PAGE_SIZE)
      capture_metadata = ->{ Thread.current[:rails_event_store] }
      super(repository: repository,
            event_broker: event_broker,
            page_size: page_size,
            metadata_proc: capture_metadata)
    end
  end
end
```

We could detect if active_job is enabled and in such cass pass `ActiveJobDispatcher` to `EventBroker`.

`ActiveJobDispatcher` or standard `Dispatcher` could be wrapped in instrumentation proxy #90 .

Related to #124 as well

paneq:
Also, I believe it could be better to have `dispatcher` as separate dependency, instead of internal dependency of the broker.

paneq:
After done, fix the documentation in http://railseventstore.org/docs/subscribe/ to no longer require custom config

paneq:
https://github.com/rubygems/rubygems/issues/346
https://github.com/rubygems/rubygems/issues/1811

PR #211, opened 2018-03-12
Publish event can link to multiple streams

gmmoreira:
I have a use case where it could be useful to publish an event and link it against multiple streams, before calling the event_broker to notify subscribers. This pull request should not break the current implementation and usage of the client.

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/client.rb
+++ b/ruby_event_store/lib/ruby_event_store/client.rb
@@ -27,7 +27,9 @@ module RubyEventStore
     def append_to_stream(events, stream_name: GLOBAL_STREAM, expected_version: :any)
       events = normalize_to_array(events)
       events.each{|event| enrich_event_metadata(event) }
-      @repository.append_to_stream(events, stream_name, expected_version)
+      first_stream, *other_streams = normalize_to_array(stream_name)
+      @repository.append_to_stream(events, first_stream, expected_version)
+      other_streams.each { |stream| link_to_stream(map_to_event_ids(events), stream_name: stream) }
       :ok
     end
 
@@ -187,6 +189,10 @@ module RubyEventStore
       metadata.merge!(@metadata_proc.call || {}) if @metadata_proc
 
       # event.class.new(event_id: event.event_id, metadata: metadata, data: event.data)
+    end
+
+    def map_to_event_ids(events)
+      normalize_to_array(events).map { |e| e.event_id }
     end
 
     class Page
--- a/ruby_event_store/spec/client_spec.rb
+++ b/ruby_event_store/spec/client_spec.rb
@@ -43,6 +43,16 @@ module RubyEventStore
       expect(client.read_stream_events_forward(GLOBAL_STREAM)).to eq([test_event])
     end
 
+    specify 'publish to multiple streams' do
+      client = RubyEventStore::Client.new(repository: InMemoryRepository.new)
+      test_event = TestEvent.new
+      stream_names = [SecureRandom.uuid, SecureRandom.uuid]
+      expect(client.publish_event(test_event, stream_name: stream_names)).to eq(:ok)
+      stream_names.each do |stream_name|
+        expect(client.read_stream_events_forward(stream_name)).to eq([test_event])
+      end
+    end
+
     specify 'publish first event, expect any stream state' do
       stream = SecureRandom.uuid
       client = RubyEventStore::Client.new(repository: InMemoryRepository.new)


paneq:
@gmmoreira 

Thank you for the PR. I appreciate.

When implementing `link_to_stream` ( https://github.com/RailsEventStore/rails_event_store/issues/134#issuecomment-347106811 ) I decided to take only one `stream_name` because the `expected_version` applies only to one of them, especially when passing an explicit number ( http://railseventstore.org/docs/expected_version/ )

I see this implementation calls `link_to_stream` without passing `expected_version` which forces using the default which might not be what the developer expects (it's hard to say what's expected).

Another issue that I see is that it breaks atomicity/transactionality of the operation. When  `append_to_stream` succeeds but linking fails the operation is half-done.

Temporary workaround for @gmmoreira :

```
class MyEventStore < RailsEventStore::Client
    def append_to_stream(events, stream_name: GLOBAL_STREAM, expected_version: :any)
      first_stream, *other_streams = normalize_to_array(stream_name)
      super(events, stream_name: first_stream, expected_version)
      other_streams.each { |stream| link_to_stream(map_to_event_ids(events), stream_name: stream) }
      :ok
    end
```

Could you tell us more about your use-case where it is nice to have this event linked before calling subscribers? I would love to hear about it.

gmmoreira:
Hi @paneq.
Nice observations about the atomicity, I didn't think about it. 
About the workaroun, we are on our way to create our own gem wrapping the rails-event-store, so maybe we will ended up creating our own Client, as you suggested.
Regarding our use-case, we are using the rails-event-store as a sort of audit-log and to run events sync or async.
We have our own Event class that inherits from the `RailsEventStore::Event`. That class has some interface methods like `process`, `process_async`, `async?`, `stream_name` and `next_events`. It also has a `publish_event` that call the `publish_event` from the RES client, using the `stream_name` of the event.
We also have a single `EventProcessor` class that subscribe to all events. When the event broker notifies it of the published event, it determines if the event is to be executed sync or async. `process_async` will enqueue a job passing the `event_id` to be executed later. When `process` is executed successfully, event processor call `next_events`, which can return an Array of new events, which are published right away. So we have a chain of events that can be fired until we reach events that do not have `next_events`.
Our problem happens when we want to link the same Event to multiple streams, for instance, we want an `Invoice::CreateEvent` to be in the stream `Invoice$1` and `Customer$1`. We made our `stream_name` method to also return Array of Strings, so the `publish_event` publish using the first element in the Array and use `link_to_stream` to link using the other stream names, similar to the pull request I made. Due to the nature of events calling next events, the `link_to_stream` methods only get executed after all events have already been executed, but they will get linked in reverse order of what we want.

Let's say we have `Invoice::CreateEvent` and `Invoice::CreatedEvent`. Both returns in `stream_name` ['Invoice$1', 'Customer$1']. The stream for `Invoice$1` would be `[Invoice::CreateEvent, Invoice::CreatedEvent]`, but `Customer$1` would be `[Invoice::CreatedEvent, Invoice::CreateEvent]`. The way to resolve our issue is to do what I made, publish and link events to all stream_names before calling the event_broker, so our EventProcessor do not get executed and more events are processed.

I try reading all the code of the gem but didn't any elegant way of resolving this, besides modifying the client class. Do you guys have any suggestions? I don't mind my PR getting closed if you want.

Sorry for the long post.

paneq:
I read this explanation 6 times in 50% of time I think I understand it and 50% of time I think I do not.

> Due to the nature of events calling next events, the link_to_stream methods only get executed after all events have already been executed, but they will get linked in reverse order of what we want.

What is preventing you from first linking to some streams and then doing publishing the _next events_ ?

paneq:
Also some comments about application architecture that you might want to disregard if you are absolutely sure you know what you are doing. I am just a random person in the Internet which does not know your app.

> we are on our way to create our own gem wrapping the rails-event-store, so maybe we will ended up creating our own Client, as you suggested.

wrapping other libraries in an adapter which interface you own is always a good idea. 👍 

> Regarding our use-case, we are using the rails-event-store as a sort of audit-log and to run events sync or async.

Sounds good.

> We have our own Event class that inherits from the RailsEventStore::Event. 

Makes sense.

> That class has some interface methods like process, process_async, async?, stream_name and next_events. It also has a publish_event that call the publish_event from the RES client, using the stream_name of the event.

That I did not expect. Sounds like this class now has the responsibilites of being a data structure, aggregate, process manager and application service.

> We also have a single EventProcessor class that subscribe to all events. When the event broker notifies it of the published event, it determines if the event is to be executed sync or async. 

Here is how we handle recognizing that some handlers (aka subsribers) are async in our ActiveJob integration:
* https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
* http://railseventstore.org/docs/subscribe/
* and btw: https://github.com/RailsEventStore/rails_event_store/issues/126

You could provide your own dispatcher that can schedule async handlers in Resque like we do with ActiveJob

> When process is executed successfully, event processor call next_events, which can return an Array of new events, which are published right away. 

Sounds like a job for a handler or a process manager. Are you sure you want to trigger events and not commands?

> So we have a chain of events that can be fired until we reach events that do not have next_events.

Sounds like a process as well.

> abstract_method :stream_name

We used to have a Event=>stream name configuration years ago but we learned from that experience and we decided not to do it that way.

> abstract_method :async?

You might want some handlers to be sync in reaction to an event and some to be async. Are you sure how it should be processed is a property of an event itself?

Also `Invoice::CreatedEvent` sounds like an event (although is Invoice really created? perhaps there is a better word) but `Invoice::CreateEvent` sounds more like a command, not an event.

It's interesting to see how you used RES in a way we completely have not anticipated.

gmmoreira:
Thanks @paneq. I will take your comments and learn from them, that's sure. Maybe we are misusing the way RES was supposed to work, but so far it has been working flawlessly for our needs. Our current architecture is based on another app of ours that had a custom implementation of an event, async processing. We found RES and saw in it a way to implement something similar in a new service that was borning. 

> What is preventing you from first linking to some streams and then doing publishing the next events ?
AFAIK, an event is only persisted in database after `append_to_stream` is executed, which is called by `publish_event` method and I can only use the `link_to_stream` after that. Our `EventProcessor` class is a subscriber to all events, so as soon as I publish an event, it automatically get processed by the event processor, which call `next_events` and publish them too. So the first event only get linked to other streams after all stack events are executed.

Regarding the provided examples, `Invoice::CreateEvent` would be created in a controller's action. The data of this event would be whatever params were used to create an invoice. The `process` method would actually execute whatever is necessary to create the invoice. I do believe here the event is acting as a command.

The `process` method, called by the `EventProcessor`, if executed successfully would make the EventProcessor call `next_events` of the processed event. `CreateEvent` would them return an Array containing an instance of `Invoice::CreatedEvent`, indicating the invoice was created successfully. `Invoice::CreatedEvent` could also return an array of `next_events`, like sending an email to the customer or creating a metric entry in InfluxDB.

That's why I say we are using RES as an audit-log thing. We have a page similar to RES-browser gem, listing all the events of the stream. If we take a look at the stream `Invoice$1`, we would see both events. One says there was an attempt to create an invoice and whatever params were used, while the other represents that the invoice was created.

I believe I still have something to learn about CQRS and ES.

Anyway, our wrapping around RES using your workaround will fix our issue regarding multiple streams. If you want please close this PR, trying to bend RES to resolve our issue is not the way to go.

paneq:
I now understood the issue. A a response to message A (let's call it a message, not event because you are using those messages as commands and events) you want to publish message B,C. If you had a handler subscribed to all events that would first linked them then you would link message B, publish messages D, E (ie in reaction to B), link messages D, E, and then link message C and that would wrong order because you want C to be after B, not after sub-messages D,E triggered by B.

Perhaps _Scheduling async handlers after commit_ could help you: http://railseventstore.org/docs/subscribe/ , https://www.dropbox.com/s/6pi41fyaveunc4b/Screenshot%202018-03-14%2007.58.50.png?dl=0 .

In a transaction first publish, then link messages, when the transaction is committed async handler could publish sub-messages.

Issue #153, opened 2017-10-26
Suggested configuration defaults


pawelpacana:
I have a feeling there are several configuration snippets floating around which don't necessarily blend well with each other. 

I consider RailsEventStore as an ecosystem of several gems which includes (and is not limited to) aggregate_root and command_bus. There's a Rails middleware addition for RailsEventStore itself and soon there will be web stream browser.

How about changing suggested defaults to following structure;

```ruby
Rails.application.configure do
  config.x.rails_event_store.client           = RailsEventStore::Client.new
  config.x.rails_event_store.request_metadata = ->(env) { { kaka: 'dudu' } }
  config.x.rails_event_store.command_bus      = Arkency::CommandBus.new
end
```

Other related gems could then rely on such defaults, i.e. `AggregateRoot.configuration.default_event_store`.

Thoughts? 

Reference: http://guides.rubyonrails.org/v5.1/configuring.html#custom-configuration

fidel:
👍🏻

paneq:
Definitely something worth improvement.

Also I don't like how differently they are configure:

```
# global
AggregateRoot.configure do |config|
  config.default_event_store = RailsEventStore::Client.new
end

# instance with arguments
event_store = RailsEventStore::Client.new(
  event_broker: RailsEventStore::EventBroker.new(
    dispatcher: RailsEventStore::ActiveJobDispatcher.new(
      proxy_strategy: RailsEventStore::AsyncProxyStrategy::AfterCommit.new
    )
  )
)

# rails config
config.to_prepare do
  config.command_bus = Arkency::CommandBus.new
  register = command_bus.method(:register)

  { FooCommand => FooService.new(event_store: event_store).method(:foo),
    BarCommand => BarService.new,
  }.map(&:register)
end
```

etc etc

paneq:
Also in RES code itself:

```
# Use active record by default
RailsEventStore.event_repository = RailsEventStoreActiveRecord::EventRepository.new
```

paneq:
I just found out about `RailsEventStore.event_repository` again and I am like 🙀

paneq:
http://railseventstore.org/docs/repository/
but request metadata config uses Rails configuration http://railseventstore.org/docs/request_metadata/

andrzejsliwa:
@pawelpacana i'm using such configuration

in config/application.rb
```ruby
    config.to_prepare do
      Rails.configuration.core = CoreConfiguration.new
    end
```

[...]

PR #230, opened 2018-03-17
Remove global for event repository config

paneq:
Issue: #153


paneq:
I already updated the docs https://github.com/RailsEventStore/railseventstore.org/commit/b0c92faff27740f57a52ab35cdbfd43393eecd0a because the way described now is compatible with current RES code and with what this PR tries to achieve.

paneq:
Related #231

Issue #231, opened 2018-03-17
ruby_event_store mutation testing have been failing for everyone recently


paneq:
https://www.dropbox.com/s/5x63t2yxzvc7xj7/Screenshot%202018-03-17%2013.44.33.png?dl=0

paneq:
Example https://travis-ci.org/RailsEventStore/rails_event_store/jobs/354694165 , passed locally for me

paneq:
@pawelpacana Can you reproduce locally on `no_global_repo` branch?

paneq:
Example 2: https://travis-ci.org/RailsEventStore/rails_event_store/builds/354834684?utm_source=github_status&utm_medium=notification

paneq:
not only i cannot reproduce locally but also

```
# RSpec.configure do |config|
#   config.around(:each) do |example|
#     Timeout.timeout(5, &example)
#   end
# end
```

when I comment this out the mutation testing still passes locally.

paneq:
https://github.com/mbj/mutant#the-crash--stuck-problem-mri says (about older ruby)

> MRI crashes with a segfault and gets stuck when handling the segfault. Depending on the number of active kill jobs mutant might appear to continue normally until all workers are stuck into this state when it begins to hang. Currently mutant must assume that your test suite simply not terminated yet as from the outside (parent process) the difference between a long running test and a stuck MRI is not observable. Its planned to implement a timeout enforced from the parent process, but ideally MRI simply gets fixed.

Perhaps this is what is happening with one mutation. Maybe because we have protobuf which is implemented as a C extension.

paneq:
```
[32m(1241/4170) 100% - killtime: 145.01s runtime: 2757.70s overhead: 2612.69s[0m
1214
evil:RubyEventStore::Client::Within#call:/home/travis/build/RailsEventStore/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:154:f5e03
1242
evil:RubyEventStore::Client::Within#call:/home/travis/build/RailsEventStore/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:154:cf2d2
[32m(1241/4170) 100% - killtime: 145.01s runtime: 2758.70s overhead: 2613.69s[0m
```

in https://travis-ci.org/RailsEventStore/rails_event_store/jobs/355120836 :(

paneq:
```
[...]

PR #232, opened 2018-03-17
Remove ActiveSupport dependency from `ruby_event_store` gem

joelvh:
Removing ActiveSupport dependency to avoid monkey patching and all the other fun stuff when not using Rails/ActiveRecord with RES


paneq:
@joelvh Thank you for your contribution 💚 

I was pretty sure this would not work with namespaced constants but I was wrong. Yay!

I would merge it as soon as possible but your build is affected by a problem that appeared around 2 days ago: #231 .

Issue #184, opened 2018-01-27
Improve read-side of Client API


pawelpacana:
By read-side I mean following client methods:

```
def read_events_forward(stream_name, start: :head, count: page_size)
def read_events_backward(stream_name, start: :head, count: page_size)
def read_stream_events_forward(stream_name)
def read_stream_events_backward(stream_name)
def read_all_streams_forward(start: :head, count: page_size)
def read_all_streams_backward(start: :head, count: page_size)
```

It always took me a while to figure out which method I really wanted to use.
In fact `read_events_forward` and `read_events_backward` cover all use cases. The others seem redundant, especially having explicit "all" stream covering all events.

Instead imagine following fluent API:

```
read.stream("asd").forward.limit(10).from(...)
```

Additionally if we could iterate over large streams in batches, that would beneficial not only for Projections.

WDYT?

andrzejsliwa:
@pawelpacana I like example with fluent api

paneq:
I definitely agree with @pawelpacana . Here is something I've been thinking about.

### How can fluent read api be handled by multiple repositories?

I believe the repository should implement a single method (maybe `read` or `get`) instead of the variety of methods we have right now. The fluent api would then be used to build a single object that would be passed to a repository `read` method.

```
stream("asd").forward.limit(10).from("012345")
```

would create a `Specification` structure with attributes/ivars such as:

* stream
* direction
* limit
* starting_point

that object would be passed to `read(specification)` method which could use it to build the proper query.

Values not provided with fluent API would have defaults (such as ie reading forward from :head from all streams).

Some methods could have set more than 1 attribute (ie reading backwards without starting_point/from provided would change the starting point to `:tail` instead of `:head`).

### What API should be available on top level Client?

In other words the question is, which of these methods could be the starting points to build from `RailsEventStore::Client` ? I see 3 options.

* none of them. We introduce a new starting method which does nothing (ie `read`) and only after that you can start calling the fluent interface methods such as `limit()`, `forward()`, `backward()` etc

        event_store.read.stream("aggregate").forward()

    Conceptually that would mean `read` would return the Specification object on which next fluent methods would be called to build it.

* all of them. You can start from any of this method. Conceptually that would mean instantiating the Specification object and passing the called method with arguments to it.

        event_store.stream("aggregate").forward()
        event_store.limit(20).backward()

* some of them. We could only support some of them (maybe 2) as starting points. Most likely `stream` and maybe `all_streams` or something similar.


[...]

PR #249, opened 2018-03-29
Move common methods to `Repository` module

joelvh:
Hi @paneq, would you accept some updates to share logic across `EventRepository` implementations?

I'm working on the ROM-based repository adapter ([feature/rom_adapter](https://github.com/joelvh/rails_event_store/tree/feature/rom_adapter)) and was looking to find ways to share code a bit more. I think that there can be more sharing between AR and ROM as well.

Let me know your thoughts on this.

Thanks.

~Joel

### DIFF
--- a/rails_event_store/lib/rails_event_store/all.rb
+++ b/rails_event_store/lib/rails_event_store/all.rb
@@ -7,6 +7,7 @@ require 'rails_event_store/deprecations'
 
 module RailsEventStore
   Event                     = RubyEventStore::Event
+  Repository                = RubyEventStore::Repository
   InMemoryRepository        = RubyEventStore::InMemoryRepository
   EventBroker               = RubyEventStore::PubSub::Broker
   Projection                = RubyEventStore::Projection
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -1,18 +1,18 @@
 require 'activerecord-import'
+require 'ruby_event_store/repository'
 
 module RailsEventStoreActiveRecord
   class EventRepository
-
-    POSITION_SHIFT = 1
+    include RubyEventStore::Repository
 
     def initialize(mapper: RubyEventStore::Mappers::Default.new)
-      verify_correct_schema_present
-      @mapper      = mapper
-      @repo_reader = EventRepositoryReader.new(mapper)
+      @mapper        = mapper
+      @repo_reader   = EventRepositoryReader.new(mapper)
     end
 
-    def append_to_stream(events, stream_name, expected_version)
-      add_to_stream(normalize_to_array(events), stream_name, expected_version, true) do |event|
+    def append_to_stream(event_ids, stream_name, expected_version)
+      event_ids = normalize_to_array(event_ids)
+      add_to_stream(event_ids, stream_name, expected_version, true) do |event|
         build_event_record(event).save!
         event.event_id
       end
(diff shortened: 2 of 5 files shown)


paneq:
@joelvh 

> would you accept some updates to share logic across EventRepository implementations?

I think we would however as always there are some gotchas / concerns to think about :)

a) is `InMemoryRepository` using this method? Nope? Perhpas it does not need to be shared? 

`InMemoryRepository`, `RailsEventStoreActiveRecord::EventRepository` and your repository could use a certain method ? That's a good sign to extract.

b) you wanted to for example extract:

```
+    def append_to_stream(events, stream_name, expected_version)
+      add_to_stream(events, stream_name, expected_version, true)
+    end
+
+    def link_to_stream(events, stream_name, expected_version)
+      add_to_stream(events, stream_name, expected_version, nil)
+    end
+  
+  private
+
+    def add_to_stream(events, stream_name, expected_version, include_global, &block)
```

sure, why not. But the truth is... We don't like the fact that `add_to_stream` takes `true` or `nil` as arugment and so many parameters. So I would not want to encourage other repository authors to go the same way. Maybe you/they figure out something more pretty and we could port back a nicer solution to our implementation.

c) `normalize_to_array` is extracted as a method only to avoid issues with a bug in mutant: https://github.com/mbj/mutant/issues/724 and it is skipped from mutations.

d) `normalize_expected_version` looks like a good candidate for extraction.

e) Some tests and mutation tests are failing so we cannot accept it.

f) Here is my guideline for you. 

* Start by copy-pasting `RailsEventStoreActiveRecord::EventRepository` into a new file/class. I suggest inlining code delegated to `EventRepositoryReader`
* Method by method change ActiveRecord calls into ROM calls. 
* Keep the tests green. You already know how to use the tests we already have.
* Commit often. That way you will see progress.
* When you eliminate all ActiveRecord usage, drop activerecord dependency from your library.
* Refactor whenever you want when you are in green.
* At the end of the process see if there is any shared code between your implementation and our implementation. If there is, create a PR suggesting to extract it, to make it shareable. If there is not, then problem solved, there is nothing you needed from us in the first place :)

What do you think about such approach?

paneq:
> It was interesting to see how the "reader" is used in the AR repository because it's more similar to how ROM thinks about a repository.

Just be aware that with fluent-api for reading #184 that we want to introduce soon we might change the repository implementation for reading events to:

```
def read(specification)
end
```

as all those reading methods are very similar, almost identical and with confusingly similar names :)

> In the meantime, do you suggest I fix the tests for this PR or close it?

let's close it, and we will revisit what can be easily shared once you have a working repository. Waiting eagerly to see it.

joelvh:
@paneq Good to know the Reader is on your radar. I definitely have some thoughts on this. I will start by getting the ROM implementation working. You can see that I've implemented 2 ROM repositories, and am just working on getting all tests to pass: https://github.com/joelvh/rails_event_store/blob/feature/rom_adapter/ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/event_repository.rb

joelvh:
@paneq quick question - how do you run tests and mutations locally?

paneq:
@joelvh 

* https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store_active_record/Makefile#L15
* https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store_active_record/Makefile#L19
* https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store_active_record/Makefile#L26
* what CI does (includes connection strings) : https://github.com/RailsEventStore/rails_event_store/blob/master/.travis.yml

joelvh:
OK, I found the commands that were failing - so it's running each by hand. Wasn't sure if you used CI locally.

Thanks!

Issue #193, opened 2018-02-19
Dummy apps and their Gemfiles


paneq:
In `bounded_context` dummy apps have Gemfile such as:

```
source 'https://rubygems.org'

gem 'rails', '~> 5.1.4'
gem 'bounded_context', path: '../../'
```

I was wondering why not:

```
source 'https://rubygems.org'

gem 'rails', '5.1.4'
gem 'bounded_context', path: '../../'
```

Why are we using `~` ?

paneq:
Related #192

paneq:
What do you think @pawelpacana ?

pawelpacana:
Makes sense!

mpraglowski:
Fixed by #253

PR #251, opened 2018-03-30
Use exact Rails version in test dummy apps

mpraglowski:
Fix for #193


pawelpacana:
On the side you could follow: f866e07c046f2b67443b89913edc485318a538eb and rename `dummy_5_1_4` to `dummy_5_1` so that we test latest released bugfix version of given minor.

mpraglowski:
yeah,, maybe that better approach, and we should add 5.2 version

paneq:
> we should add 5.2 version

not yet released but we could

mpraglowski:
The approach with testing latest version within Rails minor version implemented here in #253. Closing this one.

Issue #228, opened 2018-03-16
Protobuf cannot be used with async handlers


paneq:
because: https://github.com/google/protobuf/issues/4391 we always use `YAML` and `google-protobuf` does not work with it right now.

* [x] Figure out
* [x] Mutation testing
* [x] Can we use Protobuf's ANY data-type to serialize metadata?
* [x] Update protobuf tutorial in a documentation
* [x] Update mapper in a documentation (no longer `add_metadata`)

paneq:
hmmm, but we already have `repository.event_to_serialized_record` which we can serialize it and `repository.serialized_record_to_event` which can deserialize it. The only problem is that we would now require something to be `included/extended` in Async handler classes.

Hmmm, actually right now we require `YAML.load`, in first line of background job class. From documentation:

```
class SendOrderEmail < ActiveJob::Base
  def perform(event)
    event = YAML.load(event)
    email = event.data.fetch(:customer_email)
    OrderMailer.notify_customer(email).deliver_now!
  end
end
```

Perhaps we should require:

```ruby
class SendOrderEmail < ActiveJob::Base
  def perform(event)
    event = event_store.deserialize(event)
    # ...
```

That would probably make much more sense as we would be able to change it in any way in the future.

Or maybe:

```ruby
class SendOrderEmail < ActiveJob::Base
  def perform(event_id, data, metadata, event_type)
    event = event_store.deserialize(data, metadata, event_type)
    # ...
```

nah... that does look really good.

paneq:
yeah, `event` should be `{event_id:, data:, metadata:, event_type:}` from `SerializedRecord` where every value is a `String` and ActiveJob will know how to handle it thanks to that.

paneq:
Workaround: https://github.com/google/protobuf/issues/4391#issuecomment-373906339

paneq:
It seems that I need to:

* move `mapper` dependency higher and make it mandatory (because no matter what repo you use we need to know how to serialize/deserialize your events for async handlers.
* pass `mapper` as dependency to `ActiveJobDispatcher#call` ? How `call(subscriber, event, mapper)` ? Requires extending of `RubyEventStore::PubSub::Dispatcher`
* Calling every Event Repository method would require passing a mapper as it no longer would be a repo dependency.

So that's quite a lot of breaking changes.

paneq:
[...]

Issue #72, opened 2017-07-26
Add API for easily getting all historical events of certain kind (and/or other type of metadata)


paneq:
without the need to use underlying ActiveRecord directly

```ruby
event_ar = RailsEventStoreActiveRecord::Event.where(event_type: "MenuUpdated").find_each.select { |mu| mu.data["id"] == 72 }.last
event = RailsEventStoreActiveRecord::EventRepository.new.send(:build_event_entity, event_ar)
Catalog::MenuCatalogHandler.new.(event)
```

mlomnicki:
Idea: do it the GES way. They have a special stream `$et-{EVENT_TYPE}`. Thanks to that it's not necessary to introduce another API.

```ruby
event = res.read_stream_forward("$et-MenuUpdated").select { |e| e.data["id"] == 72 }.last
Catalog::MenuCatalogHandler.new.(event)
```

paneq:
Related: #134 #177

paneq:
Generally that's around 10 lines of code:

* similar feature https://github.com/RailsEventStore/rails_event_store/pull/382

PR #255, opened 2018-03-30
WIP: Protobuf v2

paneq:
#### before

```ruby
      client = Client.new(repository: RailsEventStoreActiveRecord::EventRepository.new(
        mapper: RubyEventStore::Mappers::Protobuf.new,
      ))

      client.subscribe(->(ev){@ev = ev}, [ResTesting::OrderCreated])
      event = ResTesting::OrderCreated.new(
        event_id: "f90b8848-e478-47fe-9b4a-9f2a1d53622b",
        customer_id: 123,
        order_id: "K3THNX9",
      )

      client.publish_event(event, stream_name: 'test')
      expect(client.read_event(event.event_id)).to eq(event)
      expect(client.read_stream_events_forward('test')).to eq([event])
      expect(@ev).to equal(event)
```

#### after

```ruby
       client = Client.new(repository: RailsEventStoreActiveRecord::EventRepository.new(
        mapper: RubyEventStore::Mappers::Protobuf.new,
      ))

      client.subscribe(->(ev){@ev = ev}, [ResTesting::OrderCreated.descriptor.name])
      event = RubyEventStore::Proto.new(
        event_id: "f90b8848-e478-47fe-9b4a-9f2a1d53622b",
        data: ResTesting::OrderCreated.new(
          customer_id: 123,
          order_id: "K3THNX9",
        )
      )

      client.publish_event(event, stream_name: 'test')
      expect(client.read_event(event.event_id)).to eq(event)
      expect(client.read_stream_events_forward('test')).to eq([event])
      expect(@ev).to eq(event)
```


paneq:
@andrzejkrzywda @mpraglowski @pawelpacana @fidel 

For discussion. I know the API seems probably worse now, but it would allow removing `add_metadata` from mappers again. We allow exchanging `data`, but we go back to where we were pre v0.27 regarding metadata.

I think it would be easier to maintain for now.

The code shows how we could start using `type` string which could be interchangeable between many apps over class name and it's coupling which is only good inside a single app (unless they share code via gems).

valscion:
This looks so much nicer than the v0.27 API. Looking forward to this :+1:

Would it be possible to configure JSON for the `metadata` format as a user, or would you consider it an implementation detail and non-configurable?

paneq:
> This looks so much nicer than the v0.27 API. Looking forward to this 

wow, thanks. I was worried we are doing a step back. And async handlers would work :)

> Would it be possible to configure JSON for the metadata format as a user, or would you consider it an implementation detail and non-configurable?

You touched on a hard topic that I've been playing with for an hour or two recently. And I have some thoughts that I will try to convert into a separate issue.

Basically I am starting to feel that we cannot allow _anything_ in metadata.

I was playing with _ Can we use Protobuf's ANY data-type to serialize metadata?_ (listed on #228 as last task) and I realized that it's much harder to do in Protobuf because the data is unstructured. We allow everything in there, including non-primitive ruby types.

So here are a few thoughts in random order:

* Some serialization formats are schema-less and some require knowing the schema upfront. Messagepack can take whatever you give it, protobuf requires knowing what's there and special care.
* Some serialization formats require primitive types (JSON/messagepack), some can handle objects and when you desierialize you get back the same object, not a Hash
* Not every format can symetrically handle `Time` which we use in `metadata[:timestamp]`.
* `data` has only meaning for the application, `metadata` not necessarily
* Let's say I build HTTP JSON Api for exposing events for another app. I can (and I must) send serialized `data` in the same format that was stored in DB. The client will deserialize based on event type. But what about `metadata` ? How do I send it? Serialized in whatever format it was stored? 

    If I deserialize it, then I need to re-serialize it again before sending over the network. Should I maybe send it as JSON since the whole API is in JSON.
* We have `event_type` field in DB which we can use to store the type of `data`, but we don't have such field for `metadata`. Why? Do we need to? What does it tell us? Perhaps there is only one possible way to serialize metadata? There is only 1 type of _envelope_ for all events? What is the schema of the envelope?

I think I am going to research how other systems handle it, but I am leaning towards a conclusions that we should allow only some primitive types to be added to metadata and that metadata is sent in the same format the request asked for (JSON/XML) and not in the format it was serialized which is internal implementation detail.

Maybe I am wrong. I consider merging it as it is with `YAML` and dealing with it later in next release maybe when I invent something smarter. Right now I feel stuck with how to serialize metadata without knowing it's schema. 

We only have 3 fields right now in metadata but more info can be provided by the developer. Maybe a metadata field needs to be _registered_ before it can be used?

If we limit `metadata` to Hash without nested hashes with strings as keys and scalar values as types (string, int, float, bool, maybe date, maybe time) then perhaps this would be simple enough for any serializer to deal with it (protobuf could represent it as 5 separate hash maps). That's quite a limitation. I don't know yet if that's necessary or not.

Maybe I am overcomplicating, dunno. 

> Would it be possible to configure JSON for the metadata format as a user

So it all goes into, how do i serialize Time or any ruby object in metadata as JSON so it deserializes to the same thing? Maybe it does not need to deserialize to the same thing and I am unnecessarily constraining myself?

That's the long story

valscion:
Thanks for the thorough reply and for sharing your thoughts on this issue ☺️.

Regarding metadata vs. data, I think that they should be used for two completely different usecases. Metadata feels more like "logging and debugging", and if I'd ever have code depending on what's in metadata, I'd move it to `data` instead.

> I consider merging it as it is with `YAML` and dealing with it later in next release maybe when I invent something smarter.

The way I see it is that `YAML` is much more opinionated in how things are represented internally and it feels a tad dangerous to use it. I'd love it if I didn't have to worry about some keys being `String` and some `Symbol` in the persisted data, but `YAML` gives me no such choice — it gives me too much power.

```rb
YAML.dump(foo: 'bar') != YAML.dump('foo' => 'bar')
JSON.dump(foo: 'bar') == JSON.dump('foo' => 'bar')
```

As I consider `metadata` to be useful for mostly logging/debugging purposes, it would be a tad strange to see the following stored in the database:

```yaml
!ruby/object:DateTime 2018-04-03 13:39:18.847172000 +03:00
```

I get why this format would be good for `data` when I'd need to be able to load the same data back as Ruby objects for further processing. I don't see the reason why I'd need to do the same for `metadata` — having it in JSON format would feel better in that case:

```js
"2018-04-03T13:39:18+03:00"
```



> Maybe I am overcomplicating, dunno.

Yeah, I understand the feeling. To battle overcomplication, I wouldn't personally use `YAML` but if that's what you feel like is the better solution to go forward with, we might be fine with storing `YAML` in database next to the encoded protobuf `data` column.

> Let's say I build HTTP JSON Api for exposing events for another app. I can (and I must) send serialized data in the same format that was stored in DB. The client will deserialize based on event type. But what about metadata ? How do I send it? Serialized in whatever format it was stored?

Why would you need to send the metadata? If metadata contains vital information for other applications, shouldn't such data be in the `data` section instead? It would feel weird to have to document the format of `metadata` alongside the format of `data` in API documentation.

> We only have 3 fields right now in metadata but more info can be provided by the developer. Maybe a metadata field needs to be registered before it can be used?

I think the value of `metadata` lies in that it is much more loosely typed compared to `data` and acts instead of as a "bag of anything". This is how we've wrote one of our commands using [`dry-types`](http://dry-rb.org/gems/dry-types/):

```rb
class MyCommand < Dry::Struct::Value
  class Data < Dry::Struct::Value
    attribute :target_venue_id, Types::Strict::Int

    class RequestParams < Dry::Struct::Value
      constructor_type :strict_with_defaults

      attribute :author_name, Types::Strict::String
      attribute :event_type_id, Types::Coercible::Int
      attribute :author_locale, Types::Coercible::String
      attribute :author_site_key, Types::Coercible::String

      attribute :author_id, Types::Strict::Int.optional.default(nil)
    end

    attribute :request_params, RequestParams
  end

  attribute :data, Data
  attribute :metadata, Types::Strict::Hash
end
```

The `data` attribute is precisely defined to contain the attributes we wanted, and nothing else. And for `metadata`, we only care that it is a hash and that's it. It would be a pain to have to register all possible keys there. In v0.26, we even had to convert the `metadata` attribute from the class provided by `dry-types` to a Ruby `Hash` for the internal mutation to work:

https://github.com/RailsEventStore/rails_event_store/blob/6b82695f7df73f91be26ecf87155d2e9975d6e31/ruby_event_store/lib/ruby_event_store/client.rb#L184-L187

---

Anyway, that's my 2 cents. I'd love to continue the discussion if you feel like this provides you value :)

andrzejkrzywda:
`if I'd ever have code depending on what's in metadata, I'd move it to data instead` - I agree with this as a rule of thumb 👍

paneq:
>> if I'd ever have code depending on what's in metadata, I'd move it to data instead

> I agree with this as a rule of thumb 👍

That is a good rule for features built by your application but it seems to be contrary to automatic features that we would like to build into RES itself such as: #72 

Imagine async handler which based on `metadata[:request_ip]` or `metadata[:user_id]` or `metadata[:correlation_id]` builds stream such as `ip-127.0.0.1` with all linked events caused be someone from that ip address.

RES cannot assume anything about your `data`, it's class, format, conventions (see protobuf), serialization etc. But it would be nice if it could assume and read/access something from `metadata`. Otherwise what's the purpose of it at all? 

cc @andrzejkrzywda @valscion 

> `metadata.merge!(@metadata_proc.call || {}) if @metadata_proc`

I am gonna drop it in favor of limiting the interface to `[](key) and `[]=(key, val)`.

```ruby
metadata_proc.call.each {|k,v| metadata[k]=(v) }
```

> Why would you need to send the metadata?

In 1st version I will probably try to avoid the issue exactly by not sending metadata at all. However the other question is... Why should I not send metadata? I see no reason not to do it.

> If metadata contains vital information for other applications, shouldn't such data be in the data section instead?

If it's vital then indeed it should be inside. But it might be non-vital but still valuable? Maybe? Dunno 😈 😄 

> It would feel weird to have to document the format of metadata alongside the format of data in API documentation.

If the format was always the same format as how you consume the API (so JSON if you ask for JSON and XML if you use XML) then you would not need to document the format. Maybe schema but not the format itself. But if the schema is free and optional then there would be nothing to document.

> Thanks for the thorough reply and for sharing your thoughts on this issue

Thank you for going through my brain dumps as well :) I really appreciate your feedback.

valscion:
> That is a good rule for features built by your application but it seems to be contrary to automatic features that we would like to build into RES itself

Ha, good point :+1:

> Imagine async handler which based on metadata[:request_ip] or metadata[:user_id] or metadata[:correlation_id] builds stream such as ip-127.0.0.1 with all linked events caused be someone from that ip address.

That would be nice indeed. This feature would still work even if the metadata would be in JSON instead of YAML 😜 



> I am gonna drop it in favor of limiting the interface to `[](key)` and `[]=(key, val)`.

Sounds reasonable :+1:. It would still mean that `dry-types` would not be able to enforce the presence of any keys, only their type — they would anyway be filled in gradually instead of at construction time. And that's completely fine! Then the "registration of metadata values" would all have to be optional instead of required to be present. I can see the value in that — but such registration might be better suited to also be... optional :smile:


> If it's vital then indeed it should be inside. But it might be non-vital but still valuable? Maybe? Dunno 😈 😄

Yeah, metadata can definitely be valuable but maybe not application-logic-wise? It could help in debugging the sources of each data, or to help log things throughout a microservices architecture.

For an API consumer, I'd like them to know that "metadata will be a hash" and that "you can use it for debugging but don't use it for application logic — the data format can change at any time".

Dunno :smile:

> If the format was always the same format as how you consume the API (so JSON if you ask for JSON and XML if you use XML) then you would not need to document the format.

Yeah, that would definitely sound legit.


> Maybe schema but not the format itself. But if the schema is free and optional then there would be nothing to document.

I'm not sure if I understood this part, could you clarify?

---

One extra thing for JSON over YAML is that with PostgreSQL, JSON could be natively stored in the `metadata` column whereas YAML parsing is strictly reserved for the application that stored the value in the first place.

I fear that YAML is even easy to break between two consecutive versions of the same application, given how it encodes the class names inside the dumped data, too 😅

paneq:
> I'm only worried about having to to use YAML for metadata when data itself is serialized with protobuf

I tried to serialized metadata with protobuf but with current state of thing I failed. This is the source of my thoughts :)

> Maybe it would even be reasonable to only allow an application-wide metadata schema to be defined? Or we could allow it to be configured on application-level, while on event-level we could allow overriding the schema for metadata?

I realized that because we don't have `event_type` for `metadata` we can right now use only 1 definition in an entire app. Alternatively the mapper needs to define the schema and metadata in 1 column like I showed for JSON.

Overriding on event level is not easy (I thought about this as well and I tried to code it) exactly because we don't have a column to store the name of schema.

> Do we need to decide on this yet, or can we delay the decision to a future time when we've got more knowledge on how people are using protobuf in real applications?

I don't see this issue as protobuf only related. It's related to all mappers. I just comes out when we try to implement protobuf serialization due to its limitations and our hidden assumptions which now needs to become public and discussed.

> What's the cost we'd have to pay to delay making a decision?

The cost of shipping with YAML is:

* performance (YAML is slow)
* later changes would require a migration converting the serialized metadata from YAML to protobuf.

I would prefer to think 1-2 days on that, implement something in 1-2 days and ship something a bit better which uses protobuf to serialize both data and metadata. YAML is here so the tests are passing but I wanted to use protobuf.

I tried 3 approaches to do it in code and they all told me something about the design. It screams:

* 1 metadata schema for whole app (because we can't separately save type of metadata)
* We don't know what's being put into metadata. We allow everything.
* Everything does not scale well to other mappers.

@pawelpacana suggested also that he sees no need for nested hashes and that it can always be workarounded by devs by using namespaced keys `one.two.three` like Rack env does ie: `rack.hijack`.

So right now I am leaning towards flat structure and limited number of types.

valscion:
> So right now I am leaning towards flat structure and limited number of types.

Yeah no problem with that :)

---

So would you consider `metadata` to be private-ish for RES usage, as you would need some values from there for some potential future features? And `data` to be relevant to application developers. If that's the case, then I don't actually care what format `metadata` will be in :smile:

paneq:
> So would you consider metadata to be private-ish for RES usage, as you would need some values from there for some potential future features?

not really. If a developer adds `correlation_id` in metadata then the developer wants to get the `correlation_id` back and they relay on that being in the metadata. But it's not part of event data. It's part of the process around the event so it lands in `metadata`.

paneq:
@pawelpacana @andrzejsliwa @valscion @joelvh @EZku 

https://github.com/RailsEventStore/rails_event_store/pull/271

paneq:
@valscion @joelvh @Ezku to be continued on master

doc changes: separate PR https://github.com/RailsEventStore/rails_event_store/pull/278

PR #258, opened 2018-03-31
Fluent read API

pawelpacana:
Work-in-progress to have a collaborative checklist.

- [ ] event_repository using #read on facade methods
- [ ] legacy_event_repository using #read on facade methods
- [ ] deprecate old read methods on client
- [ ] rewrite API usage in specs, would be lovely to have a migrator tool that can be then used by end-users to migrate codebases as well
- [ ] document migration strategy from old to current reader API


paneq:
@joelvh observe it :) related to #258 and repository methods.

pawelpacana:
Rebased, will fail on some mutations on units. When green again, I'd assume _Phase 1_ is mergeable and possibly #268 could benefit from it.

_Phase 2_ to be continued after merge.

pawelpacana:
Would appreciate review @fmwr @fidel @paneq @joelvh :)

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -0,0 +1,73 @@
...
+      def head?
+        start.equal?(:head)
+      end
+
+      def forward?
+        direction.equal?(:forward)
+      end
+
+      def backward?
+        !forward?
```
joelvh: @pawelpacana I would swap the logic so that you check to see if direction is `:backward` and if not, falls to the default of `:forward`... but I do see that `Result` is private, so you're controlling this value.

joelvh:
@pawelpacana I like the improvement! The most recent refactor of #268 implements [`read` with keyword args](https://github.com/RailsEventStore/rails_event_store/pull/268/files#diff-cfd245fc7ed47bf94338566fe046349fR75) that mimic what the new `Specification` does.

Does the `EventRepository` still need all those helper methods that build and call the`Specification`s? It seems like the `Client` has a mirror of those methods and could implement the spec directly there. This would make the repository more streamlined and look VERY similar to the ROM repository linked above.

I had thought that `Specification` would be more like a value object to define a query, which is passed to the repository to be interpreted and executed. Is there a reason to have the spec execute the query internally?

If the spec does need to execute the query, I'd prefer to change it to not have any expectations of the object that executes the query. However, I'm not a huge fan of how `Specification` and `EventRepository` both know about each other.

Some suggested changes if specs need to execute themselves:

* Validation should not call `has_event?` and leave that validation to the repository (reduces need for multiple methods on query executor object)
* Specification should take in any callable object, so in this case, `EventRepository` responds to `call` or passes in `repository.method(:read)`

Lastly, calling `each` and then `to_a` is a bit odd to me, but I'm a fan of the "streaming" enumerator interface that provides more options for piping results to the client.

Overall, the changes are very similar to what I implemented with the ROM adapter in terms of API. I'd like to hear more of your thoughts on the interdependencies between `Client`, `EventRepository` and `Specification`.

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def limit(count)
+      raise InvalidPageSize unless count && count > 0
+      result.count = count
+      self
+    end
+
+    def each
+      @repository.read(result)
```
paneq: I expected to pass `self` but this is interesting.

ruby_event_store/spec/specification_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    specify do
+      repository = InMemoryRepository.new
+      test_event = TestEvent.new(event_id: '567ef3dd-dd28-4e05-9734-9353cd8653df')
+      specification = Specification.new(repository)
+      repository.append_to_stream([test_event], "fancy_stream", -1 )
+
+      expect(specification.from('567ef3dd-dd28-4e05-9734-9353cd8653df')).to have_result(:head?, false)
+    end
+
+    RSpec::Matchers.define :have_result do |attribute, expected|
```
paneq: nice helper

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
```diff
@@ -947,4 +947,8 @@ def self.new(
     end
   end
 
+  specify 'read returns enumerator' do
+    specification = RubyEventStore::Specification.new(repository)
+    expect(repository.read(specification.result)).to be_kind_of(Enumerator)
```
paneq: If the repository decided to be eager-loading instead of lazy-loading that's ok for me. I would expect `Enumerable`, not `Enumerator`.

ruby_event_store/lib/ruby_event_store/in_memory_repository.rb
```diff
@@ -113,8 +145,9 @@ def append(events, expected_version, stream_name, include_global)
           global_stream = read_stream_events_forward(GLOBAL_STREAM)
           raise EventDuplicatedInStream if global_stream.any? { |ev| ev.event_id.eql?(event.event_id) }
           global_stream.push(event)
+          @streams[GLOBAL_STREAM] = global_stream
```
paneq: I don't like it because #270 and it makes it `all` stream equal to global stream which is not necessary in InMemory implementation.

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -216,22 +212,5 @@ def enrich_event_metadata(event)
         @mapper.add_metadata(event, key, value)
       end
     end
-
```
paneq: good to see it moved/removed

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -46,35 +46,27 @@ def delete_stream(stream_name)
     end
 
     def read_events_forward(stream_name, start: :head, count: @page_size)
-      raise IncorrectStreamData if stream_name.nil? || stream_name.empty?
-      page = Page.new(@repository, start, count)
-      deserialized_events(@repository.read_events_forward(stream_name, page.start, page.count))
+      deserialized_events(read.stream(stream_name).limit(count).from(start).each)
```
paneq: I did not expect this... I thought we would

* Step 1
  * keep using old methods in repository such as read_events_forward
  * delegate read_events_forward in repo to read(specification)
  * deprecate read_events_forward
* Step 2
  * start using repo.read
  * remove read_events_forward from repo

but that doesn't make sense as we would be using deprecated methods internally so we would need to detect if repo support old or new way. Hmmm. Probably not worth the hassle

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -39,27 +39,54 @@ def last_stream_event(stream_name)
     end
 
     def read_events_forward(stream_name, after_event_id, count)
-      @repo_reader.read_events_forward(stream_name, after_event_id, count)
+      RubyEventStore::Specification.new(self)
+        .stream(stream_name)
+        .from(after_event_id)
+        .limit(count)
+        .each
+        .to_a
```
paneq: why is this `to_a` needed?

paneq:
> Does the EventRepository still need all those helper methods that build and call theSpecifications? 

I was wondering the same thing 

> It seems like the Client has a mirror of those methods and could implement the spec directly there. 

and I noticed that as well.

I belive this is necessary untill all tests in linter are re-written to use Specification

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
pawelpacana: Reminder that in that phase public API is not yet changed and the expectation is to get an array.  This fulfills this expectation.

pawelpacana:
> I belive this is necessary untill all tests in linter are re-written to use Specification

Exactly. We did not touch specs here and they both (client and repository) touch previous API with it's method signature and expected return values.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
pawelpacana: We'll change it when client (specs it that case) call `to_a`.

joelvh:
@pawelpacana to make sure that exception is raised, we should make sure the lint spec tests for it... but i think it might already with the current implementation.

ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
paneq: @pawelpacana I was wrong and right at the same time :)

if the Specification does:

```
def each
  @repository.read(result)
```

like it is doing right now, then we would need an `Enumerator`. Otherwise we would need to do `client.read.each.each do` because we are not passing a block

If the Specification did:

```
def each(&block)
  @repository.read(result).each(&block)
```

then we would need an `Enumerable` (and `Enumerator` is `Enumerable` but so is `Array`) and I think it makes most sense.

the Specification could also do:

```
def each(&block)
  @repository.read(result, &block)
```

but I don't like it.

paneq: Seprate topic: Should we include Enumerable to Specification?

joelvh: @paneq I had thought about including `Enumerable` in `Specification` too....

pawelpacana:
(test)

PR #259, opened 2018-03-31
Allow explicit read from GLOBAL_STREAM in InMemoryRepository

pawelpacana:
We allow reading from "all" stream like it is any other stream. Due to
implementation detail InMemoryRepository does not allow that.


pawelpacana:
Same thing happens in `LegacyEventRepository`, that is however expected there.

pawelpacana:
When `InMemoryRepository` behavior is now correct and consistent, regarding always appending to `all` stream, mutant (`Client` with `InMemoryRepository`) uncovers pretty interesting problem on `Client` API:

```diff
-def publish_events(events, stream_name: GLOBAL_STREAM, expected_version: :any)
+def publish_events(events, stream_name: nil, expected_version: :any) 
```

It does't matter what's default argument value, as we always publish in `GLOBAL_STREAM` (which is `all`). That actually makes sense, we didn't found earlier due to incorrect behavior of `InMemoryRepository` -- where it still assumed that it's appending either to passed stream or to `GLOBAL_STREAM`.

PR #260, opened 2018-03-31
Unstable order of read events from GLOBAL_STREAM.

pawelpacana:
When reading via dedicated reader for traversing all streams we sort
event records by "id" only.

For traversing particular stream we primarily sort by "position", in
addition to "id".

We allow explicitly passing "all" stream to traverse this particular
stream and this is where discrepancy in resulting order appears.

Not sure if RDBMS beyond sqlite are affected.


pawelpacana:
Passes, although now-disabled branch build show red.

Issue #257, opened 2018-03-31
Move serialization/deserialization of events above repository


paneq:
## Tasks

* [ ]  update documentation that mapper/serializer are passed directly to Client instead of via repo

### Event serialization/deserialization usecases

Event needs to be serialized/deserialized in those usecases:

* saving/retrieving from DB (current solution = repository+mapper)
* async handlers, scheduling and starting a job (current solution: YAML)
* exposing stream of events over network current solution would require getting events and then serializing them again to send over network).

quote from b36a5ab1 :

```
Two additional thoughts:

* maybe repository should return RubyEventStore::SerializedRecord
  and then the higher layer (Client) would use the :mapper to transform
  it to the right event object.

* similarly when saving we could do the same, the other way, Client
  would already use mapper to serialize the event and repository would
  receive RubyEventStore::SerializedRecord.

* The benefit would be that we could get RAW data from repository, not
  transform it to domain events and stream over network.

  Current solution require getting domain events (which are
  deserialized) and then serializing them again to send over wire.
```

I think this could simplify and optimize things. I already have a feeling we need to move `mapper` higher in our dependency tree.

paneq:
also if we could ask Client to get events in `RAW` format, without mapping then this could be beneficial for sending data over network as we would need no tricks to skip deserialize/serialize process.

paneq:
Now it seems obvious to me that serialization/deserialization is a concern above transportation/storage layer.

paneq:
@andrzejkrzywda @mpraglowski @pawelpacana @fidel @fmwr what do you think

paneq:
Another benefit would be that transactions would be shorter as serialization would occur before them,

PR #264, opened 2018-04-01
(de)serialization above repository

paneq:
Issue: #257

### DIFF
--- a/rails_event_store/lib/rails_event_store/client.rb
+++ b/rails_event_store/lib/rails_event_store/client.rb
@@ -1,10 +1,12 @@
 module RailsEventStore
   class Client < RubyEventStore::Client
     def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
+                   mapper: RubyEventStore::Mappers::Default.new,
                    event_broker: EventBroker.new(dispatcher: ActiveJobDispatcher.new),
                    page_size: PAGE_SIZE)
       capture_metadata = ->{ Thread.current[:rails_event_store] }
       super(repository: repository,
+            mapper: mapper,
             event_broker: event_broker,
             page_size: page_size,
             metadata_proc: capture_metadata)
--- a/rails_event_store/spec/protobuf_integration_spec.rb
+++ b/rails_event_store/spec/protobuf_integration_spec.rb
@@ -4,9 +4,10 @@ require_relative '../../ruby_event_store/spec/mappers/events_pb'
 module RailsEventStore
   RSpec.describe Client do
     specify 'can handle protobuf event class instead of RubyEventStore::Event' do
-      client = Client.new(repository: RailsEventStoreActiveRecord::EventRepository.new(
+      client = Client.new(
         mapper: RubyEventStore::Mappers::Protobuf.new,
-      ))
+        repository: RailsEventStoreActiveRecord::EventRepository.new,
+      )
       client.subscribe(->(ev){@ev = ev}, [ResTesting::OrderCreated])
       event = ResTesting::OrderCreated.new(
         event_id: "f90b8848-e478-47fe-9b4a-9f2a1d53622b",
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -5,10 +5,9 @@ module RailsEventStoreActiveRecord
 
     POSITION_SHIFT = 1
 
-    def initialize(mapper: RubyEventStore::Mappers::Default.new)
+    def initialize
       verify_correct_schema_present
-      @mapper      = mapper
-      @repo_reader = EventRepositoryReader.new(mapper)
+      @repo_reader = EventRepositoryReader.new
     end
 
     def append_to_stream(events, stream_name, expected_version)
@@ -137,8 +136,8 @@ module RailsEventStoreActiveRecord
       IndexViolationDetector.new.detect(message)
     end
 
-    def build_event_record(event)
-      serialized_record = @mapper.event_to_serialized_record(event)
+    def build_event_record(serialized_record)
+      # serialized_record = @mapper.event_to_serialized_record(event)
       Event.new(
         id:         serialized_record.event_id,
         data:       serialized_record.data,
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
@@ -1,10 +1,6 @@
 module RailsEventStoreActiveRecord
   class EventRepositoryReader
 
-    def initialize(mapper)
-      @mapper = mapper
-    end
-
     def has_event?(event_id)
       Event.exists?(id: event_id)
     end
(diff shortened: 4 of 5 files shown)


paneq:
@andrzejkrzywda @mpraglowski @pawelpacana @fidel @fmwr this is what I meant in #257

paneq:
also cc @joelvh

joelvh:
@paneq I'm glad to see this! I was already doing this with the ROM adapter, deserializing in `EventRepository`: https://github.com/joelvh/rails_event_store/blob/feature/rom_adapter/ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/event_repository.rb

And the ROM "reader" repository uses `for_deserialization` helper to return data in the format needed for deserialization (mostly renaming `id` to `event_id` for compatibility): https://github.com/joelvh/rails_event_store/blob/feature/rom_adapter/ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/rom/events_repository.rb#L79

You can see in there it felt like some sort of "read" spec would make sense, since the queries are so similar for each method.

paneq:
Tests are passing, we might need to kill a few new mutation tests though after the re-arrangment. I consider this experiment to be successful and soon to be merged.

joelvh:
@paneq thanks! I've been building around `RubyEventStore::SerializedRecord`, so this should be a pretty clean update for the `EventRepository`.

I see various test updates. Can any of the tests be moved to the lint spec, or are they AR-specific? I'll do a diff to see what the changes are, to port them over to ROM, but ideally most of the necessary tests are shared.

joelvh:
@paneq #268 updated with serialization and test suite changes.

paneq:
@joelvh 

> I see various test updates. Can any of the tests be moved to the lint spec

all tests which are shareable are in lint spec. The rest is AR specific.

joelvh:
@paneq @pawelpacana I've begun to wonder what the purpose of this serialization is. It seems like it's purpose is to serialize event data for persistence, is that correct? If so, is the persistence schema assumed to be a specific thing? Why is serialization not inside the persistence layer instead of at the client?

There's an awkward test that I took from AR and adapted to ROM. It inserts events into the database directly and then retrieves them through the `EventRepository` to make sure that they're retrieved as expected:

https://github.com/joelvh/rails_event_store/blob/feature/rom_adapter/ruby_event_store_rom_sql/spec/event_repository_spec.rb#L57

What's awkward is that I have to assume the data is serialized YAML. Also, the latest changes require a serialized event record to have all string values. What if I want to persist the `data` and `metadata` as JSONB or something other than a string?

Am I overlooking something obvious?

paneq:
@joelvh 

> I've begun to wonder what the purpose of this serialization is. It seems like it's purpose is to serialize event data for persistence, is that correct?

As explained in #257 it turned out there are 3 usecases for serialization

* saving in DB
* sending over network
* sending for bg job processing (which is basically same as sending over network)

> Why is serialization not inside the persistence layer instead of at the client?

Because as these 3 usecases indicate, only 1 of them is related to persistance, the other 2 are not. That's why I moved mapper above repository. That was my response to this relization and a big _aha_ moment.

And I believe other systems like rabbit or kafka or eventstore.org work in a similar way. You serialize your events before sending them over network to store in those systems. That also made me think this is the right direction.

> What's awkward is that I have to assume the data is serialized YAML

All 3 repositories that we have (InMemory, AR:EventRepository, AR::LegacyRepository) no longer assume any serialized format. They receive some binary data (which is string in Ruby basically) and return binary data.

> the latest changes require a serialized event record to have all string values. 

Not really. If you don't like `SerializedRecord` or you can't work with it or you want more than just 4Strings that it contains (let's say your repository needs `user_id` as well) then you can provide your own mapper which will map to `MySerializedRecordWithDifferentAttributes` and have your repository assume it receives `MySerializedRecordWithDifferentAttributes` instead of `SerializedRecord`.

For our needs `SerializedRecord` with `event_id, event_type, data, metadata` as Strings was good enough and it was minimal subset of fields necessary to have everything working.

> What if I want to persist the data and metadata as JSONB or something other than a string?

Provide your own `mapper` and `repository` when instantiating the `Client` if you want/need different behavior. But I don't know how anything other than String can be sent over network and I want to re-use mapper for that because it already knows how to give me Strings :)

Also, I assume if you send binary string to postgres for JSONB column, it will parse it internally and do the right thing. Feel free to experiment and verify this hypothesis. After all, doesn't your DB driver need to serialize data somehow to send over network to postgres as well?

## Summary

#### Before

* repository was receiving domain-event and returning domain-event
* repository had internally mapper as a dependency
* there was no way to map domain-event to strings
  * I could extend repository with methods to do that but I realized it's not a responsibility of the repository and they would delegate to mapper anyway
* You could change repository to different implementation
* If repository was taking mapper as dependency, you could exchange the mapper as well
* If repository was not taking mapper as dependency then it was hardcoded in repository how it transformed events to binary string (if needed) for saving

#### After

 * repository is receiving serialized-event and returning serialized-event
  * repository only needs to know how to persist binary data, it does not deal with serialization at all
* client has mapper as a dependency
* there is a way to map domain-event to strings 
  * by invoking mapper
* You can change repository to different implementation
* You can change mapper to different implementation
* If repository needs certain mapper with more data, you still can change repository and mapper together to different implementations and have them working properly together.

## Conclusion

I believe whatever was previously possible is still possible. But additionally a few simplifications have been made and some new features/use-cases/combinatinos are now possible. So for me it is a win.

Instead of 

```
client - > repo
repo -> serializer (internally)
```

we now have:

```
client -> serializer
client -> repo
```

paneq:
> f so, is the persistence schema assumed to be a specific thing?

Nope. LegacyRepository uses 1 table, AR::Event repository uses 2 tables, we've discussed using 3 tables as well. All of them have their own pros and cons.

joelvh:
@paneq thanks for giving a bit more history.

So what I'm hearing is that the mapper is actually paired with the repository implementation. It's used to map the domain event to the event format that the repository accepts? The use case where there's transport over the network involved is if the client retrieves events from the repository and sends them over HTTP or queues them for Sidekiq, and then the receiver would be using the same mapper to deserialize?

This seems to make the repository more related to the transport layer than the domain.

Let me know if that's on track....

paneq:
> So what I'm hearing is that the mapper is actually paired with the repository implementation. It's used to map the domain event to the event format that the repository accepts? 

I am not sure _paired_ is the right word. As there can be multiple repositories capable of using returned value from mapper. And multiple mappers capable of working with 1 repository.

In the end it's all duck-typing in Ruby. If your mapper returns something which quacks like what your repo expects then everything is fine.

In other language we would have an Interface that both adhere to but this is Ruby so I can't express that in the code. `mapper <- serialized object -> repo.` They agree on the format of the data exchanged between them.

> The use case where there's transport over the network involved is if the client retrieves events from the repository and sends them over HTTP or queues them for Sidekiq, and then the receiver would be using the same mapper to deserialize?

The flow is different in both usecases:

1. The client receives domain-event, serializes it, persists it, sends it to active job, and the job needs to deserialize it.
2. For sending over HTTP I am leaning towards getting serialized record from repoistory, (not deserializing it, or maybe only deserializing metadata, dunno) and sending over network. Then the client which receives the data deserializes it.

> This seems to make the repository more related to the transport layer than the domain.

the repository is persistence layer. But we don't have now the classic `domain object -> persisted data` flow. We have `domain object -> serialized object -> persisted data`. And the repository handles only the last part. 

Because sometimes we only need `domain object -> serialized object` this part (in `domain object -> serialized object -> queue` scenario) and sometimes we only need this part `persisted data -> serialized object` ( `persisted data -> serialized object -> network` scenario).

joelvh:
@paneq OK - I'll let this play out for me for now, since I need to see more use cases that are outside of my context. Are there examples anywhere that illustrate different scenarios?

paneq:
@joelvh 

* #228 - first thing that triggered me. We've been using YAML for serialization to active job but I realized that using `SerializedRecord` would make much more sense if we already have a mapper which can serialize and deserialize an event then there is no need to use something special for this usecase. 
* I am experimenting in private with exposing events over HTTP, no public code yet.

So I am afraid I can't provide more examples right now.

PR #268, opened 2018-04-03
ROM-SQL `EventRepository` Adapter

joelvh:
Implemented an `EventRepository` class with ROM SQL as an alternative to the ActiveRecord or `InMemory` versions. It uses Sequel under the hood and it should be possible to use it as a drop-in replacement for ActiveRecord in terms of being able to use existing tables in the database.

* Is not backwards-compatible with "legacy" schema
* Simply pass a ROM container to the `EventRepository` constructor
  * Alternatively, set the ROM container to `RubyEventStore::ROM.env` and that will be the global default
  * See `spec_helper` for more setup info


ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/rom/repositories/events_repository.rb
```diff
@@ -0,0 +1,113 @@
...
+          events_for(stream_name, :asc)
+        end
+  
+        def backward_for(stream_name)
+          events_for(stream_name, :desc)
+        end
+
+        def events_for(stream_name, direction)
+          order_columns = %i[position id]
+          order_columns.delete(:position) if stream_name == RubyEventStore::GLOBAL_STREAM
```
pawelpacana: 👍

joelvh:
@paneq @pawelpacana can you guys give some pointers on how to resolve mutation errors? It's not clear to me where to look for what's wrong and what a suggested fix is.

Thanks!

Here's an example:

```
evil:RubyEventStoreRomSql::EventRepository#link_to_stream:/Users/Joel/Code/projects/rails_event_store/ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/event_repository.rb:19:8aa0e
@@ -1,9 +1,9 @@
 def link_to_stream(event_ids, stream_name, expected_version)
   @events.detect_invalid_event_ids(Array(event_ids)).each do |id|
     raise(RubyEventStore::EventNotFound.new(id))
   end
-  add_to_stream(Array(event_ids), stream_name, expected_version, nil) do |event_id|
+  add_to_stream(event_ids, stream_name, expected_version, nil) do |event_id|
     event_id
   end
 end
-----------------------
Mutant configuration:
Matcher:         #<Mutant::Matcher::Config ignore_expressions: [RubyEventStoreRomSql::IndexViolationDetector#detect] match_expressions: [RubyEventStoreRomSql*]>
Integration:     Mutant::Integration::Rspec
Jobs:            8
Includes:        ["lib"]
Requires:        ["ruby_event_store_rom_sql"]
Subjects:        37
Mutations:       1645
Results:         590
Kills:           589
Alive:           1
Runtime:         62.53s
Killtime:        390.55s
Overhead:        -83.99%
Mutations/s:     9.44
Coverage:        99.83%
make: *** [mutate-fast] Error 1
```

ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/event_repository.rb
```diff
@@ -0,0 +1,151 @@
...
+                                                      stream_name.eql?(RubyEventStore::GLOBAL_STREAM)
+
+      event_ids = normalize_to_array(event_ids)
+      expected_version = normalize_expected_version(expected_version, stream_name)
+
+      # TODO: Move this into repository (don't expose db internals)!!
+      @event_streams.event_streams.transaction(savepoint: true) do
+        in_stream = event_ids.flat_map.with_index do |event_id, index|
+          position = compute_position(expected_version, index)
+          event_id = to_event_id.call(event_id)
```
joelvh: @pawelpacana @paneq this logic is taken from the AR version, where the events are persisted in the loop and the `event_id` returned. Is there any reason that the events aren't bulk-inserted into the database in one step, and then linking them to the stream in a separate step? Is it simply to reuse the same loop? (I don't know how often there's the scenario of adding more than a few events at a time, but separating the steps would make things a little clearer.

paneq:
> can you guys give some pointers on how to resolve mutation errors? It's not clear to me where to look for what's wrong and what a suggested fix is.

sure thing!

* https://github.com/mbj/mutant#reading-reports

here is what happenend. Mutant changed your code this way (diff format):

```
-  add_to_stream(Array(event_ids), stream_name, expected_version, nil) do |event_id|
+  add_to_stream(event_ids, stream_name, expected_version, nil) do |event_id|
```

and the tests are still passing. Now you need to meditate on what it means to you :)

Generally you can

* Say _that's cool_ and change the code to the code that mutant proposed. This sometimes happens. You call `to_s` but you realize the input is always a `String` so calling that is not necessary.
* Write additional test to prove why you need what did. In other words, mutant pointed a missing test. You are doing something but there is not a single test showing why you are doing it. 

    Here you can see https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store_active_record/spec/event_repository_spec.rb#L27 all the additional tests that we wrote for RailsEventStoreActiveRecord::EventRepository which are specific to its implementation. Some of them were to kill a mutation and some were not, just for additional safety. 

    For example we don't need to use `preload` and eager loading and everything would still work according to the spec (linter). But it would be slower. So have tests showing why we need `preload`.


Regarding this specific issue. I think you can assume you already got an Array: https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/client.rb#L28 . If you don't want to assume that then you need to write your own tests which passes a single event to justify why you need to convert it to an Array.

ruby_event_store_rom_sql/lib/ruby_event_store_rom_sql/event_repository.rb
paneq: > Is there any reason that the events aren't bulk-inserted into the database in one step

that would make sense. Historically it was only possible to store a single event. Later we added a possibly to handle many of them. The code might be this way due to this historical reasons only,

joelvh: @paneq refactored ROM to do things in batches.

joelvh:
OK, thanks @paneq for that overview -- makes sense based on my observations. Will look at your examples for resolving further.

I have been refactoring a bit to move away from the AR implementation's structure and more toward something similar to our conversations about a "read" client. You can see those updates on this PR

I also have some thoughts about making `expected_version` a value object because it requires lots of validation, processing and logic that is becoming spread out (at the least across multiple adapters). Started work on this, and just pushed that.

paneq:
> making expected_version a value object

sounds good. But I haven't looked yet :)

pawelpacana:
One more spec to enable from `event_repository_lint`: https://github.com/RailsEventStore/rails_event_store/commit/b91512a8f27346750eae4ce18c0bb86651f61c31#diff-7df8cd4b919a768031b808d31fc4e05bR25

And the spec was introduced here: https://github.com/RailsEventStore/rails_event_store/commit/fde402066891da7d46cff9b36f04da5cf0842c80

joelvh:
@pawelpacana done! Looks like tests are passing!

ruby_event_store_rom_sql/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+Further changes can be tracked at [releases page](https://github.com/RailsEventStore/rails_event_store/releases).
+
+### 0.1.0 (03.04.2018)
```
pawelpacana: We'll bump version to be in match with `RES_VERSION` on first release:
- https://github.com/RailsEventStore/rails_event_store/releases/tag/v0.15.0
- https://blog.arkency.com/why-we-follow-rails-repo-structure-in-rails-event-store/

ruby_event_store_rom_sql/lib/ruby_event_store/expected_version.rb
```diff
@@ -0,0 +1,57 @@
+module RubyEventStore
+  class ExpectedVersion
+    POSITION_DEFAULT = -1.freeze
```
pawelpacana: Minor but I believe integers are already frozen:

```
irb(main):001:0> -1.frozen?
=> true
```

ruby_event_store_rom_sql/lib/ruby_event_store/index_violation_detector.rb
```diff
@@ -0,0 +1,24 @@
+module RubyEventStore
+  class IndexViolationDetector
```
pawelpacana: Nice!

ruby_event_store_rom_sql/ruby_event_store_rom_sql.gemspec
```diff
@@ -0,0 +1,38 @@
...
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'ruby_event_store/rom/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'ruby_event_store_rom_sql'
+  spec.version       = RubyEventStore::ROM::VERSION
+  spec.licenses      = ['MIT']
+  spec.authors       = ['Joel Van Horn']
+  spec.email         = ['joel@joelvanhorn.com']
```
pawelpacana: I haven't thought much about it yet however we'll likely need `dev@arkency.com` user to be gem owner as well to handle releases of all gems at once smoothly.

pawelpacana:
Would merge if not the currently running build after push. Please do it yourself @joelvh when it passes again 🙂

ruby_event_store_rom_sql/lib/ruby_event_store/expected_version.rb
joelvh: @pawelpacana I want to say that the mutant yelled at me until I did this.

ruby_event_store_rom_sql/lib/ruby_event_store/index_violation_detector.rb
joelvh: @pawelpacana this is copied from AR ... thoughts on creating a common SQL library or something?

pawelpacana: would make sense to have it in AR as well after merge

skoona:
@joelvh I'll be looking over the PR this week and edit this comment; thanks for the effort.

joelvh:
@paneq any idea why mutant is still failing? I've merged in the latest from master and have implemented the timeout for specs.

joelvh:
@skoona sounds good - this is a very direct SQL implementation, but I'd like to see if it makes sense to make it more abstract ROM, so that you can configure SQL or other backends. Look forward to your feedback.

paneq:
@joelvh I restarted `res-rspec` job. It's 50/50% failing due what I believe is a MRI issue. As to `res-browser`: #277

pawelpacana:
Running failing mutation test on local machine -- PASS.

pawelpacana:
Annnnd it's done! 😎🎉

joelvh:
Woohoo! Thanks @paneq and @pawelpacana

PR #276, opened 2018-04-06
Stream as value object

pawelpacana:


pawelpacana:
Mutated all on dev machine and it passed. Sad to see CI producing so much false failures.

joelvh:
@pawelpacana should `GLOBAL_STREAM` be moved to `Stream`, such as `RubyEventStore::Stream::GLOBAL_NAME` or something like that?

PR #288, opened 2018-04-10
Remove ActiveSupport dependency

jandudulski:
`to_snake_case` is almost copied from Rails
[underscore](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-underscore)
implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/railseventstore/rails_event_store/288)


pawelpacana:
Didn't notice new commit -- assuming that'll pass on new CI infrastructure.

Issue #277, opened 2018-04-07
infinite mutations on master in rails_event_store-browser


paneq:
```
(103/1263) 100% - killtime: 42.26s runtime: 1520.70s overhead: 1478.44s
28
evil:RailsEventStore::Browser::EventsController#serialize_event:/home/travis/build/RailsEventStore/rails_event_store/rails_event_store-browser/app/controllers/rails_event_store/browser/events_controller.rb:15:c8513
104
evil:RailsEventStore::Browser::EventsController#show:/home/travis/build/RailsEventStore/rails_event_store/rails_event_store-browser/app/controllers/rails_event_store/browser/events_controller.rb:4:47261
```

paneq:
For the future - how to handle such case:

### 1

```
commit 1a5e7eb5202ae62cbd7f5041f1e21b0c17dac6d5 (HEAD -> mutant_debugging2, origin/mutant_debugging2)
Author: Robert Pankowecki <robert.pankowecki@gmail.com>
Date:   Sat Apr 7 09:22:41 2018 +0200

    Debug infinite mutations

diff --git a/rails_event_store-browser/Gemfile b/rails_event_store-browser/Gemfile
index a456956..fa727a9 100644
--- a/rails_event_store-browser/Gemfile
+++ b/rails_event_store-browser/Gemfile
@@ -10,3 +10,4 @@ gem 'bounded_context',                 path: '../bounded_context'
 if v = ENV['RAILS_VERSION']
   gem 'rails', v
 end
+gem 'mutant', github: 'paneq/mutant'
```

push and run on CI.
Check which mutation is being run inifinitely in the output

### 2

locally:

* bundle show mutant
* vim `lib/mutant/mutation.rb`

```
    # SHA1 sum of source and subject identification
    #
    # @return [String]
    def sha1
      ident = subject.identification.gsub("/Users/rupert/develop/rails_event_store", "/home/travis/build/RailsEventStore/rails_event_store") 
      Digest::SHA1.hexdigest(ident + CODE_DELIMITER + source)
    end
    memoize :sha1
```

edit sha1 to be stable between local env and CI

paneq:
### 3

locally

* vim lib/mutant/mutation.rb

```
    # Mutation code
    #
    # @return [String]
    def code
      c = sha1[CODE_RANGE]
      puts source if c == "c8513"
      c
    end
    memoize :code
```

`c8513` is from the output on CI :

```
28
[...]

PR #291, opened 2018-04-12
Mutation testing on CircleCI

paneq:
#277


rails_event_store_active_record/Makefile
```diff
@@ -14,18 +14,18 @@ install: ## Install gem dependencies
 
 test: ## Run unit tests
 	@echo "Running unit tests"
-	@bundle exec rspec
+	@BUNDLE_GEMFILE=Gemfile bundle exec rspec
```
pawelpacana: Was it necessary after all?

paneq: yes because otherwise it remembers (couldn't find where on filesystem...) the last one from

```
 install: ## Install gem dependencies
 	@echo "Installing gem dependencies"
 	@bundle install
 	@echo "Installing v1_v2_schema_migration gem dependencies"
 	@BUNDLE_GEMFILE=spec/v1_v2_schema_migration/Gemfile bundle install
 	@BUNDLE_GEMFILE=spec/without_rails/Gemfile bundle install
```

Issue #280, opened 2018-04-08
Use protobuf for serializing event metadata in protobuf mapper


paneq:
Drop using YAML for metadata

PR #294, opened 2018-04-13
experiment on binary encoding

MidnightWonderer:
Hi there,  
I am currently experimenting with the binary record encoding.  
I dropped support for rails < 5 and try optimizing primary key indexing as well as adding data compression.  
  
Practically I just encode uuid primary key as a 128-bit integer and use snappy compression gem to compress the compressible data.  
  
I planned on keeping it as my personal patch as this should not be merged upstream in any way.  
However, I open the PR for the sake of discussion.  
- Are the changes good or bad in your opinion?  
- What do I miss or what should I consider before going live with my patch?  
- Can anything be used by the upstream?  
- What can be improved?


paneq:
Are you targeting postgresql or mysql? if postgresql then why not just change column type to https://www.postgresql.org/docs/9.1/static/datatype-uuid.html ?

As to Snappy, what are you trying to achieve? What are you trying to compress mostly? Data or JSON boilerplate? If boilerplate then consider just using protobuf:

* docs: #278 #273 to be merged before we release new version
* code & issues: #189 #228 #280 #255 etc
* https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb

This snappy idea is interesting but with current architecture (on master, soon to be released as new gem version) you can just achieve it by passing your own Mapper as dependency, that's what we are doing with protobuf. There is no need to hardcode it into repository.

Regarding the changes you did for repository. You don't need to change our code. You can just pass implement your own repository like @joelvh did for ROM: #268 http://railseventstore.org/docs/repository/ . You can even re-use the tests that we have. And pass that repository as a dependency. If you feel generous, you could release it as separate gem.

I don't see us merging those changes, but there are potential for releasing separate gems that could be integrated by some interested developers by themselves. That's why we made the code extendible with dependencies being passed in a constructor.

It seems to me that you implemented 3 separate changes

* binary id

    I am not sure what's the win. Anyway, this could be implemented in custom mapper. 

    Read #278 #273 and http://railseventstore.org/docs/mapping_serialization/
    
    Mappers already convert domain event to 4 strings. You could convert ID to this binary using the algorithm you implemented.

* snappy compression.

    This could be generalized as well. You could write a mapper, which would take other mapper (we already have 2) and compress the string they returned for data and metadata. That would allow compressing YAML/JSON/Protobuf, anything. You could make the compression algorithm more general as well. Assume certain interface and make the object responsible for compression also a dependency.

* keeping data&metadata in a single column.

    I don't see value in that right now.

> Are the changes good or bad in your opinion?

My guess is you are trading data size for CPU. I am not sure why so it's hard to judge. I don't know your usecases.

From the point of single responsibility I believe you are going in the wrong direction.

I am not sure why for example CompressionSerializer is in RailsEventStoreActiveRecord . As a mapper it could work with any repository we have.

> Can anything be used by the upstream?

Maybe not upstream but side-stream. Compression could be optional as a mapper to configure. Your repository could be released as a separate gem by yourself. We could mention it in a documentation.

paneq:
@MidnightWonderer https://github.com/RailsEventStore/rails_event_store/issues/295 here :)

MidnightWonderer:
Hi, thank you a bunch.  
Very detailed reply  
  
The main reason I hardcoded the changes into the current gem is to reuse the test spec.  
I did not aware that I can reuse the spec as a dependency nor aware of the mapper functionality.  
So I glad I asked.  
  
My use case simply has a potential to generate tons of events. (read billions)  
I just believe rails instance scaling is easier than RDBMS scaling. So I bet on trading scalable CPU with hard to scale storage.  
It is just an exploring though.  
The combination of 2 column simply is to make the compression more efficient.  
  
Thank you again @paneq .  
I think I still have tons of reading to catch up with this RES ecosystem.

paneq:
> I did not aware that I can reuse the spec as a dependency

yeah, just have a look: 

* https://github.com/RailsEventStore/rails_event_store/blob/e30d4d64d17dc3e1a6cb2a102facd41bb77168aa/rails_event_store_active_record/spec/event_repository_spec.rb#L3
* https://github.com/RailsEventStore/rails_event_store/blob/e30d4d64d17dc3e1a6cb2a102facd41bb77168aa/rails_event_store_active_record/spec/event_repository_spec.rb#L27

>  nor aware of the mapper functionality.

just read the code :) it's a 200 lines for getting a top-level overview: https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/client.rb

and the whole documentation. It explains existing functionalities that you can swap.

> I just believe rails instance scaling is easier than RDBMS scaling. So I bet on trading scalable CPU with hard to scale storage.

Makes sense to me.

> The combination of 2 column simply is to make the compression more efficient.

I would verify that assumption. You will be maintaining your own repository which might make it harder to stay up to date with newest releases for what I guess would be negligible win.

You might also check it it isn't better to compress with built-in DB features: https://stackoverflow.com/questions/1369864/does-postgresql-support-transparent-compressing-of-tables-fragments

> I think I still have tons of reading to catch up with this RES ecosystem.

I am pretty sure one can catch up with the whole codebase and possibilities in half a day :)

PR #298, opened 2018-04-16
Browse global stream on start.

### LINKED ISSUE #212: Browser and trouble with efficiently browsing streams
Recently we've shipped neat stream and event browser UI for RES. It allows you to navigate through list of streams and list of events in selected stream, finally giving the ability to see particulat event in detail.

Problem: we don't have best database schema to implement stream browsing. Reminder that we have two tables: https://github.com/RailsEventStore/rails_event_store/blob/2eb821c73cc3c6aa63dfe3a9680a0b689ae71193/rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/migration_template.rb#L7-L27
In order to get a list of all streams one would have to `SELECT DISTINCT stream FROM event_store_events_in_streams`. Additionally it would be nice to present `all` stream at the beginning of this list. The comes the pagination in order to present/navigate though possibly thousands of streams.

What works for small number of streams in dev environement is sort-of-pagination in memory with `Enumerable`: 
https://github.com/RailsEventStore/rails_event_store/blob/2eb821c73cc3c6aa63dfe3a9680a0b689ae71193/rails_event_store-browser/app/controllers/rails_event_store/browser/streams_controller.rb#L8-L9
https://github.com/RailsEventStore/rails_event_store/blob/2eb821c73cc3c6aa63dfe3a9680a0b689ae71193/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb#L84-L88

That however has obvious drawback of being memory-hungry for production use with lots of streams.

Solutions I see to get over it:

0. Ignore and assumme this for dev usage. Likely not going to work 😉 

1. Drop stream browsing feature, leaving only event browsing for stream name you know upfront. Works when you know how stream names are chosen in your system and you can recostruct such name yourself.

Morever this is how geteventstore proposes:
ges-browser

One more example from custom audit log implementation:
bes-browser

So in essence there's input field and/or URL in which you could enter stream name in order to land on event browsing for given stream. More effort fo end user but less work on RES side and no one would accidentally OOM-kill app.

2. Implement streams table in form of database view. This would probably allow efficiently paginating like on events. Any drawbacks I failed to see here?

3. Implement `event_store_streams`. I guess it's the least popular and most troublesome option as it requires migration simillar to what we had in v0.18.0. Not sure if such feature like stream browser ios worth this much effort.

Any ideas?

pawelpacana:
- browser starts on global stream, which by design references all events
- still allows jumping into particular stream via URL
- removes troublesome get_all_streams API

Closes #212


pawelpacana:
@fmwr @paneq Please give it a look in the light of #212

paneq:
Spoko!

PR #300, opened 2018-04-17
Expected version everywhere

pawelpacana:


rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -68,13 +68,16 @@ def read_event(event_id)
 
     private
 
+    def version_resolver(stream)
```
joelvh: @pawelpacana any reason to not just make this an inline block to the `resolve_for` method or a lambda as an argument? Just a minor observation: reading the method name, it sounds like it will return a resolver when it actually returns a resolved value. With using a method reference (as you are), I might make that method named something that represents what it does... such as `resolve_version_for(stream)` or something like that. Just my 2 cents.

pawelpacana: 👍 

Actually it is `last_stream_version` and naming it accordingly is better.

PR #302, opened 2018-04-17
ROM on CI

### LINKED ISSUE #289: CI config matrix to include ruby_event_store_rom_sql
We need both: config for Travis CI and for soon-to-be-introduced Circle CI (mutation testing)

paneq:
Closes #289, #303

- [x] 100% mutant coverage
- [x] pass PostgresSQL specs
- [ ] pass MySQL specs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,4 +127,5 @@ workflows:
     - aggregate_root
     - rails_event_store-rspec
     - rails_event_store-browser
-    - bounded_context    - ruby_event_store-rom
+    - bounded_context
+    - ruby_event_store-rom
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,5 +127,4 @@ workflows:
     - aggregate_root
     - rails_event_store-rspec
     - rails_event_store-browser
-    - bounded_context
-    - ruby_event_store-rom
+    - bounded_context    - ruby_event_store-rom
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,6 +101,8 @@ matrix:
     env: TARGET=ruby_event_store-rom SCRIPT=test DATABASE_URL=sqlite:db.sqlite3
   - rvm: 2.4.3
     env: TARGET=ruby_event_store-rom SCRIPT=test DATABASE_URL=postgres://localhost/rails_event_store_active_record?pool=5
+  - rvm: 2.4.3
+    env: TARGET=ruby_event_store-rom SCRIPT=test DATABASE_URL=mysql2://root:@127.0.0.1/rails_event_store_active_record?pool=5
 
 before_script:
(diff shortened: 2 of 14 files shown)


pawelpacana:
@paneq rebased on current master (and force pushed), I hope you don't mind

joelvh:
@paneq please verify if the tasks for this issue are complete. Thanks!

pawelpacana:
@joelvh one fail on mysql https://travis-ci.org/RailsEventStore/rails_event_store/jobs/370456355

joelvh:
@paneq @pawelpacana ActiveRecord uses a mutex when running transactions: 

* https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L187
* https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L226

However, I saw that there was still a deadlock failure, so I will keep looking into this. My guess is AR is somehow more robust at handling this with MySQL than Sequel. Your thoughts are welcome for troubleshooting.

Thanks.

pawelpacana:
> ActiveRecord uses a mutex when running transactions

On the same connection, right? 

Just to clarify, in concurrency-related tests there's a connection per thread:
```
      def verify_conncurency_assumptions
        expect(ActiveRecord::Base.connection.pool.size).to eq(5)
      end
```
https://blog.arkency.com/2015/09/testing-race-conditions/

joelvh:
@pawelpacana @paneq it looks like the deadlock culprit is ROM’s Changeset feature, which is doing something funky under the hood that causes the issue. I’ve implemented a simple workaround that I'm going to commit shortly.

Sequel has a `retry_on` option, to retry a transaction on something like a deadlock, but it relies on handling a specific exception. Somehow ROM was blocking this and the deadlock exception bubbled up and was not handled. That said, working around the ROM Changeset internals, I was seeing deadlocks very rarely (which were automatically retried, allowing the tests to pass), and now I don't seem to see them at all.

Will see how CI fares with it.

joelvh:
Do you think AR is retrying transactions on MySQL deadlocks? Sequel clearly has an option to retry, so that solves the issue, but I've implemented that for MySQL specifically because PostgreSQL didn't have an issue.

joelvh:
Fixed config conflicts

joelvh:
Looks like things finally check out.

Spent a lot of time trying to diagnose the MySQL deadlocks. I saw a post about MySQL 5.7 seeming to get more deadlocks for one StackOverflow user, and the answer was that there was a new deadlock detector added. That can be disabled to rely on the lock timeout instead, which can be more performant. However, I believe this is a server-level configuration change.

That said, I don't know why AR isn't running into deadlocks while ROM-SQL is. Will have to see if this becomes an issue for MySQL users... (I personally use PostgreSQL these days).

Issue #303, opened 2018-04-17
ROM tests are not passing on travis CI for postgres and mysql


paneq:
https://travis-ci.org/RailsEventStore/rails_event_store/builds/367754123?utm_source=slack&utm_medium=notification

paneq:
```
     Failure/Error: expect(rom_db.connection.pool.size).to eq(5)
     
       expected: 5
            got: 1
```

@joelvh Do you know what's the culprit? should I provide a different `DATABASE_URL` for travis? code in: #302

joelvh:
@paneq fixed in #304

joelvh:
@paneq I think we can close this issue

paneq:
As far as I am aware we are still not testing rom on postgres and mysql on `master` branch on CI.

paneq:
Will be closed when #302 issues are addressed

Issue #301, opened 2018-04-17
34 alive mutations on rom-sql


paneq:

paneq:
I looked into a mutation changing `(savepoint: true)` to `(savepoint: false)` in:

```ruby
        def create(serialized_records, stream: nil, expected_version: nil)
          events.transaction(savepoint: true) do
            events.changeset(CreateEventsChangeset, serialized_records).commit.tap do
              if stream
                link(
                  serialized_records.map(&:event_id),
                  stream,
                  expected_version || ExpectedVersion.any,
                  global_stream: true
                )
              end
            end
          end
        end
```

and the  I don't know why the test does not kill the mutation.

```ruby
    specify "nested transaction - events still not persisted if append failed" do
      repository = EventRepository.new
      repository.append_to_stream([
        event = SRecord.new(event_id: SecureRandom.uuid),
      ], default_stream, RubyEventStore::ExpectedVersion.none)

      rom_db.transaction do
        expect do
          repository.append_to_stream([
            SRecord.new(
              event_id: '9bedf448-e4d0-41a3-a8cd-f94aec7aa763'
            ),
          ], default_stream, RubyEventStore::ExpectedVersion.none)
        end.to raise_error(RubyEventStore::WrongExpectedEventVersion)
        expect(repository.has_event?('9bedf448-e4d0-41a3-a8cd-f94aec7aa763')).to be_falsey
        expect(repository.read_all_streams_forward(:head, 2)).to eq([event])
      end
      expect(repository.has_event?('9bedf448-e4d0-41a3-a8cd-f94aec7aa763')).to be_falsey
      expect(repository.read_all_streams_forward(:head, 2)).to eq([event])
    end
```

logs (when false):

```
I, [2018-04-17T18:43:31.098243 #10802]  INFO -- : (0.000028s) BEGIN
I, [2018-04-17T18:43:31.099967 #10802]  INFO -- : (0.000044s) INSERT INTO `event_store_events` (`id`, `event_type`, `metadata`, `data`, `created_at`) VALUES ('9bedf448-e4d0-41a3-a8cd-f94aec7aa763', 'ce40960d-52e0-4ce8-a220-87aa156cccfa', 'c8335ad6-dd35-4ea7-a7a2-ffdf89b74779', 'c8add981-9923-4a19-86dc-cb2aa3318ab9', '2018-04-17 16:43:31.099786')
[...]

PR #304, opened 2018-04-17
Fix database variations for ROM

joelvh:
Addresses failing tests in #303 and also removes tests for internals noted in #301


ruby_event_store_rom_sql/spec/rom/event_repository_spec.rb
```diff
@@ -237,15 +237,15 @@ def cleanup_concurrency_test
     end
 
     def verify_conncurency_assumptions
-      expect(rom_db.connection.pool.size).to eq(5)
+      expect(rom_db.connection.pool.size).to eq(1)
```
paneq: that does not look right. if there is 1 connection in a pool then the multi-threading tests don't make much sense. Is there a way for us to configure more in tests? Like 5 maybe?

joelvh: I wasn’t able to find any Sequel/ROM equivalents to this and not sure of the total context of this test either.

paneq: ah yes, there is:


:preconnect | Whether to automatically make the maximum number of connections when setting up the pool. Can be set to “concurrently” to connect in parallel.

we should use it :)

joelvh: I had looked at the docs, but wasn't sure what specifically was being tested. I haven't played with these settings, but will look into it a bit more.

joelvh:
@paneq added concurrent connections!

PR #307, opened 2018-04-20
Protobuf metadata

paneq:


paneq:
@valscion care to review?

ruby_event_store/spec/mappers/protobuf_spec.rb
```diff
@@ -150,50 +151,72 @@ module RubyEventStore
...
-      let(:proto) { "\n\aK3THNX9\x10{" }
+
+      specify "initialize requires protobuf_nested_struct" do
+        p = Protobuf.allocate
+        def p.require(_name)
+          raise LoadError
+        end
+        expect do
+          p.send(:initialize)
+        end.to raise_error(LoadError, "cannot load such file -- protobuf_nested_struct. Add protobuf_nested_struct gem to Gemfile")
```
valscion: Seems like `LoadError` isn't captured with the `expect` block but it fails the CI currently

ruby_event_store/spec/mappers/protobuf_spec.rb
```diff
@@ -150,50 +151,72 @@ module RubyEventStore
...
   module Mappers
     RSpec.describe Protobuf do
       let(:event_id)     { "f90b8848-e478-47fe-9b4a-9f2a1d53622b" }
+      let(:metadata) { {
+        one: 1,
+        two: 2.0,
+        three: true,
+        # four: Date.new(2018, 4, 17),
+        five: "five",
+        # six: Time.utc(2018, 12, 13, 11 ),
```
valscion: Hmm could you shed some light on why these are commented out? Was there something in the gem that didn't work and so these types had to be skipped for now?

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -39,13 +40,14 @@ def pool
...
       def initialize(events_class_remapping: {})
+        require_optional_dependency
         @events_class_remapping = events_class_remapping
       end
 
       def event_to_serialized_record(domain_event)
         SerializedRecord.new(
           event_id:   domain_event.event_id,
-          metadata:   YAML.dump(domain_event.metadata),
+          metadata:   ProtobufNestedStruct::HashMapStringValue.dump(domain_event.metadata.each_with_object({}){|(k,v),h| h[k.to_s] =v }),
```
valscion: So did I get this right:

When we are going _to_  the protobuf world, we need the hash keys to all be of type `String` and that's why we need to call `k.to_s` here?

And when we are in the `serialized_record_to_event` method, we need to do the opposite, i.e. call `k.to_sym` as the `Metadata` class requires keys to be symbols?

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -1,8 +1,8 @@
 module RubyEventStore
   class Proto < RubyEventStore::Event
-    def initialize(event_id: SecureRandom.uuid, metadata: {}, data:)
+    def initialize(event_id: SecureRandom.uuid, metadata: nil, data:)
       @event_id = event_id
-      @metadata = metadata
+      @metadata = Metadata.new(metadata.to_h)
```
valscion: Oh wow, today I learned that `nil.to_h` returns an empty hash. Huh.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -12,14 +12,15 @@ def type
...
+      coder['metadata']   = ProtobufNestedStruct::HashMapStringValue.dump(metadata.each_with_object({}){|(k,v),h| h[k.to_s] =v })
       coder['data.proto'] = data.class.encode(data)
       coder['data.type']  = type
     end
 
     def init_with(coder)
       @event_id = coder['event_id']
-      @metadata = coder['metadata']
+      @metadata = Metadata.new
+      ProtobufNestedStruct::HashMapStringValue.load(coder['metadata']).each_with_object(metadata){|(k,v),meta| meta[k.to_sym] = v }
```
valscion: Would it be more obvious that the `metadata` in `each_with_object(metadata)` also used the instance variable prefix `@`? It feels a bit confusing that one line above, we create the `@metadata` but then in here we don't use it yet.

I'm not sure about your coding style, so this comment might be irrelevant. Let me know what you think :relaxed:

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
paneq: yep, I don't like it as well. but `Metadata` is already using Symbols so...

ruby_event_store/spec/mappers/protobuf_spec.rb
paneq: No reason, I should uncomment them.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
valscion: Yeah, just making sure I followed this logic. Nothing wrong with it, even though it means we have to jump through a few loops here :)

ruby_event_store/spec/mappers/protobuf_spec.rb
valscion: Ok great ☺️ I was worried that these types didn't work for some reason, glad that it wasn't so

ruby_event_store/spec/mappers/protobuf_spec.rb
paneq: This works fine. `rails_event_store` specs failed because `Gemfile` did not contain `protobuf_nested_struct`. Easy fix.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
paneq: indeed :)

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
paneq: Mutant mutates `@metadata` to `metadata` and it still works so it complains. Minor ;)

paneq:
@valscion unfortunately https://github.com/RailsEventStore/rails_event_store/issues/308 happened :)

PR #312, opened 2018-04-24
POC of no EventInStream entries for GLOBAL_STREAM

pawelpacana:
Failing, but should be complete enough to present idea.

- no reserved stream name ("all")
- one insert less per append_to_stream
- boolean param is no more ;)

Successor of #141.


paneq:
@pawelpacana green light, does not matter anymore :)

mpraglowski:
Continued on #514

Issue #325, opened 2018-04-27
Add ability to dynamically set metadata for all events


paneq:
instead of:

```ruby
begin
  Thread.current[:rails_event_store][:correlation_id] = event.event_id
  Thread.current[:rails_event_store][:process_id] = "Manager X"
  command_bus.execute(command)
ensure
  Thread.current[:rails_event_store][:correlation_id] = nil
  Thread.current[:rails_event_store][:process_id] = nil
```

we could do:

```ruby
  event_store.with_metadata(correlation_id: ...) do
    command_bus.execute(command)
  end
```

PR #327, opened 2018-04-30
Allow to dynamically enrich events metadata by using `with_metadata`

jakubkosinski:
#325 

Instead of using metadata_proc when initializing RubyEventStore::Client one can now
use with_metadata instance method to dynamically set the event metadata for all events
published inside a block, e.g.

```
event_store.with_metadata(request_ip: '127.0.0.1', correlation_id: '1234567890') do
event_store.publish(event)
end
```

`with_metadata` calls can be nested but the passed metadata will not be merged with one from
the upper level. This allows to clear the metadata set in the upper level by calling `with_metadata(nil)`
- in this case only default timestamp will be added to published events metadata


ruby_event_store/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 Further changes can be tracked at [releases page](https://github.com/RailsEventStore/rails_event_store/releases).
 
+### 0.15.0 (not released yet)
```
paneq: We've moved changelog to https://github.com/RailsEventStore/rails_event_store/releases as the 1st line says :)

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -198,13 +208,27 @@ def normalize_to_array(events)
     end
 
     def enrich_event_metadata(event)
-      event.metadata[:timestamp] ||= clock.()
       if metadata_proc
+        warn "`RubyEventStore::Client#metadata_proc` has been deprecated. Use `RubyEventStore::Client#with_metadata` instead."
```
paneq: I am not sure we want to deprecate it... Let me think/check.

paneq: hmm too compilicated for quick review:

* https://github.com/RailsEventStore/rails_event_store/blob/3b191017dd854607b88c5e01e9328a4f13789a98/rails_event_store/lib/rails_event_store/middleware.rb#L9
* https://github.com/RailsEventStore/rails_event_store/blob/3b191017dd854607b88c5e01e9328a4f13789a98/rails_event_store/lib/rails_event_store/client.rb#L7

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -171,6 +171,16 @@ def within(&block)
       Within.new(block, event_broker)
     end
 
+    def with_metadata(metadata, &block)
+      previous_metadata = self.metadata
+      begin
```
paneq: i think we could use just `ensure` without `begin`

jakubkosinski: yes, you're right

rails_event_store/spec/middleware_spec.rb
```diff
@@ -1,42 +1,77 @@
...
 require 'rails_event_store/middleware'
 require 'rack/lint'
 
 module RailsEventStore
   RSpec.describe Middleware do
-    specify 'lint' do
-      request = ::Rack::MockRequest.new(::Rack::Lint.new(Middleware.new(
-        ->(env) { [200, {}, ['Hello World']] },
-        ->(env) { { kaka: 'dudu' } } )))
+    specify 'calls app within with_metadata block when app has configured the event store instance' do
```
pawelpacana: Isn't that a bit too much of testing implementation details, something which one would call _klasotesty_?

rails_event_store/spec/middleware_spec.rb
```diff
@@ -1,42 +1,77 @@
...
+require 'action_controller/railtie'
+require 'rails_event_store/railtie'
+require 'securerandom'
 require 'rails_event_store/middleware'
 require 'rack/lint'
 
 module RailsEventStore
   RSpec.describe Middleware do
-    specify 'lint' do
-      request = ::Rack::MockRequest.new(::Rack::Lint.new(Middleware.new(
```
pawelpacana: Why have you decided to remove`Rack::Lint` that [verifies](https://www.rubydoc.info/gems/rack/Rack/Lint) Middleware in accordance to `Rack` specification?

rails_event_store/spec/middleware_spec.rb
pawelpacana: nvmd, changed later

pawelpacana:
Looks good, the questions remain open and can be addressed async -- as usual.

PR #328, opened 2018-05-01
Client read API

pawelpacana:


aggregate_root/spec/aggregate_root_spec.rb
```diff
@@ -69,15 +69,15 @@
...
+    expect(event_store.read.stream(stream).each.to_a).to eq [order_created]
 
     restored_order = Order.new.load(stream, event_store: event_store)
     expect(restored_order.status).to eq :created
     order_expired = Orders::Events::OrderExpired.new
     restored_order.apply(order_expired)
     restored_order.store(stream, event_store: event_store)
 
-    expect(event_store.read_stream_events_forward(stream)).to eq [order_created, order_expired]
+    expect(event_store.read.stream(stream).each.to_a).to eq [order_created, order_expired]
```
mpraglowski: `each.to_a` look strange for me, full enumerable implementation maybe? (separate PR of course)

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -31,13 +31,14 @@ def backward?
...
-    def initialize(repository, result = Result.new(:forward, :head, NO_LIMIT, Stream.new(GLOBAL_STREAM)))
+    def initialize(repository, mapper, result = Result.new(:forward, :head, NO_LIMIT, Stream.new(GLOBAL_STREAM)))
+      @mapper = mapper
       @repository  = repository
       @result = result
     end
 
     def stream(stream_name)
-      Specification.new(repository, result.dup.tap { |r| r.stream = Stream.new(stream_name) })
+      Specification.new(repository, mapper, result.dup.tap { |r| r.stream = Stream.new(stream_name) })
```
mpraglowski: Maybe `result.dup.tap { |r| r.stream = Stream.new(stream_name)` (and similar) should be extracted to:

`result.with_stream(steam_name)` where implementation would be

```
def with_stream(stream_name)
  self.dup.tap { |r| r.stream = Stream.new(stream_name)
end
```

PR #357, opened 2018-05-15
Batch reader

fidel:
We decided to go with `in_batches` method (inspired by what's inside Rails), and the API looks like that: 
- `stream("foo").forward.each` - unlimited
- `stream("foo").forward.in_batches.each` - batched, default size
- `stream("foo").forward.in_batches(1000).each` - batched, custom size

ActiveRecord & InMemory repositories are done, ROM is in progress (tests passing, some mutations to kill).

It's also probably a good moment to drop LegacyEventRepository 😉


pawelpacana:
> It's also probably a good moment to drop LegacyEventRepository

Not now, added support.

PR #360, opened 2018-05-19
AggregatRoot loads events in batches

fidel:
To prevent killing someones app by performing massive query via repository. 

We decided to ignore it since it's very hard to satisfy mutant within current `AggregateRoot` design. To kill the mutation we would need to be sure that *Repository performs certain operation. See df502227d5d5dae26ad0e88420a99473fc5d8206 for reference.

### DIFF
--- a/aggregate_root/Makefile
+++ b/aggregate_root/Makefile
@@ -1,7 +1,8 @@
 GEM_VERSION = $(shell cat ../RES_VERSION)
 GEM_NAME    = aggregate_root
 REQUIRE     = $(GEM_NAME)
-IGNORE      =
+IGNORE      = AggregateRoot\#events_enumerator
+
 SUBJECT     ?= AggregateRoot*
 
 install: ## Install gem dependencies
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -36,7 +36,7 @@ module AggregateRoot
 
   def load(stream_name, event_store: default_event_store)
     @loaded_from_stream_name = stream_name
-    event_store.read.stream(stream_name).each.with_index do |event, index|
+    events_enumerator(event_store, stream_name).with_index do |event, index|
       apply(event)
       @version = index
     end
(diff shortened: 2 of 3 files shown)


paneq:
> We decided to ignore it since it's very hard to satisfy mutant within current AggregateRoot design.

If `RubyEventStore::BatchEnumerator` was executed above the repository and called `repository.read(spec)` where the `spec` was limited to part of the batch, you could check how many times `read(spec)` was called.

Notice how these 2 parts of code from `AR::EventRepository` and `Legacy::EventRepository` are similar and almost duplicated:

```
      if spec.batched?
        batch_reader = ->(offset, limit) { stream.offset(offset).limit(limit).map(&method(:build_event_instance)) }
        RubyEventStore::BatchEnumerator.new(spec.batch_size, total_limit(spec), batch_reader).each
      else
        stream.map(&method(:build_event_instance)).each
      end
```

```
        if spec.batched?
          batch_reader = ->(offset, limit) { stream.offset(offset).limit(limit).map(&method(:build_event_entity)) }
          RubyEventStore::BatchEnumerator.new(spec.batch_size, total_limit(spec), batch_reader).each
        else
          stream.map(&method(:build_event_entity)).each
        end
```

that suggests to me that batching is a concept which most likely can be generalized and used above repository. The logic for every repo would be the same imho. Read N events from position X then read N events from position N+X etc ... I believe this logic could be moved to `Specification#each` and/or `Specification#each_batch`.

Perhaps this comments is expressed at the wrong issue/pull request. But that's how I imagined batching internally and I wanted to comment this way of checking how many times `read` was called on Repo could be used for spying if AggregateRoot uses batching. But I realized right now batching is an internal detail used in `read` so it's not the design I imagined.

I am not sure the described solution would be definitely better but these are my current doubts.

Feel free to ignore and merge. Looks good.

pawelpacana:
Thanks for insights. In fact calling read mutliple times was interim implementation used in `RES::Projection`. 

> If RubyEventStore::BatchEnumerator was executed above the repository and called repository.read(spec) where the spec was limited to part of the batch, you could check how many times read(spec) was called.

While I agree on usefulness of this implementation in terms of writing test assertion, I have a feeling it's getting a little bit to intimate with implementation details.

Issue #346, opened 2018-05-05
Correlation ID as 1st class citizen


paneq:
Imagine if dispatcher wrapped calling event handlers in

```ruby
def call(subscriber, event)
  res.with_metadata(correlation_id: event.metadata[:correlation_id] || event.event_id) do
    subscriber = subscriber.new if Class === subscriber
    subscriber.call(event)
  end
end
```

That would immediately help in debugging and propagating correlation between initial event and events triggered by it. If the initial event gets correlation_id copied from command then we also know which command/request triggered the whole _tree_ of events.

Question: How to handle async event handlers.

Related: 

* "Experiment with metadata-driven streams #221"
* "with_metadata" #327 
* Storing Commands in RES #340

paneq:
```
The are both really simple patterns I have never quite understood why 
they end up so misunderstood. 

Let's say every message has 3 ids. 1 is its id. Another is correlation 
the last it causation. 

The rules are quite simple. If you are responding to a message, you 
copy its correlation id as your correlation id, its message id is your 
causation id. 

This allows you to see an entire conversation (correlation id) or to 
see what causes what (causation id). 
```

https://groups.google.com/forum/#!topic/event-store/pvk2iEBgBtA

paneq:
Just as ruby event store automatically fills out the timestimep, it should also automatically fill out `causation_id` and `correlation_id` from event.id if they are blank/not provided. That way the 1st event in chain is going to have those 3 ids filled out instead of having ID only.

paneq:
`event.correlate_with(another_event_or_command)` could be a nice low-level API to copy those values to metadata.

joelvh:
Should correlation ID and causation ID be individual columns on SQL store? If first class citizens, this would at least make storage and retrieval more efficient. I think of metadata as storage of data for when an individual event is read, whereas I can see querying on the other IDs to get a subset of events.

paneq:
@joelvh my plan is to link them to streams https://blog.arkency.com/correlation-id-and-causation-id-in-evented-systems/

joelvh:
Fair enough. I think having some of this behavior as options you can toggle on or off would be ideal, such as automatically linking causation and correlation streams.

paneq:
@joelvh yep, exactly. I was thinking about an event handler that you can subscribe manually.

paneq:
Please send feedback for https://github.com/RailsEventStore/rails_event_store/pull/362/files

[...]

PR #362, opened 2018-05-22
Event#correlate_with

paneq:
https://blog.arkency.com/correlation-id-and-causation-id-in-evented-systems/

> Let’s say every message has 3 ids. 1 is its id. Another is correlation the last
> it causation. If you are responding to a message, you copy its correlation id
> as your correlation id, its message id is your causation id. This allows you to
> see an entire conversation (correlation id) or to see what causes what (causation id).

Issue: #346

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/event.rb
+++ b/ruby_event_store/lib/ruby_event_store/event.rb
@@ -56,7 +56,7 @@ module RubyEventStore
     end
 
     def causation_id=(val)
-      metadata[:causation_id]=(val)
+      metadata[:causation_id]= val
     end
 
     def correlate_with(other_message)
--- a/ruby_event_store/lib/ruby_event_store/event.rb
+++ b/ruby_event_store/lib/ruby_event_store/event.rb
@@ -6,8 +6,6 @@ module RubyEventStore
       @event_id = event_id.to_s
       @metadata = Metadata.new(metadata.to_h)
       @data     = data.to_h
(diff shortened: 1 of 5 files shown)


ruby_event_store/lib/ruby_event_store/event.rb
```diff
@@ -6,6 +6,8 @@ def initialize(event_id: SecureRandom.uuid, metadata: nil, data: nil)
       @event_id = event_id.to_s
       @metadata = Metadata.new(metadata.to_h)
       @data     = data.to_h
+      self.correlation_id ||= @event_id
```
pkondzior: Is this for subclassing of Event and having set corelation_id and causation_id before super() call ?

pkondzior: Makes sens, I was thinking that this implies order of using super() in subclass initializer, this probably doesn't matter that much.

andrzejkrzywda: `I want every message to always have these 3 IDs so I never need to think that they could be nil and workaround for that on any higher layer.` I have no idea what's the proper solution yet, but it sounds like there can be something which avoids this problem.

It doesn't make sense that we set the causation_id or correlation_id to the event itself.

paneq: > It doesn't make sense that we set the causation_id or correlation_id to the event itself.

Thanks! Obviously it made sense to at least someone (me)...

Issue #308, opened 2018-04-21
Cannot serialize event's data and metadata in binary format due to DB schema


paneq:
While working on #307 I encountered an unexpected problem. If event's data or metadata is serialized in a binary format which has bytes outside of utf8/latin encoding currently generated by our migration then storing it fails. Verified on Sqlite3 and Mysql.

Test for `event_repository_lint`:

```ruby
  specify 'can store arbitrary binary data' do
    binary = "\xB0"
    expect(binary.valid_encoding?).to eq(false)
    binary.force_encoding("binary")
    expect(binary.valid_encoding?).to eq(true)

    repository.append_to_stream(
      event = SRecord.new(data: binary, metadata: binary),
      RubyEventStore::Stream.new('stream'),
      RubyEventStore::ExpectedVersion.none
    )
  end
```

Exception sqlite:

```
     Encoding::UndefinedConversionError:
       "\xB0" from ASCII-8BIT to UTF-8
     # /Users/rupert/develop/rails_event_store/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb:14:in `block in append_to_stream'
```

Exception mysql:

```
  1) RailsEventStoreActiveRecord::EventRepository behaves like event_repository can store arbitrary binary data
     Failure/Error: build_event_record(event).save!
     
     ActiveRecord::StatementInvalid:
       Mysql2::Error: Incorrect string value: '\xB0' for column 'metadata' at row 1: INSERT INTO `event_store_events` (`event_type`, `id`, `metadata`, `data`, `created_at`) VALUES ('4052b236-36b3-44d6-b1a8-639acb846bb8', 'a9b55df4-34ec-47e9-94fa-c0936e27704b', '?', '?', '2018-04-21 08:34:04')
     Shared Example Group: :event_repository called from ./spec/event_repository_spec.rb:26
     # ./lib/rails_event_store_active_record/event_repository.rb:14:in `block in append_to_stream'
```

Related:

* Compression mapper: #295
* encryption mapper / GDPR
* #189

paneq:
Our migration uses `text` fields but `binary` would be better for storing anything. I am wondering whether to make it mandatory or optional.

After all a developer might always choose to edit a migration or not run it so I don't see an issue if someone wants to store with `text` or `json` ...

@pawelpacana @joelvh @andrzejsliwa @valscion @mlomnicki what do you think?

paneq:
[...]

PR #363, opened 2018-05-25
Serialize events during dispatch through mapper

pkondzior:
This pull request changes the way how we are serializing/deserializing events for asynchronous processing. It's main purpose is to reuse of mappers for serialization and deserialization, this has been split into two smaller objectives:

* Propagate serialized events with domain events directly to dispatcher
* Provide generic method for deserializing events back to domain events for async handlers

How to use new API with default mapper:

RES clients Default mapper is going to use YAML serialization for each of payload arguments, you have to use `RES::Client#load_serialized_event` in order to instantiate domain event:

```Ruby
EventStore = RailsEventStore::Client.new

class AsyncHandler < ActiveJob::Base
  def perform(payload)
    event = EventStore.load_serialized_event(payload)
  end
end

EventStore.subscribe_to_all_events(AsyncHandler)
EventStore.publish_event(RailsEventStore::Event.new)
```

If you would like to change mapper to Protobuf you should configure Your RES client with Protobuf mapper instead of default one, just remember that Protobuf requires different type of column for storing data and metadata see https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store/spec/protobuf_integration_spec.rb#L42 and https://github.com/RailsEventStore/rails_event_store/issues/308: 

```Ruby
mapper = RubyEventStore::Mappers::Protobuf.new
EventStore = RailsEventStore::Client.new(mapper: mapper)

class AsyncHandler < ActiveJob::Base
  def perform(payload)
    event = EventStore.load_serialized_event(payload)
  end
end

EventStore.subscribe_to_all_events(AsyncHandler)

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_message "TestEvent" do
    optional :foo, :string, 1
  end
end

TestEventData = Google::Protobuf::DescriptorPool.generated_pool.lookup("TestEvent").msgclass
EventStore.publish_event(RubyEventStore::Proto.new(data: TestEventData.new(foo: '123')))
```

Either Default or Protobuf mappers allows you to dynamically map event classes:

```Ruby
mapper = RubyEventStore::Mappers::Default.new(events_class_remapping: {'TestEvent' => 'RefactoredTestEvent'})
EventStore = RailsEventStore::Client.new(mapper: mapper)

class AsyncHandler < ActiveJob::Base
  def perform(payload)
    event = EventStore.load_serialized_event(payload) # event.class == RefactoredTestEvent
  end
end

EventStore.subscribe_to_all_events(AsyncHandler)
class TestEvent < RailsEventStore::Event
end
class RefactoredTestEvent < RailsEventStore::Event
end
EventStore.publish_event(TestEvent.new)
```

Public API changes:

* Change in signature of `RubyEventStore::PubSub::Dispatcher#call` and `RailsEventStore::ActiveJobDispatcher#call` method from `(subscriber, event)` to `(subscriber, payload)` where payload becomes `[event, serialized_event]`
* New method `RubyEventStore::PubSub::Dispatcher#load_serialized_event(event_type:, event_id:, data:, metadata:)`
* Change in signature of `RailsEventStore ::AsyncProxyStrategy:: AfterCommit#call` from `(klass, event)` to `(klass, serialized_event)`
* Change in signature of `RailsEventStore ::AsyncProxyStrategy:: Inline#call` from `(klass, event)` to `call(klass, serialized_event)`
* changes in signature of `RubyEventStore:: PubSub::Broker#notify_subscribers` from `(event)` to `(payload)` where payload becomes `[event, serialized_event]`

### DIFF
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -51,8 +51,9 @@ module RailsEventStore
       @async_proxy_strategy = proxy_strategy
     end
 
-    def call(subscriber, event, serialized_event)
+    def call(subscriber, payload)
       if async_handler?(subscriber)
+        _, serialized_event = *payload
         @async_proxy_strategy.call(subscriber, serialized_event)
       else
         super
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -51,7 +51,7 @@ module RailsEventStore
       @async_proxy_strategy = proxy_strategy
     end
 
-    def call(subscriber, event, serialized_event)
+    def call(subscriber, _, serialized_event)
       if async_handler?(subscriber)
         @async_proxy_strategy.call(subscriber, serialized_event)
       else
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -3,26 +3,26 @@ require 'active_job'
 module RailsEventStore
   module AsyncProxyStrategy
     class AfterCommit
-      def call(klass, event)
+      def call(klass, serialized_event)
         if ActiveRecord::Base.connection.transaction_open?
           ActiveRecord::Base.
             connection.
             current_transaction.
-            add_record(AsyncRecord.new(klass, event))
+            add_record(AsyncRecord.new(klass, serialized_event))
         else
-          klass.perform_later(YAML.dump(event))
+          klass.perform_later(serialized_event)
         end
       end
 
       private
       class AsyncRecord
-        def initialize(klass, event)
+        def initialize(klass, serialized_event)
           @klass = klass
-          @event = event
+          @serialized_event = serialized_event
         end
 
         def committed!
-          @klass.perform_later(YAML.dump(@event))
+          klass.perform_later(serialized_event)
         end
 
         def rolledback!(*)
@@ -32,14 +32,16 @@ module RailsEventStore
         end
 
         def add_to_transaction
-          AfterCommit.new.call(@klass, @event)
+          AfterCommit.new.call(klass, serialized_event)
         end
+
+        attr_reader :serialized_event, :klass
       end
     end
 
     class Inline
-      def call(klass, event)
-        klass.perform_later(YAML.dump(event))
+      def call(klass, serialized_event)
+        klass.perform_later(serialized_event)
       end
     end
   end
@@ -49,9 +51,9 @@ module RailsEventStore
       @async_proxy_strategy = proxy_strategy
     end
 
-    def call(subscriber, event)
+    def call(subscriber, event, serialized_event)
       if async_handler?(subscriber)
-        @async_proxy_strategy.call(subscriber, event)
+        @async_proxy_strategy.call(subscriber, serialized_event)
       else
         super
       end
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -51,9 +51,8 @@ module RailsEventStore
       @async_proxy_strategy = proxy_strategy
     end
 
-    def call(subscriber, payload)
+    def call(subscriber, event, serialized_event)
       if async_handler?(subscriber)
-        _, serialized_event = *payload
         @async_proxy_strategy.call(subscriber, serialized_event)
       else
         super
--- a/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
@@ -10,7 +10,7 @@ module RailsEventStore
             current_transaction.
             add_record(AsyncRecord.new(klass, serialized_event))
         else
-          klass.perform_later(serialized_event)
+          klass.perform_later(serialized_event.to_h)
         end
       end
 
@@ -22,7 +22,7 @@ module RailsEventStore
         end
 
         def committed!
-          klass.perform_later(serialized_event)
+          klass.perform_later(serialized_event.to_h)
         end
 
         def rolledback!(*)
@@ -41,7 +41,7 @@ module RailsEventStore
 
     class Inline
       def call(klass, serialized_event)
-        klass.perform_later(serialized_event)
+        klass.perform_later(serialized_event.to_h)
       end
     end
   end
--- a/rails_event_store/spec/active_job_dispatcher_spec.rb
+++ b/rails_event_store/spec/active_job_dispatcher_spec.rb
@@ -26,7 +26,6 @@ module RailsEventStore
 
     let!(:event) { RailsEventStore::Event.new(event_id: "83c3187f-84f6-4da7-8206-73af5aca7cc8") }
     let!(:serialized_event)  { RubyEventStore::Mappers::Default.new.event_to_serialized_record(event) }
-    let!(:yaml) { '123' }
 
     it "verification" do
       expect do
--- a/rails_event_store/spec/active_job_dispatcher_spec.rb
+++ b/rails_event_store/spec/active_job_dispatcher_spec.rb
@@ -24,8 +24,9 @@ module RailsEventStore
       AsyncHandler.reset
     end
 
-    let!(:event) { RailsEventStore::Event.new }
-    let!(:yaml)  { YAML.dump(event) }
+    let!(:event) { RailsEventStore::Event.new(event_id: "83c3187f-84f6-4da7-8206-73af5aca7cc8") }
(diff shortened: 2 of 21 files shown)


rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
```diff
@@ -32,14 +32,16 @@ def before_committed!
         end
 
         def add_to_transaction
-          AfterCommit.new.call(@klass, @event)
+          AfterCommit.new.call(klass, serialized_event)
         end
+
+        attr_reader :serialized_event, :klass
```
paneq: 👍

rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
```diff
@@ -49,9 +51,10 @@ def initialize(proxy_strategy: AsyncProxyStrategy::Inline.new)
       @async_proxy_strategy = proxy_strategy
     end
 
-    def call(subscriber, event)
+    def call(subscriber, payload)
       if async_handler?(subscriber)
-        @async_proxy_strategy.call(subscriber, event)
+        _, serialized_event = *payload
```
paneq: no idea what's going on here? are we passing an array as `payload` here?

paneq: if so, why not go with

```
call(subscriber, event, serialized_event)
```

?

rails_event_store/spec/active_job_dispatcher_spec.rb
```diff
@@ -41,42 +41,60 @@ module RailsEventStore
...
       expect(AsyncHandler.received).to be_nil
       perform_enqueued_jobs(AsyncHandler.queue_adapter)
-      expect(AsyncHandler.received).to eq(yaml)
+      expect(AsyncHandler.received).to eq({
+        "event_id"        => "83c3187f-84f6-4da7-8206-73af5aca7cc8",
+        "data"            => "--- {}\n",
+        "metadata"        => "--- {}\n",
+        "event_type"      => "RubyEventStore::Event",
+        "_aj_symbol_keys" => ["event_id", "data", "metadata", "event_type"]
+      })
```
paneq: This assertion does not seem valuable to me. I would prefer the handler to deserialize the event and we should compare that what we received is equal to `event` and has the same `metadata` (checking metadata is not part of == definition for events)

rails_event_store/spec/active_job_dispatcher_spec.rb
```diff
@@ -41,42 +41,60 @@ module RailsEventStore
...
     it "async proxy for defined adapter enqueue job immediately when no transaction is open" do
       dispatcher = ActiveJobDispatcher.new(proxy_strategy: AsyncProxyStrategy::AfterCommit.new)
       expect_to_have_enqueued_job(AsyncHandler) do
-        dispatcher.call(AsyncHandler, event)
+        dispatcher.call(AsyncHandler, [event, serialized_event])
       end
       expect(AsyncHandler.received).to be_nil
       perform_enqueued_jobs(AsyncHandler.queue_adapter)
-      expect(AsyncHandler.received).to eq(yaml)
+      expect(AsyncHandler.received).to eq({
```
paneq: ditto

rails_event_store/spec/protobuf_integration_spec.rb
```diff
@@ -4,8 +4,10 @@
 class AsyncProtoHandler < ActiveJob::Base
   self.queue_adapter = :inline
 
-  def perform(event)
-    @@event = YAML.load(event)
+  cattr_accessor :event_store
```
paneq: not ideal but I don't have better idea right now.

I prefer our tests to use similar constructs as devs would use for their application: `Rails.configuration.event_store` but that's not easy to achieve here.

paneq:
`load_serialized_event` - how about just `deserialize` ?

paneq:
@pkondzior because docs changes are deployed immediately it's (currently) based to have a separate PR for them which is merged only after new gem release.

rails_event_store/spec/protobuf_integration_spec.rb
pkondzior: Some global scope is necessary here, maybe just `EventStore` as it is kind of similar to `Rails`

railseventstore.org/source/docs/protobuf.html.md.erb
```diff
@@ -108,8 +108,8 @@ event_store.subscribe(->(ev){ },  to: [MyApp::OrderPlaced.descriptor.name])
 class SendOrderEmailHandler < ActiveJob::Base
   self.queue_adapter = :inline
 
-  def perform(event)
-    event = YAML.load(event)
+  def perform(payload)
+    event = Rails.configuration.event_store.load_serialized_event(payload)
```
paneq: I like to extract `Rails.configuration.event_store` to private method. Even in the documentation :)

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -16,21 +16,22 @@ def initialize(repository:,
     end
 
     def publish_events(events, stream_name: GLOBAL_STREAM, expected_version: :any)
-      append_to_stream(events, stream_name: stream_name, expected_version: expected_version)
-      events.each do |ev|
-        event_broker.notify_subscribers(ev)
+      prepared_events = prepare_events(events)
+      serialized_events = serialize_events(prepared_events)
+      append_to_stream_serialized_events(serialized_events, stream_name: stream_name, expected_version: expected_version)
+      prepared_events.zip(serialized_events) do |payload|
```
paneq: +1 for using `zip`

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -16,21 +16,22 @@ def initialize(repository:,
...
+        event_broker.notify_subscribers(payload)
```
paneq: But I believe it would be better to call with 2 arguments directly

```
prepared_events.zip(serialized_events) do |(event, serialized_event)|
  event_broker.notify_subscribers(event, serialized_event)
end
```

`payload` does not really mean much to me here...

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -16,21 +16,22 @@ def initialize(repository:,
...
       end
       :ok
     end
 
     def publish_event(event, stream_name: GLOBAL_STREAM, expected_version: :any)
-      publish_events([event], stream_name: stream_name, expected_version: expected_version)
+      publish_events(event, stream_name: stream_name, expected_version: expected_version)
```
paneq: seems weird to me we are not passing `[event]` . but not worth much discussion considering #366

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -216,6 +221,16 @@ def enrich_event_metadata(event)
       event.metadata[:timestamp] ||= clock.call
     end
 
+    def append_to_stream_serialized_events(serialized_events, stream_name:, expected_version:)
+      repository.append_to_stream(serialized_events, Stream.new(stream_name), ExpectedVersion.new(expected_version))
+    end
+
+    def prepare_events(events)
```
paneq: This method name is suboptimal `enrich_events_metadata` or `enrich_events` would be better, I believe.

rails_event_store/lib/rails_event_store/active_job_dispatcher.rb
pkondzior: Sync handlers need event, Async handlers need serialized_event,I though it might be better idea to just keep them together as one argument and  unpack them depending on the dispatcher implementation instead of passing them always as separate arguments and then experiencing mutant brag about one or another being not used. 

But I'm not really opinionated here, we can go with explicit arguments and figure out how to deal with mutant either by providing right subject spec (optimal) or ignoring Dispatcher#call at all (suboptimal)

rails_event_store/spec/active_job_dispatcher_spec.rb
pkondzior: Do we really care about event deserialization at this point ? Isn't this more about right stuff being received through ActiveJob layer ?

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -189,9 +190,13 @@ def with_metadata(metadata, &block)
       self.metadata = previous_metadata
     end
 
+    def load_serialized_event(event_type:, event_id:, data:, metadata:)
```
paneq: somehow I expected a single string here, but we've discussed it a bit :)

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -189,9 +190,13 @@ def with_metadata(metadata, &block)
...
+      mapper.load_serialized_event(event_type: event_type, event_id: event_id, data: data, metadata: metadata)
```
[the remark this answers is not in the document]
pkondzior: Because I think it's faster to go directly from serialized attributes to event. There is nothing special in SerializedRecord here that we need, it would be extra allocation, this is how ActiveRecord become useless... extra allocations everywhere.

PR #365, opened 2018-05-28
ROM Memory adapter implementation to allow easily adding alternate ROM implementations

joelvh:
This implements the ROM Memory adapter as an alternative to the ROM SQL adapter. Tests are structured to test ROM implementations easily. A new ROM adapter simply needs to implement the two relations required.


joelvh:
@paneq there are 7 mutations locally that need to be killed. Mutant says that it re-inserted the original code and tests failed. However, there is no failure output.

Any idea what to do about that?

joelvh:
@paneq or @pawelpacana you guys know how to make sense of the results here? https://circleci.com/gh/RailsEventStore/rails_event_store/2492

Thanks!

PR #374, opened 2018-06-15
Correlation between events in sync handlers

paneq:
#346 & #347

If event triggers directly another event.


paneq:
@jakubkosinski @pawelpacana @fidel please have a look

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -206,17 +208,13 @@ def normalize_to_array(events)
     end
 
     def enrich_event_metadata(event)
-      if metadata_proc
-        md = metadata_proc.call || {}
-        md.each{|k,v| event.metadata[k]=(v) }
-      end
       if metadata
-        metadata.each { |key, value| event.metadata[key] = value }
+        metadata.each { |key, value| event.metadata[key] ||= value }
```
paneq: @jakubkosinski what do you think?

PR #381, opened 2018-06-26
Allow to use any background job library to handle async jobs

mpraglowski:
Breaking change: signature of dispatcher's proxy strategy is changed

### DIFF
--- a/rails_event_store/lib/rails_event_store/async_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/async_dispatcher.rb
@@ -1,28 +1,27 @@
-require 'active_job'
-
 module RailsEventStore
   module AsyncProxyStrategy
     class AfterCommit
-      def call(klass, serialized_event)
+      def call(klass, serialized_event, async_call)
         if ActiveRecord::Base.connection.transaction_open?
           ActiveRecord::Base.
             connection.
             current_transaction.
-            add_record(AsyncRecord.new(klass, serialized_event))
+            add_record(AsyncRecord.new(klass, serialized_event, async_call))
         else
-          klass.perform_later(serialized_event.to_h)
+          async_call.call(klass, serialized_event)
         end
       end
 
       private
       class AsyncRecord
-        def initialize(klass, serialized_event)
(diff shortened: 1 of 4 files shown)


rails_event_store/spec/async_dispatcher_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+      def call(event)
+        @@received = event
+      end
+    end
+    class NotCallableHandler
+    end
+
+    class MyAsyncHandler
+      @@received = nil
```
pawelpacana: I'm afraid this and other class variables will make this test brittle

rails_event_store/lib/rails_event_store/async_dispatcher.rb
```diff
@@ -1,27 +1,25 @@
 module RailsEventStore
```
pawelpacana: `ActiveJobDispatcher` is `RailsEventStore::`, however `AsyncDispatcher` can be moved to `RubyEventStore::` and be used in non-rails scenarios.

mpraglowski: We still use there `AsyncProxyStrategy::AfterCommit` which uses `ActiveRecord` but maybe this should stay in `RailsEventStore::` and only `AsyncProxyStrategy::Inline` should moved with `AsyncDispatcher` to `RubyEventStore::`

pawelpacana: :+1:

paneq:
Looks good (and I am happy it was merged) except for one thing. `async_call` is a suboptimal name. Let's figure out something better? 

How about 

* `scheduler`, 
*`queuing`, 
* `enqueuing`, 
* `enqueue`,
* `enqueuer`, 
* `adapter`

?

paneq:
@mpraglowski do you plan updating the documentation in a separate PR?

pawelpacana:
> async_call is a suboptimal name

`scheduler` and `executor` were on my list

pawelpacana:
btw. docs can live in master via toggle

Issue #221, opened 2018-03-13
Experiment with metadata-driven streams


pawelpacana:
One could image forming streams based on some metadata values.

For example if we record in metadata:
- requester IP address
- logged-in user
- `correlation_id`, `request_id` or any other identifier of single activity

we could then partition streams by this metadata:
- everything that happened manifested by domain events with the same `correlation_id`
- all events originating from single IP address
- all events of particular user revealing flows of events in the system

Might be related to: https://github.com/RailsEventStore/rails_event_store/issues/72#issuecomment-326513289

paneq:
This might require:

* exposing metadata API that can be used for events inheriting from `RubyEventStore::Event` and protobuf defined events
  * exposing in repository
  * exposing in mappers
* writing a global handler/subscriber (configurable?) that would link to stream based on event metadata (which one? all?). What convention should it use for stream names? Maybe `$#{metadata_key}-#{metadata_value}` ?

pawelpacana:
https://github.com/EventStore/EventStore/pull/1622

pawelpacana:
👍

PR #382, opened 2018-06-27
LinkByMetadata, LinkByCorrelationId, LinkByCausationId introduced

paneq:
Issue: #221 #346


ruby_event_store/spec/link_by_metadata_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        mapper: RubyEventStore::Mappers::Protobuf.new,
+        repository: InMemoryRepository.new
+      )
+      event_store.subscribe_to_all_events(LinkByMetadata.new(event_store: event_store, key: :city))
+      ev = RubyEventStore::Proto.new(
+        data: ResTesting::OrderCreated.new(
+          customer_id: 123,
+          order_id: "K3THNX9",
+        ),
+        metadata: { city: "Chicago" }
```
mpraglowski: I would prefer some mote "technical" example here to not encourage users to put business data into metadata.

ruby_event_store/spec/link_by_metadata_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      )
+      event_store.publish(ev)
+
+      expect(event_store.read.stream("$by_city_Chicago").each.to_a).to eq([ev])
```
mpraglowski: Case sensitive?

rails_event_store/lib/rails_event_store/all.rb
```diff
@@ -20,6 +21,7 @@ module RailsEventStore
   InvalidHandler              = RubyEventStore::InvalidHandler
   InvalidPageStart            = RubyEventStore::InvalidPageStart
   InvalidPageSize             = RubyEventStore::InvalidPageSize
+  CorrelatedCommands          = RubyEventStore::CorrelatedCommands
```
mpraglowski: Why here? Seems not to be correlated with that PR :)

mpraglowski:
Documentation missing. Maybe use version management to add it within this PR ?

ruby_event_store/spec/link_by_metadata_spec.rb
paneq: I see your point. However I considered this to be a technical metadata, the location of an user responsible for this event :) So you can have a look what are the people from _chicago_ doing to your system. Does that make sense?

ruby_event_store/spec/link_by_metadata_spec.rb
paneq: Good point. Yes. If that's not good enough for someone, they can write these 3 lines as a separate handler that transforms the data.

ruby_event_store/lib/ruby_event_store/link_by_metadata.rb
```diff
@@ -0,0 +1,44 @@
...
+  class LinkByMetadata
+
+    def initialize(
+      event_store:,
+      key:,
+      prefix: nil
+    )
+      @event_store = event_store
+      @key = key
+      @prefix = prefix || ["$by", key, nil].join("_")
```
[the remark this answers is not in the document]
paneq: It was `""` but mutants says it works with nil as well :)

ruby_event_store/spec/link_by_metadata_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      )
+
+      event_store.publish(ev = OrderCreated.new(metadata:{
+        city: "Paris",
+      }))
+
+      expect(event_store.read.stream("sweet+Paris").each.to_a).to eq([ev])
+    end
+
+    specify "array of ids" do
```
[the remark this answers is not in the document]
paneq: The point is that link_to_stream receives an array of ids instead of a single id (which also works but I believe accidentally rather than explicitly)

ruby_event_store/spec/link_by_metadata_spec.rb
mpraglowski: Chicago looks like "business" description of location. IP address would be perfectly fine here :)

paneq:
@mpraglowski documentation added in a separate PR. It already contains metadata changes so I just increased the scope.

ruby_event_store/lib/ruby_event_store/link_by_metadata.rb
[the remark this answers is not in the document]
mpraglowski: 👍

rails_event_store/lib/rails_event_store/all.rb
paneq: I forgot to do it in the other PR :) it's slightly related via being about _correlation_ :)

ruby_event_store/spec/link_by_metadata_spec.rb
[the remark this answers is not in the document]
mpraglowski: Yeah I know, but it took me a while to understand that test.

PR #385, opened 2018-06-28
Missing step needed to run locally raileventstore.org docs

mpraglowski:


railseventstore.org/README.md
```diff
@@ -11,6 +11,7 @@ Provided you have Ruby installed, you can serve the docs locally at `http://loca
 
 ```
 bundle install
+bin/build
```
pawelpacana: actually it is `./bin/contributors` as build script further produces static html while we prefer the middleman server in development

PR #378, opened 2018-06-20
Use new publish API in docs [#366]

pkondzior:
Should be deployed when #366 is released

[ci skip]

PR #387, opened 2018-06-29
append_to_stream rename

mpraglowski:


railseventstore.org/source/docs/exceptions.html.md.erb
```diff
@@ -34,13 +34,21 @@ If you want to have an event present in multiple streams, you have to link it wi
 
 ### RubyEventStore::InvalidExpectedVersion    
 
+<% if version_above('0.30.0') %>
+Occurs when invalid `exception_version` is passed in `append`, `link_to_stream`, `publish_event` or `publish_events`.
```
joelvh: Was `publish` rename merged?

mpraglowski: https://github.com/RailsEventStore/rails_event_store/pull/377 yes - but I don't want to change docs in this PR - keep it focused on append_to_stream - however this needs update.

mpraglowski: The changes in docs for #377 are here https://github.com/RailsEventStore/rails_event_store/pull/378

paneq:
@mpraglowski what do you think about `persist` instead of `append` ? Imagine we start persisting commands as well without triggering handlers.

mpraglowski:
Some said:

> I don't assume commands must to be stored in the same stream as as events for an object/entity/aggregate.

So why you assume we will use the same method to store events & commands ;)

paneq:
> So why you assume we will use the same method to store events & commands ;)

I see no reason to have 2 different methods for it. For me it's about persisting a message in both cases.

pawelpacana:
@mpraglowski would merge but conflicts

PR #388, opened 2018-06-30
link_to_steam rename

mpraglowski:


mpraglowski:
Rebased with master because conflict after YARN docs added.

PR #389, opened 2018-06-30
Dispatcher is not an event broker dependency

mpraglowski:


ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -1,15 +1,19 @@
 require 'concurrent'
 
 module RubyEventStore
+  DEFAULT_DISPATCHER = PubSub::Dispatcher.new
```
paneq: this is constant :( please don't

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -1,15 +1,19 @@
...
+
   class Client
     def initialize(repository:,
                    mapper: Mappers::Default.new,
-                   event_broker:  PubSub::Broker.new,
+                   event_broker: PubSub::Broker.new,
```
paneq: this is always a new, separate instance, initialized for every new client. no shared dependencies.

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -174,31 +181,38 @@ def call
...
+      def verify_subscriber(subscriber)
+        raise SubscriberNotExist if subscriber.nil?
+        @dispatcher.verify(subscriber)
+      end
     end
 
     def within(&block)
       raise ArgumentError if block.nil?
-      Within.new(block, event_broker)
+      Within.new(block, event_broker, dispatcher)
```
paneq: step in a good direction, i wonder if we could not pass `self`.

rails_event_store/lib/rails_event_store/client.rb
```diff
@@ -4,12 +4,14 @@ class Client < RubyEventStore::Client
 
     def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
                    mapper: RubyEventStore::Mappers::Default.new,
-                   event_broker: EventBroker.new(dispatcher: ActiveJobDispatcher.new),
+                   event_broker: EventBroker.new,
```
paneq: is Broker still actually a broker if it does not handle dispatching the events?

ruby_event_store/lib/ruby_event_store/client.rb
paneq: probably not.

paneq:
I know it was my idea, but I don't think I am particularly happy with the end result.

I was not happy with broker taking dispatcher as a dependency. I am not so happy with client orchestrating the broker&dispatcher flows either. I think we are missing a 3rd option.

I think we need a new class that takes Dispatcher and Broker as dependencies. And since broker no longer deals with dispatching, I believe it should be named differently. It's more like a... list of subscribers, list of handlers, ... handlers register, observers ? This class basically implements pub/sub. I think the API would be that of current Broker.

Hmmm, I think that basically means. Instead of doing all that we did here. Let's extract those 4 methods from Broker as a separate class. And broker (aka pub-sub) just wires these 2 dependencies together.

```ruby
class Client
  def initialize(dispatcher:, register:, ...)
    @broker = EventBroker.new(dispatcher, register)
  end
```

* `dispatcher` - knows how to call sync/async/whatever handlers that subscribed/registered for certain events
* `register` - keeps a querable list of who subscribed for what.
* `broker` - a faced for both. A full pub/sub implementation. Might even be called `PubSub` :)

ruby_event_store/lib/ruby_event_store/client.rb
mpraglowski: Nope. We do not have thread subscribers in client, we need a broker here (and dispatcher too). We might have get this from client but they are protected and exposing this outside client is not good.

mpraglowski:
It makes sense. We will have 3 classes. 

Dispatcher - responsible for dispatching events to subscribers
Subscriber - responsible for managing subscriptions
Broker - "just" a pub sub - this will have both 2 previous ones given as dependency and will do the notifications - the code from client to notify subscribers should go there

paneq:
@mpraglowski exactly.

I am not sure about the names though:

Dispatcher - mkay
Subscriber - is it really a subscriber? We call a subscriber the thing that subscribes (handlers), not the thing that keeps a list of subscriptions.
Broker - mkay

I was even thinking about

* pub
* sub
* PubSub.new(pub, sub)

😂 

Feel free to come up with anything you feel good about.

pawelpacana:
no deprecations for this change?

PR #393, opened 2018-07-06
Update YARD docs for append after rename

mpraglowski:


ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -58,10 +58,7 @@ def publish_event(event, stream_name: GLOBAL_STREAM, expected_version: :any)
       publish(event, stream_name: stream_name, expected_version: expected_version)
     end
 
-    # Persists new event(s) without notifying any subscribed handlers
-    #
-    # @param (see #publish)
-    # @return [:ok]
+    # @deprecated Use {#publish} instead
```
pawelpacana: publish for append_to_stream?

Issue #106, opened 2017-09-09
Event Store as queue


paneq:
Implement Event Store in a way that would allow clients to safely iterate over global stream of all events with pagination and no race conditions. The client would remember its position in the stream of events in its own side.

paneq:
By safely I mean in a way which allows to avoid this potential problem when reading events:

![race_condition_query](https://user-images.githubusercontent.com/65587/30241715-2f09723e-9589-11e7-9f41-9868f6e794e1.png)

In such case we can get IDs of events 0 and 2, but skip reading uncommitted event nr 1.

We would like to avoid such problems.

From my reading about it and thinking over it, I see 2 potential solutions:

* linearize writing to EventStore by using exclusive lock on underlying DB. This can however limit write throughput which I would like to avoid.
* The other solution I thought about is different. It would delay the problem until read.
  * The problem is that assuming auto-incremented IDs of events in global stream there can be a gap in read data. We don't easily know if that data was not yet committed or rolled back.
  * what if we used `after_commit` to marked rolled-backed positions as taken so the reading process would know to skip them
  * but `after_commit` can fail when server is turned off. What if we had a constantly running external process that would try to fill the gaps with missing numbers. If the transaction was rolled-back it would simply fill the missing number. If the transaction was long and still running it would cause a deadlock (ideally in a way that it would loose and keep trying until the competing transaction is committed or rolled-back)
  * that supervisor process theoretically should be needed very rarely. Only when there was rollback and server was turned off so the usual after_commit callbacks did not have enough time to finish.

In case of a very long transaction with events, the write throughput would not be affected but a separate queue reading process would not be able to proceed until the transaction is finished. As very long transactions don't happen often it might not be a problem every in an app.

paneq:
Imagine starting point like this:

```
require 'active_record'

ENV['DATABASE_URL'] ||= "postgres://localhost/rails_event_store_active_record?pool=5"

ActiveRecord::Base.establish_connection(ENV['DATABASE_URL'])
ActiveRecord::Schema.define do
  self.verbose = true

  enable_extension "plpgsql"
  enable_extension "pgcrypto"

  create_table(:event_store_events, id: :uuid, force: true) do |t|
    t.string      :event_type,  null: false
    t.text        :metadata
    t.text        :data,        null: false
    t.datetime    :created_at,  null: false
  end
  add_index :event_store_events, :created_at

  create_table(:event_store_global_stream, force: true) do |t|
[...]

PR #403, opened 2018-07-22
PgLinearizedEventRepository

paneq:
repository which uses lock (released at the end of a transaction) to
guarantee only one batch of events written concurrently.

* works only with postgres
  * mysql advisory locks are not easily releasible at the end of a
    transaction

* global/all stream still does not have "position" ordered but we are
  planning to drop it anyway.

* not mutated because mutations would fail on non-pg connections

* reading query result just in case with `each{}`. Perhaps not
  necessary.

Workaround for #106 which should be suitable on greenfield projects
which only use short transactions.

Not recommended to use when you publish from within long transactions as
it can block other concurrent operations.

Throughput and performance not benchmarked yet.


paneq:
Already merged but @pawelpacana @mpraglowski @mlomnicki @pkondzior review welcomed!

Issue #407, opened 2018-07-30
Read specification changed when each_batch used to fetch events


mpraglowski:
Here https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/specification.rb#L118

Read in batched is making changes in Specification definition

Issue #398, opened 2018-07-10
Make RubyEventStore::Specification#result method private


paneq:
As far as I see it does not need to be exposed.

Atul9:
Hi @paneq I checked `ruby_event_store/lib/ruby_event_store/specification.rb` file. I didn't find any method named `result`

paneq:
Hi @Atul9 , I was talking about this:

https://github.com/RailsEventStore/rails_event_store/blob/a609202e664aad027efa25b9c71b2880ddf0defd/ruby_event_store/lib/ruby_event_store/specification.rb#L46

PR #408, opened 2018-07-30
No result in read specification & make it really immutable

mpraglowski:
Fix for #398 and #407

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/specification.rb
+++ b/ruby_event_store/lib/ruby_event_store/specification.rb
@@ -4,68 +4,82 @@ module RubyEventStore
   class Specification
     DEFAULT_BATCH_SIZE = 100
 
-    class Result < Struct.new(:direction, :start, :count, :stream, :read_as, :batch_size)
-      def limit?
-        !count.nil?
-      end
-
-      def limit
-        count || Float::INFINITY
-      end
-
-      def global_stream?
-        stream.global?
-      end
-
-      def stream_name
-        stream.name
-      end
-
-      def head?
-        start.equal?(:head)
-      end
-
-      def forward?
-        direction.equal?(:forward)
-      end
-
-      def backward?
-        !forward?
-      end
-
-      def batched?
-        read_as.equal?(:batch)
-      end
-
-      def first?
-        read_as.equal?(:first)
-      end
-
-      def last?
-        read_as.equal?(:last)
-      end
+    attr_reader :start, :batch_size
+    def limit?
+      !@count.nil?
     end
-    private_constant :Result
 
-    # @api private
-    # @private
-    attr_reader :result
+    def count
+      @count || Float::INFINITY
+    end
 
-    # @api private
-    # @private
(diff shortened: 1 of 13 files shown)


ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -22,15 +22,15 @@ def stream_name
...
+      start.equal?(:head)
     end
 
     def forward?
       @direction.equal?(:forward)
     end
 
     def backward?
-      !forward?
+      @direction.equal?(:backward)
```
pawelpacana: `start` over `@start`, `@direction` over `direction`?

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -4,68 +4,82 @@ module RubyEventStore
...
-    # @private
-    attr_reader :result
+    def last?
+      @read_as.equal?(:last)
+    end
 
-    # @api private
-    # @private
-    def initialize(repository, mapper, result = Result.new(:forward, :head, nil, Stream.new(GLOBAL_STREAM), :all, DEFAULT_BATCH_SIZE))
+    def initialize(repository, mapper,
```
paneq: Why did you remove the `private` marker? We don't want to document this method, it is not intended to be used publicly by RES users.

Hmmm. It seems this we dropped `Result` class and now this is a `Specification` initializer. Would be nice to move it to the top.

paneq: I would consider if `Specification` is a private class, not intended for public documentation.

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -4,68 +4,82 @@ module RubyEventStore
...
     # Limits the query to certain stream.
     # {http://railseventstore.org/docs/read/ Find out more}.
     #
-    # @param stream_name [String] name of the stream to get events from
+    # @param name [String] name of the stream to get events from
     # @return [Specification]
-    def stream(stream_name)
-      Specification.new(repository, mapper, result.dup.tap { |r| r.stream = Stream.new(stream_name) })
+    def stream(name)
+      Specification.new(repository, mapper,
```
paneq: quite verbose. I would consider something along the line of:

```
new = self.dup
new.instance_variable_set(:@stream, name)
new.freeze
new
```

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -190,10 +245,49 @@ def first
...
+    # determine if two objects reference the same hash key.
+    #
+    # This hash is based on
+    # * class
+    # * stream_name
+    # * start
+    # * direction
+    # * read_as
+    # * batch_size
+    def hash
```
paneq: Could we use all `instance_variables` ? Would that be safer? It's hard for me to say if we forgot to include something or not.

mpraglowski:
This was a step in wrong direction. Let's try smaller steps like https://github.com/RailsEventStore/rails_event_store/pull/415

PR #413, opened 2018-08-01
Make RubyEventStore methods return self instead of ok

swistak35:
Issue: #392

### DIFF
--- a/ruby_event_store/spec/client_spec.rb
+++ b/ruby_event_store/spec/client_spec.rb
@@ -25,8 +25,8 @@ module RubyEventStore
       expect(client.read.limit(100).each.to_a).to eq([test_event])
     end
 
-    specify 'delete_stream returns :ok when success' do
-      expect(client.delete_stream(stream)).to eq(:ok)
+    specify 'delete_stream returns client when success' do
+      expect(client.delete_stream(stream)).to eq(client)
     end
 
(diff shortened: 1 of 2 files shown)


ruby_event_store/spec/client_spec.rb
```diff
@@ -7,47 +7,47 @@ module RubyEventStore
     let(:client) { RubyEventStore::Client.new(repository: InMemoryRepository.new) }
     let(:stream) { SecureRandom.uuid }
 
-    specify 'publish returns :ok when success' do
-      expect(client.publish(TestEvent.new)).to eq(:ok)
+    specify 'publish returns self when success' do
+      expect(client.publish(TestEvent.new)).to eq(client)
```
paneq: I usually tested it as `client.publish().publish()` but that's also fine :)

paneq: hmmm that could be good places to start going with mutant annotations:

```
specify 'publish returns self when success', mutant: "RubyEventStore::Client#publish" do
```

which should help with making mutation testing run faster. Do you think you could try that here and in a few other places where you know which method you test.

swistak35: @paneq I'm not a fan (yet) of this `mutant: ` keyword, so I will pass on this here. Maybe another `good first issue` for that?

PR #415, opened 2018-08-02
Extract all read operations on repository from Specification

mpraglowski:
A new class SpecificationReader is now an interface between
Specification and event repository. All reads are done only via reader
class. Also all mapping is done there. Specification just have a reader
and knows nothing about repository & mapper used.


ruby_event_store/lib/ruby_event_store/specification_reader.rb
```diff
@@ -0,0 +1,35 @@
...
+    # @api private
+    # @private
+    def initialize(repository, mapper)
+      @repository = repository
+      @mapper = mapper
+    end
+
+    # @api private
+    # @private
+    def single(specification_result)
```
joelvh: @mpraglowski if `single` isn't a term already established, I'd vote for `one`, which is the word used by ROM

mpraglowski: Or maybe `first` (inspired by Enumerable) - just it could be misunderstood in `Specification#last` method. `first(specification_result)` is actually what we do here.

mpraglowski: Another name could be `find` (also inspired by Enumerable) - the usage will be `find(specification_result)`.

mpraglowski: choices, choices ... ;)

joelvh: I've seen naming move closer and closer to `Enumerable`. Are you thinking that this might be the class to really align with that interface? Any reason not to do that?

pawelpacana: `one` appeals to me more than `single`

mpraglowski: So let's there be `one` ;)

PR #417, opened 2018-08-02
SpecificationResult as a value object

### LINKED ISSUE #398: Make RubyEventStore::Specification#result method private

mpraglowski:
Closes #398


mpraglowski:
@pawelpacana review would be appreciated ;)

Issue #244, opened 2018-03-26
Add AS::Notifications instrumentation


paneq:
* Aggregate sourcing time
* Handler time
* event storing time

Related: #73 #90

pawelpacana:
- serialization time

pawelpacana:
- process manager / saga completion time and event receiving intervals

paneq:
I was looking at #90

and its first case `publish_events`.

```
    def publish_events(events, stream_name: GLOBAL_STREAM, expected_version: :any)
      append_to_stream(events, stream_name: stream_name, expected_version: expected_version)
      events.each do |ev|
        event_broker.notify_subscribers(ev)
      end
      :ok
    end
```

if we instrument this method it's an aggregated statistic for storing events + scheduling async handlers + processing sync handlers. I am not sure how valuable such an aggregated statistic is.

I was thinking.... Perhaps it's best to start with low-level stats. We already know the repository API, it's limited very much.

So I think maybe we could wrap `repository` into a proxy/decorator that can pass all the arguments down but also measure the time and instrument them:

```ruby
module RailsEventStore
  class Client < RubyEventStore::Client
    def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
                   mapper: RubyEventStore::Mappers::Default.new,
                   event_broker: EventBroker.new(dispatcher: ActiveJobDispatcher.new),
                   request_metadata: default_request_metadata,
                   page_size: PAGE_SIZE)
      super(repository: repository,
            mapper: mapper,
            event_broker: event_broker,
            page_size: page_size)
      @request_metadata = request_metadata
    end

    def repository
      @instrumented_repository ||= RepositoryInstrumenterDecorator.new(super)
    end
```

or we could pass a decorated version via `super()`, whatever.

The decorator would be very simple:

```
class RepositoryInstrumenterDecorator
  def initialize(repository)
    @repository = repository
  end

  def append_to_stream(events, stream, expected_version)
    ActiveSupport::Notifications.instrument("append_to_stream.repository.rails_event_store", events: events, stream: stream) do
      repository.append_to_stream(events, stream, expected_version)
    end
  end
  
  # etc etc for link_to_stream, has_event? ...
end
```

I assume we could similarly benchmark `dispatcher` by decorating it.

@pawelpacana @andrzejkrzywda @joelvh @fmwr  what do you think?

pawelpacana:
> Aggregate sourcing time

Moving to separate issue.

[...]

PR #423, opened 2018-08-08
Add instrumentation for repositories

swistak35:
Issue: #244


pawelpacana:
Interesting, I thought that `Client` would be more appropriate than `Repository`. It might be a bit too low-level. And definitely not all methods there deserve instrumenting.

rails_event_store/lib/rails_event_store/instrumented_repository.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+
+    def delete_stream(stream)
+      ActiveSupport::Notifications.instrument("delete_stream.repository.rails_event_store", stream: stream) do
+        repository.delete_stream(stream)
+      end
+    end
+
+    def has_event?(event_id)
```
pawelpacana: meh

swistak35:
My take on instrumentation, learning on what was in #90. I've started with Robert's idea on instrumented repository, because I wanted something easily deliverable. I believe that higher-level instrumentation still can be interesting in cases different than benchmarking (logging maybe?), but that's some start.

Also, I was not sure about instrumenting all public methods of repository, i.e. has_event?(...) - I'm not sure how useful is that instrumentation, except some hardcore benchmarking :P But I guess that in the end, instrumentation is to allow others to play with their ideas how to use it, not only some predefined usecases.

What do you guys think?

rails_event_store/spec/instrumented_repository_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+      InstrumentedRepository.new(RubyEventStore::InMemoryRepository.new)
+    end
+
+    let(:test_race_conditions_any)   { false }
+    let(:test_race_conditions_auto)  { false }
+    let(:test_expected_version_auto) { false }
+    let(:test_link_events_to_stream) { false }
+    let(:test_binary) { false }
+
+    it_behaves_like :event_repository, InstrumentedRepository
```
pawelpacana: 👍

rails_event_store/lib/rails_event_store/client.rb
```diff
@@ -8,7 +8,7 @@ def initialize(repository: RailsEventStoreActiveRecord::EventRepository.new,
                    dispatcher: ActiveJobDispatcher.new,
                    request_metadata: default_request_metadata,
                    page_size: PAGE_SIZE)
-      super(repository: repository,
+      super(repository: InstrumentedRepository.new(repository),
```
pawelpacana: :+1:

rails_event_store/lib/rails_event_store/instrumented_repository.rb
pawelpacana: From one standpoint, as long as this is repository interface, then it deserves instrumenting. On the other hand what's the value of instrumenting this particular one?

swistak35: @pawelpacana check my PR comment, I have big doubt about that one too.

pawelpacana:
Regarding which are worth instrumenting -- from maintainer perspective it is better to add less and reconsider adding more in the future than removing published instruments.

swistak35:
@pawelpacana Which one would you leave? I tend to all except `has_event?` and `last_stream_event`.

pawelpacana:
> @pawelpacana Which one would you leave? I tend to all except has_event? and last_stream_event.

Good call, lets drop those two for now.

rails_event_store/spec/instrumented_repository_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+module RailsEventStore
+  RSpec.describe InstrumentedRepository do
+    subject do
+      InstrumentedRepository.new(RubyEventStore::InMemoryRepository.new)
+    end
+
+    let(:test_race_conditions_any)   { false }
+    let(:test_race_conditions_auto)  { false }
+    let(:test_expected_version_auto) { false }
```
paneq: why `false` ?

rails_event_store/spec/instrumented_repository_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    let(:test_link_events_to_stream) { false }
```
paneq: why `false` here as well ?

rails_event_store/lib/rails_event_store/instrumented_repository.rb
```diff
@@ -0,0 +1,48 @@
+module RailsEventStore
+  class InstrumentedRepository
```
paneq: I like this decorator :)

rails_event_store/spec/client_spec.rb
```diff
@@ -43,5 +43,18 @@ module RailsEventStore
       expect(published.first.metadata[:request_id]).to eq('dummy_id')
       expect(published.first.metadata[:timestamp]).to be_a Time
     end
+
+    specify 'wraps repository into instrumentation' do
```
paneq: 👍

paneq:
The tests are super verbose...

```
      specify "instruments" do
        some_repository = double
        allow(some_repository).to receive(:append_to_stream)
        instrumented_repository = InstrumentedRepository.new(some_repository)
        notification_calls = subscribe_to("append_to_stream.repository.rails_event_store")
        event1 = Object.new
         instrumented_repository.append_to_stream([event1], "SomeStream", "c456845d-2b86-49c1-bdef-89e57b5d86b1")
         expect(notification_calls).to eq([
          { events: [event1], stream: "SomeStream" }
        ])
      end
    end
```

Allow me to offer some alternative:

```
let(:instrumented_repository) { InstrumentedRepository.new(InMemoryRepository.new) }
let(:event) { RubyEventStore::Event.new }
let!(:notification_calls) { subscribe_to(/repository.rails_event_store/) }

      specify "instruments append_to_stream" do
         instrumented_repository.append_to_stream([event], "SomeStream", "c456845d-2b86-49c1-bdef-89e57b5d86b1")
         expect(notification_calls).to eq([
          { events: [event], stream: "SomeStream" }
        ])
      end
    end
```

I believe these 3 `lets` can be easily reused between many specs. Definitely no need to use mocks in every spec when you can just decorate `InMemoryRepo` and have it working with no pain.

The only downside is that we don't verify `exact` notification name but only that it belongs to the namespace `repository.rails_event_store`. But even if you don't go with that the reduced verbosity of not using mocks could be worth dozens lines of tests less.

paneq:
> Interesting, I thought that Client would be more appropriate than Repository. It might be a bit too low-level. And definitely not all methods there deserve instrumenting.

Good point. I felt the same. But on the other hand... It should be relatively easy to write similar decorators for `Broker` and `Mapper` and then one should be able to pin-point exact performance problems to lower-level components and maybe monitoring `Client` would be lower prio. It does not do that much on itself (but it does a few things).

rails_event_store/spec/instrumented_repository_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+        instrumented_repository.read(specification)
+
+        expect(notification_calls).to eq([
+          { specification: specification }
+        ])
+      end
+    end
+
+    def subscribe_to(name)
```
paneq: As we never unsubscribe, doesn't it lead to memory leaks in all RailsEventStore tests?

Wouldn't these arrays continue receiving payloads for all tests (in this file and other files) which call methods on `RailsEventStore::Clients`. It might be better to verify with temporary subscriptions:

```
callback = lambda {|*args| ... }
ActiveSupport::Notifications.subscribed(callback, /rails_event_store/) do
  ...
end
```

paneq:
```
let(:received) { [] }
around do |spec|
  callback = lambda {|*args| received << args }
  ActiveSupport::Notifications.subscribed(callback, /rails_event_store/) do
    spec.call
  end
end
```

or

```
let(:received) { [] }
around do |spec|
  callback = lambda {|*args| received << args }
  ActiveSupport::Notifications.subscribed(callback, "#{spec.metadata[:subscription_name]}.repository.rails_event_store") do
    spec.call
  end
end

specify "instruments append_to_stream", subscription_name: "append_to_stream" do
  # ...
end
```

rails_event_store/spec/instrumented_repository_spec.rb
swistak35: Dunno about these, I've initially skipped these values because they've required some additional methods to be defined. I will look at it.

paneq: It doesn't make sense probably to test race conditions but it does make sense to verify that linking events still works correctly, don't you think?

swistak35:
@paneq mesa like use `InMemoryRepository` instead of mocks, but mesa no like too many `let`s :)

swistak35:
> then one should be able to pin-point exact performance problems to lower-level components

As I've said earlier, I still think higher-level instrumentation can be valuable. As I understand it, instrumentation is not only about _performance problems_, but also about hooking in the right moment because of some particular usecase we can not think about. Where are you, AOP ;)

paneq:
> As I've said earlier, I still think higher-level instrumentation can be valuable. 

That's unquestionable. It's rather a matter of priorities. Do we start with low-level instrumentation or higher-level.

> As I understand it, instrumentation is not only about performance problems, but also about hooking in the right moment because of some particular usecase we can not think about. Where are you, AOP ;)

Interesting. I don't see it much that way because that's only instrumentation and not AOP :) You can't affect the code from within the subscribed code.

paneq:
@swistak35 any pland for documentation and RDoc?

swistak35:
@paneq yeah, I haven't started it because I expected that the implementation may change rapidly, I'd prefer to make documentation in separate PR after merge

paneq:
good for me :)

pawelpacana:
Let's deal with `brew cask` issue on master.

PR #429, opened 2018-08-10
Correlation Causation readme fix

piotrjaworski:
Hey,

I noticed a small problem when I was reading the Correlation Causation Readme page. 

I think that it should be fine now.

Thanks,
Piotrek


paneq:
@piotrjaworski Thank you and have a lovely day!

PR #433, opened 2018-08-10
Rails Event Store ActiveRecord - wrong Rails version in a migration file

piotrjaworski:
Active migration is generated incorrectly - if a Rails version is 5.0+, a version number is the class name is always set to `4.2.`, example:
`class CreateEventStoreEvents < ActiveRecord::Migration[4.2]`

This PR fixes this behavior and by default sets the version number to the real version number with sub number, for example `[5.2]`.


piotrjaworski:
@paneq Not sure if it was set to `4.2.` by a reason, changed that to a proper version number.

paneq:
@piotrjaworski Thank you for your contribution milorodzie.

However, I recommend searching historical issues and pull requests before submitting a new one:

 https://github.com/RailsEventStore/rails_event_store/pull/208#issuecomment-371106333

Issue #214, opened 2018-03-13
Pass RES client instance to Browser


pawelpacana:
It would be useful to be allowed to pass particular RES client instance to browser. Currently we look into usual place and that works for most of us:
https://github.com/RailsEventStore/rails_event_store/blob/2eb821c73cc3c6aa63dfe3a9680a0b689ae71193/rails_event_store-browser/app/controllers/rails_event_store/browser/application_controller.rb#L8-L10

I could however imagine having client instance with configurations somewhere else.

pawelpacana:
One way to do it: https://stackoverflow.com/questions/20734766/rails-mountable-engine-how-should-apps-set-configuration-variables

pawelpacana:
If Browser was _just_ a Rack app, then we could probably pass initialized app instance here: http://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Base.html#method-i-mount

But it's a Rails Engine so likely dragons on that way. Ideas?

pawelpacana:
Done in new browser version.

joelvh:
Didn’t even see this ticket before the rework!

PR #434, opened 2018-08-10
Separated Elm out into Sinatra app and share classes with Rails Engine

joelvh:
* Created `ruby_event_store-browser` gem
  * Moved Elm front-end project to new gem
  * Encapsulate controller logic into classes
* Updated `rails_event_store-browser` gem
  * Reference `ruby_event_store-browser` classes to use same backend logic
  * Updated `make` command to build Elm project and copy to `public` folder


joelvh:
@paneq have you noticed frequent network issues with Travis CI? the failed tests are purely setup issues and re-running them seems to fix things....

joelvh:
@paneq @pawelpacana thoughts on this decoupling from Rails?

joelvh:
It moves the Elm project to a new gem, which is potentially the "biggest" shift.

pawelpacana:
@joelvh I didn't have a chance to look yet, scheduling on my evening

pawelpacana:
A couple of observations:
- making Browser work outside Rails is a good thing :+1:
- I was afraid some convenience for Rails would be lost but it isn't the case as you had to manually mount the app before
-  super cool for allowing to pass event_store instance (https://github.com/RailsEventStore/rails_event_store/issues/214)
- I have some doubts about RubyEventStore rename but that's minor -- mostly due to unavoidable conflict between RubyEventStore vs. RailsEventStore

Aside, to give some background -- the idea with RailsEventStore is that you get preconfigured set of tools that integrate well with Rails with some of the decisions being already made (convention :wink:).
RubyEventStore is on the other hand something you'd likely assemble on your own (as RailsEventStore is built from the same building blocks).
From marketing POV RailsEventStore allows to reach wider audience and I believe it is where majority of our users are. Bu we know there's nothing preventing you to use it without Rails (and Browser is now Rails free 🎈).

pawelpacana:
Would merge :)

Makefile
```diff
@@ -4,14 +4,15 @@ CURRENT_REV  = `git rev-parse HEAD`
...
 GEMS         = aggregate_root \
-	       bounded_context \
-	       ruby_event_store \
-	       ruby_event_store-rom \
-	       rails_event_store \
-	       rails_event_store_active_record \
-	       rails_event_store_active_record-legacy \
-	       rails_event_store-browser \
-	       rails_event_store-rspec
+							 bounded_context \
```
antonpaisov: what's the reason behind this indentation? :)

joelvh:
@pawelpacana thanks for the review! I totally understand and keep in consideration that desire to keep this project aligned with the Rails community, to allow the RES to get wide-spread adoption more easily. When it comes to naming for the non-Rails components, I typically go for name changes that make sense from a building-block perspective, but definitely look for your input to see if there are better ways to achieve the branding you're going for without causing confusion.

To reiterate, I am aligned with you to keep the Rails community as the priority and I'm working to continue making the Ruby* components building blocks to be used in various contexts. My goal is to continue expanding the Ruby* components as a toolkit and have Rails* be an implementation of those.

In terms of merging, let me know if you want me to merge it or I should wait for more feedback.

Thanks!

Makefile
joelvh: @antonpaisov good question. I am not sure what happened there. Will see if I can fix it.

joelvh:
@pawelpacana just pushed some updates to the docs. I used `[ci-skip]`, but it seems tests might still be triggered?

pawelpacana:
Long time no release, we'd need to describe this in a releases page :)

I'm wondering how to deprecate old browser (notify about change without making it unusable yet), while removing the actual code of previous browser. Maybe we could invoke new code from Engine wrapper while emitting a warning?

pawelpacana:
> Maybe we could invoke new code from Engine wrapper while emitting a warning?

Silly me, I forgot this is already happenning 👌

PR #441, opened 2018-08-27
Streams of event

mpraglowski:
Finds all streams for given event id, excluding global stream.


paneq:
> Finds all streams for given event id, excluding global stream.

Could you elaborate a little bit about what could be the usecase for that?

pawelpacana:
@mpraglowski Care to describe this in changelog for release?

Issue #400, opened 2018-07-19
Serialization/Mapping pipeline idea


paneq:
### Goals

* make mapping/serialization more composable
  * it should be possible to compress or encrypt events no matter what is used serializing them
* assuming downstream repository can persist more data make it possible to expose additional data beyond just event id, serialized data, serialized metadata, event type. This could be
  * keys or identifiers which can be helpful for encrypting events
  * additional metadata
  * event version
  * etc
* Solve issues with JSON and symbols
* Allow easier adding of default fields to deserialized events

Related: #367 #361 #295 #189 #307 

### Execution

Use open `Hash` instead of `SerializedRecord` for passing data from domain event to list of serialized strings and back.

### Benefits

* every pipe (step) in the pipeline can override or extend what's in the Hash freely
  * so it can for example change `hash[:data]` from compressed to uncompressed string
  * but it can also extract something from top level domain event which can be used by next layer
* that makes it possible to pass data between steps which are not next to each other

### Code

```ruby
class Default
  def initialize(events_class_remapping: {})
    @events_class_remapping = events_class_remapping
  end

  def serialize(hash)
    domain_event = hash.fetch(:domain_event)
    hash.merge!({
      domain_event: domain_event,
      event_id:   domain_event.event_id,
      metadata:   domain_event.metadata.to_h,
      data:       domain_event.data,
      event_type: domain_event.type
    })
    hash
  end

  def deserialize(hash)
    event_type = hash.fetch(:event_type)
    event_type = events_class_remapping.fetch(event_type) { event_type }
    hash[:domain_event] = Object.const_get(event_type).new(
      event_id: hash.fetch(:event_id),
      metadata: hash.fetch(:metadata),
      data:     hash.fetch(:data),
    )
    hash
  end

  private

  attr_reader :events_class_remapping
end

class SymbolizeMetadataKeys
  def serialize(hash)
    hash.tap{|h| h[:metadata] = h[:metadata].stringify_keys }
  end

  def deserialize(hash)
    hash.tap{|h| h[:metadata] = h[:metadata].symbolize_keys }
  end
end

class Protobuf
  def initialize(events_class_remapping: {})
    require_optional_dependency
    @events_class_remapping = events_class_remapping
  end

  def serialize(hash)
    data = hash.fetch(:data)
    hash[:metadata] = ProtobufNestedStruct::HashMapStringValue.dump(hash[:metadata])
    hash[:data] = data.class.encode(data)
    hash
  end

[...]

PR #442, opened 2018-08-29
A spike for the dispatchers refactoring

swistak35:
I wanted to know what do you think about the following refactoring of the code around dispatchers.
The commit present just a spike, I'm not even sure it entirely works, I wanted the comments on the direction.

The things I didn't like:
1) `AsyncDispatcher` inheriting from the `Dispatcher`, and then `ActiveJobDispatcher` inheriting on `AsyncDispatcher`
2) concept of `async_proxy_strategy` which isn't descriptive until you'll take a look at `AsyncProxyStrategy::AfterCommit`
3) the fact that `AsyncDispatcher` was not really async, but "maybe async, maybe sync", and in the end the only fact that `AsyncDispatcher` was doing was this "maybeness" - the rest of the actual async-related code was either in `async_proxy_strategy` or specific scheduler
4) `async_handler?` naming in scheduler -- it was leaking into the scheduler's, that the actual "scheduling" is "maybe async, maybe sync", and scheduler is making decisions.

The things I do like about new code:
1) `async_proxy_strategy` is no more, now there are two different dispatchers: `ImmediateAsyncDispatcher` and `AfterCommitAsyncDispatcher`, neither of them inheriting on `Dispatcher` (because they don't share even any code)
2) The "maybeness" is handled by another, composing, dispatcher. I see two routes here, both included in this spike, but in the final code it probable make sense to choose only one.

First is `MaybeAsyncDispatcher`, which is similar to the old `AsyncDispatcher`. It's semantics is clear, it is "maybe async, maybe sync" and one only have to pass chosen implementation of asynchronicity and the strategy how to choose when the subscriber is async or not. I don't think I like that solution very much, but I wanted to add it for completion.

Second is `ComposedDispatcher`, which is (obviously) composed dispatcher. It dispatches to the first dispatcher, for which `strategy` for given `subscriber` is true. Old `AsyncDispatcher` could be implemented (roughly) by 

```
active_job_scheduler = ActiveJobScheduler.new
ActiveJobDispatcher = ComposedDispatcher.new
  .add_dispatcher(AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler), ->(subscriber) { Class === subscriber && subscriber < ActiveJob::Base })
  .add_dispatcher(PubSub::Dispatcher, -> { true })
```

3) In the async dispatchers, scheduler knows how to verify it. It did makes sense to me, because scheduler is usually an `ActiveJob` or `Sidekiq` or something like that, and "verification" is closely related to that implementation (responding to the `perform_async` or having `sidekiq_options` method, etc.).

4) I think I'm able to do all these changes with all proper deprecations :) (i.e. not removing some code from version to version)

Things I don't like in new solution and I'm especially open to feedback:

1) That whole passing `strategy` to the `ComposedDispatcher`. It could have been easier by just using `verify` on dispatcher and using the first `verify` that does not raise an error, but then `verify` would have two responsibilities and I guess that's not good. But maybe there's something in between, like allowing to pass strategy, but in default case use `verify`...


paneq:
@swistak35 it seems you thought this thoroughly :) I agree with almost everything you said.

These comments are coming without even looking at the code, purely based on the proposed API in the description.

I prefer more flexible `ComposedDispatcher` over `MaybeAsyncDispatcher` on the grounds similar to #400 . I now see these things more in terms of _pipes_ rather than _delegation_.

But:

```ruby
ActiveJobDispatcher = ComposedDispatcher.new
  .add_dispatcher(AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler), ->(subscriber) { Class === subscriber && subscriber < ActiveJob::Base })
  .add_dispatcher(PubSub::Dispatcher, -> { true })
```

there is something not right here... Ideally that would be:

```
ComposedDispatcher.new(one_thing, the_other_thing)
```

or

```
ComposedDispatcher.new.add(one_thing).add(the_other_thing)
```

Which means to me that the second argument such as `->(subscriber) { Class === subscriber && subscriber < ActiveJob::Base }` or `-> { true }` should be a method on the passed thing.

Now the problem with that is `AfterCommitAsyncDispatcher` does not know how to recognize _async_ because it depends on the passed scheduler (which might be ActiveJob related or Sidekiq related etc). What's the solution? Delegate to `scheduler` in such case. As you already said these things are coupled to the final implementation of doing the scheduling anyway. I am not sure if that would be done with `verify` or other method.

```ruby
async = AfterCommitAsyncDispatcher.new(scheduler: ActiveJobScheduler.new)
sync = PubSub::Dispatcher.new
ActiveJobDispatcher = ComposedDispatcher.new(async, sync)
```

What do you think? Does it make sense? Is it possible?

Inspiration: middlewares in general (rack, rails etc).

swistak35:
> Now the problem with that is AfterCommitAsyncDispatcher does not know how to recognize async because it depends on the passed scheduler (which might be ActiveJob related or Sidekiq related etc). What's the solution? Delegate to scheduler in such case. As you already said these things are coupled to the final implementation of doing the scheduling anyway. I am not sure if that would be done with verify or other method.

I'm not sure whether pipes are good approach here. In `mappers`, we usually want more than one `mapper` to do something, so pipes fit in. Here, do we want to support dispatch to more than one dispatcher (maybe that's good sci-fi, not sure ;) )? Dispatcher is more of like adapter to me, where it rarely make sense to have more than one of them (similar case as repository - we have only one).

---

Other API suggestion, still assuming delegation: Maybe blocks?

```
ComposedDispatcher.new
  .add(async_dispatcher) { |s| ... custom rule whether subscriber should be handled by this dispatcher }
  .add(sync_dispatcher) # no block, use `verify` instead
```

paneq:
> Dispatcher is more of like adapter to me, where it rarely make sense to have more than one of them 

And yet we are talking about composed dispatcher created from having 2 of them. What if you want to schedule to Amazon SQS or something similar? I guess that would be a third.

paneq:
```
.add(async_dispatcher) { |s| ... custom rule whether subscriber should be handled by this dispatcher }
```

Still no idea why would this rule be outside a passed object?

swistak35:
*tl;dr* I still have some doubts, but small enough to go with pipes instead

> Still no idea why would this rule be outside a passed object?

If we think about dispatcher in terms of pipes, there is no reason (and it should be in the object). If we think in terms of delegation, then we don't want the object to decide whether we want to delegate to it or not, because it's not delegation anymore.

> And yet we are talking about composed dispatcher created from having 2 of them. What if you want to schedule to Amazon SQS or something similar? I guess that would be a third.

That example with external queue almost convince me :)

One more doubt: Let's assume that scheduler has one more method `handle?` to answer whether it should be handled by this dispatcher or not (which probably is very similar to `verify` in implementation). Now, if I want some class to be async-able (so it inherits from `ActiveJob` or something), but I want to use the class as a sync handler, I need to make new scheduler, inheriting on the `ActiveJobScheduler`, implement my new `handle?`, and then pass to the client new composed dispatcher with async dispatcher using my new scheduler. That's pretty "deep" change. If `ComposedDispatcher` would be delegation on other hand, I would only make these changes using it's API.

paneq:
> If we think in terms of delegation, then we don't want the object to decide whether we want to delegate to it or not, because it's not delegation anymore.

```ruby
potential_dispatchers.each do |dispatcher|
  next unless dispacher.can_handle?(something)
  return dispatcher.dispatch(something)
end
raise "ojojoj, no capable dispatcher found, abort"
```

Looks _delegatish_ enough for me. I don't see what _specific_ objections we could have for writing `ComposedDispatcher` in that style? I just think of it in terms of cooperation between two objects.

P.S. I am not sure if we should dispatch with ALL capable dispatchers or first one matched. But I think that's other discussion. One can always compose dispatchers through custom code to behave properly anyway.

paneq:
> One more doubt:

That's complicated. I need more time :)

mpraglowski:
What if we will separate decision from dispatching completely?

In basic setup (all sync) we will just have:

```
RubyEventStore::Client.new(dispatcher: PubSub::Dispatcher.new)
```

The same if we want to have all hanslers processed async:

```
RubyEventStore::Client.new(dispatcher: AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler))
```

No strategy. No problems.

In a system where we will have sync & async handlers mixed we might use `ComposedDispatcher`:

```
strategy = DispatchStrategy.new
dispatcher = ComposedDispatcher.new(
  AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler),
  ImmediateAsyncDispatcher.new(scheduler: active_job_scheduler),
  default: PubSub::Dispatcher.new)
RubyEventStore::Client.new(dispatcher: AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler))
```

Also `DispatchStrategy` coudl answer the question "I am not sure if we should dispatch with ALL capable dispatchers or first one matched.".
Default implementation should dispatch to single dispatcher (compatibility with current code) but final decision will be made by developer.

Also scheduler should be just a scheduler. No `verify(subscriber)` method. That will be decided by `DispatchStrategy`.

swistak35:
> What if we will separate decision from dispatching completely?
> 
> In basic setup (all sync) we will just have:
> 
> `RubyEventStore::Client.new(dispatcher: PubSub::Dispatcher.new)`
> 
> The same if we want to have all hanslers processed async:
> 
> `RubyEventStore::Client.new(dispatcher: AfterCommitAsyncDispatcher.new(scheduler: active_job_scheduler))`
> 
> No strategy. No problems.
> 
> In a system where we will have sync & async handlers mixed we might use ComposedDispatcher:
> ...

Isn't that exactly the same I've suggested in the code? Or am I missing something?

---
re: Composed dispatcher example. Could you show how you would implement the Composed Dispatcher (roughly ;)?

> Also scheduler should be just a scheduler. No verify(subscriber) method. That will be decided by DispatchStrategy.

How do you understand `verify`? I understand it as "this object is a thing my scheduler understand". Why wouldn't it be part of scheduler? For `ActiveJob` it is checking inheritance on ActiveJob, for sidekiq it would probably check whether it has `Sidekiq::Worker` included... If I understand this `verify` correctly I see no point in separating it out of scheduler.

mpraglowski:
> Isn't that exactly the same I've suggested in the code? Or am I missing something?

Very similar. But no blocks on adding dispatchers, no `handle?` (or `verify`) methods.


> How do you understand verify? I understand it as "this object is a thing my scheduler understand". Why wouldn't it be part of scheduler? For ActiveJob it is checking inheritance on ActiveJob, for sidekiq it would probably check whether it has Sidekiq::Worker included... If I understand this verify correctly I see no point in separating it out of scheduler.

Verification on scheduler is too late. I would skip that and replace that by developer's decision by implementing `DispatchStrategy`. And then if the decision was wrong just let it fail.

swistak35:
Done in #445

Issue #443, opened 2018-08-29
API for changing events


swistak35:
From time to time, developers may decide that they want to modify the events, i. e. enrich them. Obviously we can agree that there's always other solution (like enrichment in application code, for example). However, considering the fact that we usually use SQL database underneath, there's probably no technical reason why wouldn't we provide at least some basic API to change the events. Now, changing events require using internal details of repository, pushing serialized record through mappers (in two directions) and finally also calling some raw persistence calls (like `save!` in case of Active Record).

joelvh:
@swistak35 I was thinking about this exact thing today as well, as we have a desire to rewrite event data rather than handle in code early in a project when some naming has changed. I'm curious how you are thinking of adding this capability. I see it as something similar to the projection functionality, but also would want to implement this in a way that doesn't encourage abuse or using it inefficiently.

paneq:
```
require 'ruby_event_store/danger_zone'

# events: contains events with same ID as existing ones but updated data or metadata
client.overwrite(events)
```

I believe that would be sufficient top-level API.

This API would be useful for remapping with different mapper. You instantiate events with one Client instance using old mapper, change them (or not) and pass to another client (with new mapper configured) to `overwrite` and it updates their serialized value in DB.

It would probably be implemented similar to:

```ruby
    def overwrite(events)
      serialized_events = serialize_events(events) # no enrichment !!!
      repository.overwrite(serialized_events)
      self
    end
```

paneq:
Whole operation of overwriting N events should be atomic imho. It should not affect streams the event is in.

I can confirm it would be useful in other projects as well!

joelvh:
I assume there would still be a need for mapping events to handle old and new format, and then the old format of the event can be removed from the code in a subsequent deployment — is that how you think of it? If the event name stays the same, the event class needs to somehow tell if it’s old or new data format until the “migration” of the event data is complete.

I haven’t messed with this yet, but imagine that’s the approach.

swistak35:
In our case it was like this:
1) handle both old and new version in application code (by enrichment, for example)
2) overwrite events' data
3) remove application code handling old version

So we didn't need from event store anything more than the API to overwrite. I imagine there could be some API supporting enrichmnet of events in application code, but I would say that could be separate issue :) If there would be an API for such enrichment, maybe we could event have some automated script to overwrite all events using enriched attributes...

paneq:
[...]

PR #444, opened 2018-09-03
Changing messages

paneq:
Issue: #443

### DIFF
--- a/rails_event_store_active_record/Makefile
+++ b/rails_event_store_active_record/Makefile
@@ -3,7 +3,7 @@ GEM_NAME     = rails_event_store_active_record
 REQUIRE      = $(GEM_NAME)
 IGNORE       = RailsEventStoreActiveRecord::IndexViolationDetector\#detect \
                RailsEventStoreActiveRecord::PgLinearizedEventRepository* \
-               RailsEventStoreActiveRecord::EventRepository#update_messages
+               RailsEventStoreActiveRecord::EventRepository\#update_messages
 SUBJECT      ?= RailsEventStoreActiveRecord*
 DATABASE_URL ?= sqlite3::memory:
 
--- a/rails_event_store_active_record/Makefile
+++ b/rails_event_store_active_record/Makefile
@@ -2,7 +2,8 @@ GEM_VERSION  = $(shell cat ../RES_VERSION)
 GEM_NAME     = rails_event_store_active_record
 REQUIRE      = $(GEM_NAME)
 IGNORE       = RailsEventStoreActiveRecord::IndexViolationDetector\#detect \
-               RailsEventStoreActiveRecord::PgLinearizedEventRepository*
+               RailsEventStoreActiveRecord::PgLinearizedEventRepository* \
+               RailsEventStoreActiveRecord::EventRepository#update_messages
 SUBJECT      ?= RailsEventStoreActiveRecord*
 DATABASE_URL ?= sqlite3::memory:
 
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -49,9 +49,9 @@ module RailsEventStoreActiveRecord
     def update_messages(messages)
       hashes = messages.map(&:to_h)
       hashes.each{|h| h[:id] = h.delete(:event_id) }
-      for_update = messages.map(&:event_id).to_set
+      for_update = messages.map(&:event_id)
       start_transaction do
-        existing = Event.where(id: for_update).pluck(:id).to_set
+        existing = Event.where(id: for_update).pluck(:id)
         (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
         Event.import(hashes, on_duplicate_key_update: [:data, :metadata, :event_type])
       end
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -46,6 +46,17 @@ module RailsEventStoreActiveRecord
       @repo_reader.read(specification)
     end
 
+    def update_messages(messages)
+      hashes = messages.map(&:to_h)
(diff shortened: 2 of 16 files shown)


rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -46,6 +46,17 @@ def read(specification)
       @repo_reader.read(specification)
     end
 
+    def update_messages(messages)
+      hashes = messages.map(&:to_h)
+      hashes.each{|h| h[:id] = h.delete(:event_id) }
+      for_update = messages.map(&:event_id).to_set
+      start_transaction do
+        existing = Event.where(id: for_update).pluck(:id).to_set
+        (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
```
swistak35: Maybe `tap` on `try(:first)` instead of `each`? I'm curious whether mutant will complain on that `each` :)

rails_event_store_active_record/Makefile
```diff
@@ -2,7 +2,8 @@ GEM_VERSION  = $(shell cat ../RES_VERSION)
 GEM_NAME     = rails_event_store_active_record
 REQUIRE      = $(GEM_NAME)
 IGNORE       = RailsEventStoreActiveRecord::IndexViolationDetector\#detect \
-               RailsEventStoreActiveRecord::PgLinearizedEventRepository*
+               RailsEventStoreActiveRecord::PgLinearizedEventRepository* \
+               RailsEventStoreActiveRecord::EventRepository#update_messages
```
swistak35: We could add a separate task for it, because when https://github.com/zdennis/activerecord-import/issues/540 will be done we would probably be able to do that

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: This is _easier_ imho to do nothing for empty case compared to try/if statements.

rails_event_store_active_record/Makefile
paneq: 👍

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -46,6 +46,17 @@ def read(specification)
...
     end
 
+    def update_messages(messages)
+      hashes = messages.map(&:to_h)
+      hashes.each{|h| h[:id] = h.delete(:event_id) }
+      for_update = messages.map(&:event_id)
+      start_transaction do
+        existing = Event.where(id: for_update).pluck(:id)
+        (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
+        Event.import(hashes, on_duplicate_key_update: [:data, :metadata, :event_type])
```
joelvh: @paneq should this import in batches? How does this kind of rewriting of many events impact things?

Also, wonder if "rewrite" is a term to consider.

joelvh:
@paneq i'll take a look at the ROM implementation when I get a chance.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
paneq: @joelvh I don't understand this question "should this import in batches?"

Obviously I believed it should, otherwise I wouldn't write it that way. If you believe it shouldn't please state the reasons.

> How does this kind of rewriting of many events impact things?

What _things_ exactly do you have in mind ? How many is many? You can pass a single event here and you can pass a thousand as well. I was yesterday rewriting in batches of 10K and it worked fine for me.

> Also, wonder if "rewrite" is a term to consider.

sounds like a good option.

joelvh: @paneq right, my question wasn't clear. Looking at this method in isolation, the `start_transaction` part made me think I'd want to rewrite all events in a single transaction, so it made me think that this was handling all changes in one call. But what you're indicating makes sense, which is that you might batch on the read side, and call `update_messages` for each batch. If errors occur, then the transaction would roll back for that last batch.

Regarding "things", I was generalizing about what considerations should be called out in using this API around performance. Rewriting many many events in one transaction is probably something to be avoided. Rewriting in batches is more efficient and the code should handle old and new events until all events are successfully migrated.

joelvh: And regarding "many" -- that is obviously relative, point being whether some best practices should be called out (maybe in the docs) about how to use this API. My last comment clarifying what I meant with batches is related to clarifying how you would think about using this API for, let's say 100k-1m events. (I haven't used this volume myself yet, so I don't know what volume is a lot or an unreasonable amount of events to expect to process or load into memory.)

Hope that helps clarify the view point I was coming from.

joelvh: P.S. One last note: I haven't yet had to use all features of RES, particularly batching events and things related to retroactive projections. (I'm updating read models with event handlers only, right now.) I'm on the verge of diving into those features, so that is why maybe some of my questions sound naive in relation to RES and existing best practices. I am currently planning out building a system for rewriting events and also building up new read models from existing events (using RES projections) in the coming weeks. Maybe that will give me better context when I get into that.

paneq: > Rewriting many many events in one transaction is probably something to be avoided. Rewriting in batches is more efficient and the code should handle old and new events until all events are successfully migrated.

Indeed. That's what's shown in docs. But it depends if you are doing live migration or have a planned downtime.

> How you would think about using this API for, let's say 100k-1m events

Tests on my local machine indicated that updating in batches of 1000 and 10_000 gave me the same performance. But 1000 was definitely faster than 100. I think you can apply all your intuitions and experiences from updating rows in any SQL db. 

The set:
* must fit into memory
* must not generate too big sql statement

joelvh:
@paneq working on the ROM implementation. Saw that `update_messages` uses `Event.import` with upsert options. What SQL databases is RES meaning to support?

PR #445, opened 2018-09-04
Dispatchers refactoring

swistak35:
Real implementation for spike from #442 

Easy to-dos:

- [ ] linter for schedulers
- [ ] adding deprecation messages
- [ ] changing default RailsES client

---

A question about `verify`.
Before:
* it was raising an error when the handler was incorrect
* in practice it was used only in `PubSub::Dispatcher`, `AsyncDispatcher` didn't care

Now:
* it still raise an error
* it is used correctly everywhere

However, because of error-driven control flow, `ComposedDispatcher` looks pretty ugly. Adding `handle?` to dispatchers would solve one "ugliness", but it won't help in `#verify`.

1) Leave `#verify` it as it is and add `handle?` method?
2) Refactor `verify` so it is returning value instead of exception?
3) other suggestion?

I'm in favor of 1), because:
* Using 2) increase the scope (and deprecated changes) of this PR even more
* using exceptions for control flow here has an advantage of easy way to pass nicely formatted information why the handler was wrong


rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
```diff
@@ -0,0 +1,50 @@
...
+  class AfterCommitAsyncDispatcher
+    def initialize(scheduler:)
+      @scheduler = scheduler
+    end
+
+    def call(subscriber, _, serialized_event)
+      run(->() { @scheduler.call(subscriber, serialized_event) })
+    end
+
+    def run(schedule_proc)
```
paneq: why not `&schedule_proc` ?

rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
```diff
@@ -0,0 +1,50 @@
+module RailsEventStore
+  class AfterCommitAsyncDispatcher
+    def initialize(scheduler:)
+      @scheduler = scheduler
+    end
+
+    def call(subscriber, _, serialized_event)
+      run(->() { @scheduler.call(subscriber, serialized_event) })
```
paneq:
```
run{ @scheduler.call(subscriber, serialized_event) }
```

rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
paneq: Why is this `run` method public?

ruby_event_store/lib/ruby_event_store/immediate_async_dispatcher.rb
```diff
@@ -0,0 +1,15 @@
+module RubyEventStore
+  class ImmediateAsyncDispatcher
+    def initialize(scheduler:)
+      @scheduler = scheduler
+    end
+
+    def call(subscriber, _, serialized_event)
+      @scheduler.call(subscriber, serialized_event)
+    end
```
paneq: Notice that if scheduler `call` API was identical to dispatcher's API, you could simply substitute scheduler for a dispatcher without the need for `ImmediateAsyncDispatcher`. Wouldn't that be nice?

rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
paneq: Notice that if scheduler call API was identical to dispatcher's API, you could simply substitute scheduler for another dispatcher. Wouldn't that be nice?

```
AfterCommitAsyncDispatcher.new(ActiveJobDispatcher.new)
```

paneq:
@swistak35 

> 2. Refactor verify so it is returning value instead of exception?

Specifically `true` / `false` sounds like a better option.

I usually don't have anything against exceptions. But here we are talking about:

* code executed in a loop
* case where probability for exception is high. It's possibly you have only very few async handler and tons of sync ones in a hypothetical codebase.

> 3. other suggestion?

* drop `verify`. We don't need to check handlers at registration time to see if any dispatcher can handle them. That's what tests are for. If tests pass, it works.
* return `false` from `call` if dispatcher can't handle it. Higher layer (`ComposedDispatcher`, `PubSub::Broker`) can interpret it and react:
  * `ComposedDispatcher` - try next in list. If next not available, return `false`
  * `PubSub::Broker` - raise an exception if dispatcher returned `false`.

Oooooor:

* Make dispatchers configurable in regards to whether they return bool or raise an exception:

    ComposedDispatcher.new(Async.new(raise: false), Sync.new(raise: false))

* But I don't think I like this approach :)

swistak35:
Changes in the PR (written now to add later):

rbes:
* Change: Deprecate old `AsyncDispatcher` and `AsyncProxyStrategy::Inline`. Now if you want to have only asynchronous inline dispatch, you can use `ImmediateAsyncDispatcher`. If you want to have old behaviour (both async and sync dispatcher), use `ComposedDispatcher.new(ImmediateAsyncDispatcher.new(scheduler: scheduler), PubSub::Dispatcher.new)`
* Change: Dispatchers should no longer raise errors on `verify`, but return `true/false` instead. It is especially important if you want to use new `ComposedDispatcher`, not relevant otherwise.
* Add: `ComposedDispatcher` -- new dispatcher, which accepts at least one dispatcher and dispatch the event to the first dispatcher which accepts the subscriber.
* Add: linter for schedulers
* Fix: `RubyEventStore::InvalidHandler` now does not return very customized (and often wrong) error message, it's now simple error inheriting on `StandardError`.

res:
* Change: Deprecate `RailsEventStore::ActiveJobDispatcher` and `RailsEventStore::ActiveJobDispatcher::ActiveJobScheduler`. Read `RubyEventStore` changelog with information how to use new dispatchers, and use new `RailsEventStore::ActiveJobScheduler` scheduler.
* Change: Deprecate `RailsEventStore::AsyncProxyStrategy::AfterCommit`. Use `AfterCommitAsyncDispatcher` instead.
* Change: `RailsEventStore` default dispatcher has now changed to use new dispatchers, but still preserving old overall behaviour (dispatch asynchronous if possible, synchronous otherwise)

PR #452, opened 2018-09-20
Get list of streams of given event

mpraglowski:
Any stream where event is stored or linked.

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/client.rb
+++ b/ruby_event_store/lib/ruby_event_store/client.rb
@@ -194,6 +194,13 @@ module RubyEventStore
       Specification.new(SpecificationReader.new(repository, mapper))
     end
 
+    # Gets list of stream names where event is stored or linked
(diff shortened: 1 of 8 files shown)


ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -194,6 +194,13 @@ def read
       Specification.new(SpecificationReader.new(repository, mapper))
     end
 
+    # Gets list of stream names where event is stored or linked
+    #
+    # @return [Array<String>] names of streams where event is stored or linked
+    def streams_of(event_id)
+      repository.streams_of(event_id)
```
andrzejkrzywda: would it make sense to return a collection of Stream objects? Given that we already have them as a kind of value objects.

mpraglowski: yeah - makes sense

PR #461, opened 2018-09-27
2.6.0 missing test entries

pawelpacana:


pawelpacana:
Somehow it behaves slightly different than on macOS — the error has a bit different message

PR #463, opened 2018-09-30
Update messages with ROM repository

joelvh:
* Implements `update_messages`
* Uses concatenated SQL `UPDATE` statements to batch update event records rather than utilizing "upsert" method used by AR repository

@pawelpacana @paneq do you prefer using "upsert" updates rather than SQL `UPDATE`?


joelvh:
Closing in favor of #476

Issue #456, opened 2018-09-25
Debugging API - filtering by event types


swistak35:
When debugging / console-work, one often wants to filter the events by their type.
Thus, we probably would like to allow to do that in our specification reader?

pawelpacana:
Would it make sense to have such filtering available in Browser as well?

PR #472, opened 2018-10-11
Filtering by event types

mpraglowski:
It solves #456

### DIFF
--- a/rails_event_store_active_record-legacy/lib/rails_event_store_active_record/legacy/generators/templates/v1_v2_migration_template.rb
+++ b/rails_event_store_active_record-legacy/lib/rails_event_store_active_record/legacy/generators/templates/v1_v2_migration_template.rb
@@ -51,7 +51,6 @@ class MigrateResSchemaV1ToV2 < ActiveRecord::Migration<%= migration_version %>
     add_index :event_store_events_in_streams, [:stream, :event_id], unique: true
     add_index :event_store_events_in_streams, [:created_at]
 
-    remove_index  :event_store_events, :event_type
     remove_column :event_store_events, :stream
     remove_column :event_store_events, :id
     rename_column :event_store_events, :event_id, :id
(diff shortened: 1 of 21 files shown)


mpraglowski:
@joelvh again need your help here :( Memory adapter needs some attention.

joelvh:
Fixed in #475

joelvh:
@mpraglowski were you able to merge my changes in?

mpraglowski:
@joelvh yes - I will close #475 - your changes are already here

joelvh:
Awesome, thanks!

pawelpacana:
Worth reminding: https://github.com/RailsEventStore/rails_event_store/issues/72

ruby_event_store-rom/db/migrate/20181026152045_index_by_event_type.rb
```diff
@@ -0,0 +1,9 @@
+require 'rom/sql'
+
+::ROM::SQL.migration do
+  change do
+    alter_table(:event_store_events) do
+      add_index :event_type
```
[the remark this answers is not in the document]
joelvh: ROM uses Sequel under the hood. You could use `add_index :event_type, name: 'event_store_events_event_type_index', if_not_exists : true`

The  `if_not_exists` option is PG only. Would that work? (Seems MySQL requires a query or stored procedure to check if the index exists by name.)

Also, maybe use the index name that Rails generates in the AR migration?

https://sequel.jeremyevans.net/rdoc/classes/Sequel/Schema/AlterTableGenerator.html#method-i-add_index

rails_event_store_active_record-legacy/lib/rails_event_store_active_record/legacy/generators/templates/v1_v2_migration_template.rb
```diff
@@ -71,6 +70,7 @@ def up
       end
       add_index :event_store_events, :id, unique: true if rails_42
       add_index :event_store_events, :created_at
+      add_index :event_store_events, :event_type
```
pawelpacana: Isn't it redundant? We don't remove existing index from legacy schema and add a new one on the same column?

pawelpacana: Oh, sqlite...

Issue #440, opened 2018-08-27
RubyEventStore::EventDuplicatedInStream does not provide valuable debugging info


paneq:
such as id of duplicated event and stream name.

pawelpacana:
First of all we do not allow events with duplicate ids, regardless of stream. We allow linking to multiple streams but also disallow linking to same event twice.

For both cases we reuse the same error, which might be fine. But communicating the issue could be different:
- in case of append/publish it doesn't matter which stream you choose, all subsequent appends of that event_id would fail
- in case of link you possibly can succeed by picking different stream name

pawelpacana:
At least communicating event_id would definitely help

PR #473, opened 2018-10-11
Communicate event_id along EventDuplicatedInStream

pawelpacana:
[#440]

### DIFF
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -12,8 +12,12 @@ module RailsEventStoreActiveRecord
 
     def append_to_stream(events, stream, expected_version)
       add_to_stream(Array(events), stream, expected_version, true) do |event|
-        build_event_record(event).save!
-        event.event_id
+        begin
+          build_event_record(event).save!
+          event.event_id
+        rescue ActiveRecord::RecordNotUnique => e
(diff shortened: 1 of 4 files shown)


pawelpacana:
@joelvh any ideas on passing `event_id` of violated constraint to https://github.com/RailsEventStore/rails_event_store/blob/c05d64db8eac6eb6778ab4627e2e7c957ea179ce/ruby_event_store-rom/lib/ruby_event_store/rom/event_repository.rb#L24?

joelvh:
Oh boy.... Will take a look.

joelvh:
@pawelpacana made an update to pass the `event_id` for application-level uniqueness exceptions. However, I'm not sure how to do this for database-level errors because the inserts are done in batches, not per-event. Do you know if there's a way to get the database to include the violation values in the error message?

pawelpacana:
> I'm not sure how to do this for database-level errors because the inserts are done in batches, not per-event. 

I believe there's no need to do that on database-level (if I understand that term correctly). We do insert in batches, but as a one transaction —https://github.com/RailsEventStore/rails_event_store/blob/master/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb#L69-L88. A single uniqueness violation is enough to roll it back.

> Do you know if there's a way to get the database to include the violation values in the error message?

Each RDBMS may implement it differently so we'd have to check all of the supported ones. I haven't checked so far if you get a list or just a one value of a row that didn't meet the constraint. I think we don't have to do that.

Issue #477, opened 2018-10-24
expected_version doesn't prevent publishing an event in the incorrect order


vaihtovirta:
Hello!

Recently I started the integration of `rails_event_store` in a couple of projects, and I run into an issue with the behavior of `expected_version` option.

## Setup (steps to reproduce)
```ruby
client = RailsEventStore::Client.new
OrderUpdated = Class.new(RailsEventStore::Event)
stream_name = "Order_1"
RailsEventStoreActiveRecord::EventInStream.where(stream: stream_name)
=> # []

event_1 = OrderUpdated.new(data: {})
client.publish(event1, stream_name: stream_name, expected_version: -1)

event_2 = OrderUpdated.new(data: {})
client.publish(event2, stream_name: stream_name, expected_version: 2)
```
## Expected behavior

- `RubyEventStore::WrongExpectedEventVersion` exception is raised since there is no event with a `position: 2` in the stream
- The stream contains the first event with `position: 0`

## Actual behavior

- `event_2` got published to the stream
- The stream contains two events with `position: 0` and  `position: 3`

```ruby
RailsEventStoreActiveRecord::EventInStream.where(stream: stream_name)
=> [#<RailsEventStoreActiveRecord::EventInStream:0x0000555950a07c70
  id: 1,
  stream: "Order_1",
  position: 0,
  event_id: "ce08208a-ae2b-4de9-932f-95cef20ac9d8",
  created_at: Wed, 24 Oct 2018 11:03:43 CEST +02:00>,
 #<RailsEventStoreActiveRecord::EventInStream:0x0000555950a04908
  id: 3,
  stream: "Order_1",
  position: 3,
  event_id: "7d136865-57de-4c01-bcbf-2f0bc199f691",
  created_at: Wed, 24 Oct 2018 11:03:46 CEST +02:00>]
```

Is it supposed behavior of `expected_version`?

That's how I understood it should work from the [docs](https://railseventstore.org/docs/expected_version/#explicit-number-integer-from-1), but maybe I'm missing something.

## Possible solution

It can be either a change in default `RailsEventStoreActiveRecord::EventRepository` or a new event repository class with the following check:

```ruby
def assert_version!
  return if expected_version == last_position

  raise RubyEventStore::WrongExpectedEventVersion
end

def last_position
  @last_position ||= RailsEventStoreActiveRecord::EventInStream.
    where(stream: stream_name).order("position DESC").first&.position || -1
end
```

[...]

PR #478, opened 2018-10-24
WrongExpectedEventVersion when expected_version greater than last stream position

pawelpacana:
See #477


paneq:
Looks good. 

The only issue is, it slows down 99,9% of writes for named streams, across projects using RES, which don't need this check. So there is that to consider.

pawelpacana:
Yep, I'd rather push it down to repository and make optional.

PR #479, opened 2018-10-26
Add limit for event

mpraglowski:


pawelpacana:
> It seems like UUID string limit... for MySQL?

Exactly that

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/limit_for_event_id_template.rb
```diff
@@ -0,0 +1,6 @@
+class LimitForEventId < ActiveRecord::Migration<%= migration_version %>
```
pawelpacana: in specs:
```
     NameError:
       uninitialized constant IndexByEventType
```

I guess you meant `LimitForEventId.new.up` like above class definition

mpraglowski: `MigrationRubyCode` used twice - pushing a fix

PR #492, opened 2018-11-20
Do not compare object instances

mpraglowski:


ruby_event_store/lib/ruby_event_store/spec/event_repository_lint.rb
```diff
@@ -1164,21 +1164,20 @@ def read_all_streams_backward(repository, start, count)
...
+      event_2 = SRecord.new(event_id: '8cee1139-4f96-483a-a175-2b947283c3c7')
+      event_3 = SRecord.new(event_id: 'd345f86d-b903-4d78-803f-38990c078d9e')
       stream_a = Stream.new('Stream A')
       stream_b = Stream.new('Stream B')
       stream_c = Stream.new('Stream C')
       repository.append_to_stream([event_1, event_2], stream_a, version_any)
       repository.append_to_stream([event_3], stream_b, version_any)
       repository.link_to_stream(event_1.event_id, stream_c, version_none)
 
-      expect(repository.streams_of(event_1.event_id)).to eq [stream_a, stream_c]
```
pawelpacana: Right part of the comparison looked ok, that is `[stream_a, stream_c]`. In RSpec, `eq` is equivalent to `==` comparison: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/equality-matchers#compare-using-eq-(==)

PR #503, opened 2018-11-28
Debugging API: get size of a stream to read

mpraglowski:
When you have read specification like:

```
scope = res.read.stream('Dummy').backward.from(:head).limit(10)
```

you could get a size of result set by using new API method: `scope.count` before you read all data from the stream by `scope.each { .... }` or `scope.to_a`.


pawelpacana:
`count` seems to be the method on `Enumerable` as well so good choice: https://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-count

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
```diff
@@ -35,8 +27,25 @@ def read(spec)
...
     end
 
+    def count(spec)
+      read_scope(spec).count
+    end
+
     private
 
+    def read_scope(spec)
+      raise RubyEventStore::ReservedInternalName if spec.stream.name.eql?(EventRepository::SERIALIZED_GLOBAL_STREAM_NAME)
```
pawelpacana: not a part of `read_scope`, please move it back to `read`

ruby_event_store-rom/lib/ruby_event_store/rom/event_repository.rb
```diff
@@ -81,6 +81,12 @@ def read(specification)
         @events.read(specification)
       end
 
+      def count(specification)
+        raise ReservedInternalName if specification.stream.name.eql?(@stream_entries.stream_entries.class::SERIALIZED_GLOBAL_STREAM_NAME)
```
pawelpacana: not a part of `count`

ruby_event_store-rom/lib/ruby_event_store/rom/repositories/events.rb
```diff
@@ -42,29 +42,15 @@ def last_stream_event(stream)
...
-          query = query.by_event_id(specification.with_ids) if specification.with_ids
-          query = query.by_event_type(specification.with_types) if specification.with_types?
+          query = read_scope(specification)
 
           if specification.batched?
             reader = ->(offset, limit) do
               query_builder(query, offset: offset, limit: limit).to_ary
             end
-            BatchEnumerator.new(specification.batch_size, limit || Float::INFINITY, reader).each
+            BatchEnumerator.new(specification.batch_size, specification.limit, reader).each
```
pawelpacana: Missed change from a different commit?

ruby_event_store-rom/lib/ruby_event_store/rom/event_repository.rb
mpraglowski: should `count` allow for reserved name used in stream specification? maybe this check should be in `Specification` class (via `SpecificationReader` where we know repository)

ruby_event_store-rom/lib/ruby_event_store/rom/repositories/events.rb
mpraglowski: Nope - just simplification.

Was:
```
limit = specification.limit if specification.limit?
and later as argument: limit || Float::INFINITY
```

but that's not needed because `specification.limit` is `Float::INFINITY` if limit has not been specified.

ruby_event_store-rom/lib/ruby_event_store/rom/event_repository.rb
pawelpacana: oh, if `count` omits `read` than it can check for reserved name (reminder that name will no longer be reserved after #312 materializes)

mpraglowski: So we should remove that check in #312 or after.

ruby_event_store-rom/lib/ruby_event_store/rom/repositories/events.rb
```diff
@@ -78,8 +64,33 @@ def read(specification)
           end
         end
 
+        def count(specification)
+          query = read_scope(specification)
+          limit = specification.limit if specification.limit?
+          query = query_builder(query, limit: limit)
```
joelvh: @mpraglowski I would say just do `query = query.take(limit) if specification.limit?` and  skip `query_builder` because it does some combining of association data and mapping that you don't need to get the `count`

PR #422, opened 2018-08-08
Check how TravisCI react for TruffleRuby in build matrix

swistak35:
Issue: #419

PR #505, opened 2018-11-28
Test with truffleruby, again.

pawelpacana:
https://github.com/RailsEventStore/rails_event_store/pull/422#issuecomment-442598046


pawelpacana:
To be revisited after https://github.com/oracle/truffleruby/issues/1503 and 
https://github.com/RailsEventStore/rails_event_store/pull/422#issuecomment-442613591 are solved.

PR #514, opened 2018-12-08
No global stream entries in event_store_events_in_streams

mpraglowski:
What this means is that we get one INSERT statement less for non-named stream appends and we store half of the entries in event_store_events (if all your event publishes are in named-streams). 

Reduced storage size and theoretical write-side improvement in one case.

- [x] adjust initial migration and adapter
- [x] add 0.39.0 to current migration
- [x] testing not only reads but also subsequent writes after migration
- [x] document suggested migration procedure (blogged — https://blog.arkency.com/how-to-migrate-large-database-tables-without-a-headache/)
- [ ] tested manually on larger postgresql dataset
- [ ] tested manually on larger mysql dataset

### DIFF
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
@@ -5,7 +5,6 @@ require 'activerecord-import'
 module RailsEventStoreActiveRecord
   class EventRepository
     POSITION_SHIFT = 1
-    SERIALIZED_GLOBAL_STREAM_NAME = "all".freeze
 
     def initialize(model_factory: WithDefaultModels.new, serializer:)
       @serializer  = serializer
@@ -21,17 +20,17 @@ module RailsEventStoreActiveRecord
         hashes    << import_hash(record, record.serialize(serializer))
         event_ids << record.event_id
       end
-      add_to_stream(event_ids, stream, expected_version, true) do
+      add_to_stream(event_ids, stream, expected_version) do
         @event_klass.import(hashes)
       end
     end
 
     def link_to_stream(event_ids, stream, expected_version)
       event_ids = Array(event_ids)
-      (event_ids - @event_klass.where(id: event_ids).pluck(:id)).each do |id|
+      (event_ids - @event_klass.where(event_id: event_ids).pluck(:event_id)).each do |id|
         raise RubyEventStore::EventNotFound.new(id)
       end
-      add_to_stream(event_ids, stream, expected_version, nil)
+      add_to_stream(event_ids, stream, expected_version)
     end
 
     def delete_stream(stream)
@@ -58,15 +57,15 @@ module RailsEventStoreActiveRecord
       hashes  = Array(records).map{|record| import_hash(record, record.serialize(serializer)) }
       for_update = records.map(&:event_id)
       start_transaction do
-        existing = @event_klass.where(id: for_update).pluck(:id)
-        (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
+        existing = @event_klass.where(event_id: for_update).pluck(:event_id, :id).to_h
+        (for_update - existing.keys).each{|id| raise RubyEventStore::EventNotFound.new(id) }
+        hashes.each { |h| h[:id] = existing.fetch(h.fetch(:event_id)) }
         @event_klass.import(hashes, on_duplicate_key_update: [:data, :metadata, :event_type])
       end
     end
 
     def streams_of(event_id)
       @stream_klass.where(event_id: event_id)
-        .where.not(stream: SERIALIZED_GLOBAL_STREAM_NAME)
         .pluck(:stream)
         .map{|name| RubyEventStore::Stream.new(name)}
     end
@@ -74,29 +73,21 @@ module RailsEventStoreActiveRecord
     private
     attr_reader :serializer
 
-    def add_to_stream(event_ids, stream, expected_version, include_global)
+    def add_to_stream(event_ids, stream, expected_version)
       last_stream_version = ->(stream_) { @stream_klass.where(stream: stream_.name).order("position DESC").first.try(:position) }
       resolved_version = expected_version.resolve_for(stream, last_stream_version)
 
       start_transaction do
         yield if block_given?
-        in_stream = event_ids.flat_map.with_index do |event_id, index|
-          position = compute_position(resolved_version, index)
-          collection = []
-          collection.unshift({
-            stream: SERIALIZED_GLOBAL_STREAM_NAME,
-            position: nil,
-            event_id: event_id,
-          }) if include_global
-          collection.unshift({
+        in_stream = event_ids.map.with_index do |event_id, index|
+          {
             stream:   stream.name,
-            position: position,
-            event_id: event_id
-          }) unless stream.global?
-          collection
+            position: compute_position(resolved_version, index),
+            event_id: event_id,
+          }
         end
         fill_ids(in_stream)
-        @stream_klass.import(in_stream)
+        @stream_klass.import(in_stream) unless stream.global?
       end
       self
     rescue ActiveRecord::RecordNotUnique => e
@@ -122,7 +113,7 @@ module RailsEventStoreActiveRecord
 
     def import_hash(record, serialized_record)
       {
-        id:         serialized_record.event_id,
+        event_id:   serialized_record.event_id,
         data:       serialized_record.data,
         metadata:   serialized_record.metadata,
         event_type: serialized_record.event_type,
--- a/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
+++ b/rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
@@ -10,7 +10,7 @@ module RailsEventStoreActiveRecord
     end
 
     def has_event?(event_id)
-      @event_klass.exists?(id: event_id)
+      @event_klass.exists?(event_id: event_id)
     end
 
     def last_stream_event(stream)
@@ -19,13 +19,12 @@ module RailsEventStoreActiveRecord
     end
 
     def read(spec)
-      raise RubyEventStore::ReservedInternalName if spec.stream.name.eql?(EventRepository::SERIALIZED_GLOBAL_STREAM_NAME)
-
       stream = read_scope(spec)
 
       if spec.batched?
         batch_reader = ->(offset_id, limit) do
-          records = offset_id.nil? ? stream.limit(limit) : stream.where(start_offset_condition(spec, offset_id)).limit(limit)
+          search_in = spec.stream.global? ? 'event_store_events' : 'event_store_events_in_streams'
+          records = offset_id.nil? ? stream.limit(limit) : stream.where(start_offset_condition(spec, offset_id, search_in)).limit(limit)
           [records.map(&method(:record)), records.last]
         end
         BatchEnumerator.new(spec.batch_size, spec.limit, batch_reader).each
@@ -41,8 +40,6 @@ module RailsEventStoreActiveRecord
     end
 
     def count(spec)
-      raise RubyEventStore::ReservedInternalName if spec.stream.name.eql?(EventRepository::SERIALIZED_GLOBAL_STREAM_NAME)
-
       read_scope(spec).count
     end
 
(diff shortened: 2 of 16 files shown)


pawelpacana:
May I ask all of you: @jakubkosinski @paneq and @mlomnicki for a review? We change schema a bit again and such changes are, well, uneasy 😓 

Things to consider:
- foreign keys when primary key is changed
- indexes, anything missing?
- order of records when adding autoincrement pkey (for mysql that was random thus rewrite from another table)

joelvh:
@pawelpacana Are there tests in place that will also make sure ROM is in sync with the same changes? ROM has a copy of migrations converted for Sequel and also copied the list of SQL driver violation error messages to be handled.

pawelpacana:
@joelvh I'm not aware of tests in place to ensure ROM copy of migrations are in sync. If there were, comparing resulting SQL might be most approachable way of comparing different schema DSLs (ROM vs. ActiveRecord::Schema).

ROM adapter passes event repository lint, which proves that with its current schema works fine. Though it might be lagging in terms of changes introduced in RailsEventStoreActiveRecord schema and the ones above.

I don't have bandwidth to keep ROM schema DSL up-to-date and more importantly guide users through changes and how to apply them in ROM flavor over (recently fast) releases. I'd ensure schemas are in-sync just before 1.0. At that time we'll be removing transition migrations from RESAR to reduce maintenance surface as well.

I imagine 0.99 release to catch-up in ROM and then from 1.0 hopefully don't change schemas significantly anymore.

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
```diff
@@ -0,0 +1,48 @@
...
+      SQL
+      drop_table :old_event_store_events
+    when "PostgreSQL"
+      rename_column :event_store_events, :id, :event_id
+      change_column_default :event_store_events, :event_id, nil
+      add_column :event_store_events, :id, :serial
+
+      execute <<-SQL
+        ALTER TABLE event_store_events DROP CONSTRAINT event_store_events_pkey;
+        ALTER TABLE event_store_events ADD PRIMARY KEY (id);
```
paneq: ` ALTER TABLE event_store_events ADD PRIMARY KEY (id)` does that indicate auto-increment?

paneq: or maybe that `add_column :event_store_events, :id, :serial` ? :)

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
```diff
@@ -0,0 +1,48 @@
...
+      add_index :event_store_events, :event_type
+
+      execute <<-SQL
+        INSERT INTO event_store_events(event_id, event_type, metadata, data, created_at)
+        SELECT id, event_type, metadata, data, created_at FROM old_event_store_events;
+      SQL
+      drop_table :old_event_store_events
+    when "PostgreSQL"
+      rename_column :event_store_events, :id, :event_id
+      change_column_default :event_store_events, :event_id, nil
```
paneq: So `event_id` must be always provided by the client.

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
```diff
@@ -0,0 +1,48 @@
...
+      add_column :event_store_events, :id, :integer
+
+      execute <<-SQL
+        UPDATE event_store_events
+        INNER JOIN event_store_events_in_streams ON (event_store_events.event_id = event_store_events_in_streams.event_id)
+        SET event_store_events.id = event_store_events_in_streams.id
+        WHERE event_store_events_in_streams.stream = 'all';
+      SQL
+
+      execute 'ALTER TABLE event_store_events DROP PRIMARY KEY, ADD PRIMARY KEY (id), MODIFY id INT AUTO_INCREMENT;'
```
paneq: Don't we need to explicitly set starting point for auto increment counter to be max of last (possibly +1) or something like that?

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/create_event_store_events_template.rb
```diff
@@ -18,25 +16,18 @@ def change
...
+        t.references :event, null: false, type: :string, limit: 36
       end
+      t.string      :event_type,  null: false
+      t.binary      :metadata
+      t.binary      :data,        null: false
+      t.datetime    :created_at,  null: false
     end
+    add_index :event_store_events, :event_id, unique: true
     add_index :event_store_events, :created_at
     add_index :event_store_events, :event_type
```
paneq: Not sure if in scope but It might be more useful to have `:event_type, :id` than just `:event_type`. Maybe...

paneq:
A couple of points (not necessarily helpful, I know):

* Any reason why we are using custom generators instead of having migrations in an engine which cna be then copied to the main rails app as described in: https://edgeguides.rubyonrails.org/engines.html#engine-setup 
* Perhaps we should not change `migrate_res_schema_v1_to_v2_template.rb` file and leave it as it is? And instead create a new file. It's hard for me to understand whether it now migrates `v1 to v2` or `v1 to v2'` (a different v2) or `v1 to v3` and whether projects which used the old version of that file reach the same point that the ones using new version of that file. Please clarify.
* > order of records when adding autoincrement pkey (for mysql that was random thus rewrite from another table)
  * and what was it for PG and Sqlite? I doubt it would be correct out of the box but maybe... Even if it is I don't think we should rely on it because it might depend on PG version etc. 
  * So maybe we should use the same approach as for mysql in all cases? I assume longer downtime is the potential issue.
* > foreign keys when primary key is changed
  * I don't think we use FKEYS so I don't think we need to do anything. If a project added them themselves then that's not our scope.
* > indexes, anything missing?
  * I don't think so.
* Generally I would be more at ease if we just used migrations to manage the schema (instead of also editing `create_event_store_events_template.rb`) . Even if we have 4 or 5 of them, that's not an issue for new projects because they would very quickly run all of them. And then we would be more sure that everyone went through all of them in the same order and with the same code inside them. Right now it is hard to track this. But perhaps that's something more of 0.99 or 1.0 concern and I doubt this can be fixed easily.
* Sadly this is not really _zero downtime friendly_ but I guess that would be much more expensive to provide. Perhaps a good option for paid support ;)

rails_event_store_active_record/spec/migrations_spec.rb
```diff
@@ -39,4 +39,52 @@
...
+        96bdacda-77dd-4d7d-973d-cbdaa5842855
+        94688199-e6b7-4180-bf8e-825b6808e6cc
+        68fab040-741e-4bc2-9cca-5b8855b0ca19
+        ab60114c-011d-4d58-ab31-7ba65d99975e
+        868cac42-3d19-4b39-84e8-cd32d65c2445
+        f5253bb5-9307-4a87-ab2d-c170df2874e6
+      ]
+      EOF
+    end
+  end
```
paneq: I like this test and its style of testing. Definitely a great idea for testing backwards compatibility with older library versions. **Blogabble as hell**. 

Consider testing not only reads but also subsequent writes after migration. Test writes are possible and don't cause exceptions and that then reads still properly return the data in the right order (old writes, new writes).

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
pawelpacana: Indeed. We never relied on it in our test cases and this default was not symmetrical with mysql/sqlite. 

Also: https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/event.rb#L15

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
pawelpacana: It does via `serial`: https://www.postgresql.org/docs/8.1/datatype.html#DATATYPE-SERIAL

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/create_event_store_events_template.rb
pawelpacana: What benefits do you see of specifying  add_index `:event_store_events, [:event_type, :id]`?

pawelpacana:
> Any reason why we are using custom generators instead of having migrations in an engine which cna be then copied to the main rails app as described in (...)

From the end-user perspective both involve running a command. Correct me if I'm wrong but I don't see a benefit of changing one into another.

pawelpacana:
> Perhaps we should not change migrate_res_schema_v1_to_v2_template.rb file and leave it as it is? And instead create a new file. It's hard for me to understand whether it now migrates v1 to v2 or v1 to v2' (a different v2) or v1 to v3 and whether projects which used the old version of that file reach the same point that the ones using new version of that file. Please clarify.

In 1.0 we intend to have a single migration file and legacy schema support to be removed by that time.

So far (in previous releases) we've taken an approach to ship single migration file for new installations and some migrations one could apply to jump into most up-to-date schema at the time they were introduced.

We've also changed v1_to_v2 a few times already so that one can jump from legacy schema into newest one in a single migration. 
The fact that it is easier to migrate from legacy schema (as in the legacy gem) to this one (without global stream) was deciding in letting it live a bit longer. It would not have been that operationally easy with dropping  integer pkey and then in next migration recreating it. Although I agree that is simpler to reason about when costs (migration time, etc.) are not involved.

I'm open to changing policy on migrations (i.e. switching to sourcing current schema from subsequent migrations that are never changed and copied to project migrations) but I'd defer it post 1.0 release.

pawelpacana:
> Sadly this is not really zero downtime friendly but I guess that would be much more expensive to provide. Perhaps a good option for paid support ;)

I doubt one could completely rely on MySQL docs stating something happens in-place (let's call it experience). I'm not using Postgres on daily basis so I might miss some tricks. I'm open for hints to make this migration better. 

The most universal approach with very short downtime window when you have gigabytes of events is to write new events twice. Have a new table with desired schema, update it on write to old schema (i.e. with trigger on create) and backfill historical data. When all in-sync -- downtime and switch the tables.
This is something that was presented in a case study at RESCON and worked well for some RES users.

pawelpacana:
>> order of records when adding autoincrement pkey (for mysql that was random thus rewrite from another table)

> and what was it for PG and Sqlite? I doubt it would be correct out of the box but maybe... Even if it is I don't think we should rely on it because it might depend on PG version etc.

For sqlite one has to always copy tables (no ALTER). You might be correct regarding postgresql, it might be related to order in which it was written on a disk. 

> So maybe we should use the same approach as for mysql in all cases? I assume longer downtime is the potential issue.

Tempting, if only for the peace of mind. In the end one can modify migration to better suite their needs.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
```diff
@@ -30,45 +28,54 @@ def read(spec)
...
+      if spec.stream.global?
+        stream = Event.order(id: order(spec))
+        stream = stream.where(event_id: spec.with_ids) if spec.with_ids?
+        stream = stream.where(event_type: spec.with_types) if spec.with_types?
+        stream = stream.limit(spec.limit) if spec.limit?
+        stream = stream.where(start_condition_in_global_stream(spec)) unless spec.head?
+        stream
+      else
+        stream = EventInStream.preload(:event).where(stream: spec.stream.name)
+        stream = stream.where(event_id: spec.with_ids) if spec.with_ids?
```
joelvh: Shared logic, can move outside of if/else

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository_reader.rb
```diff
@@ -30,45 +28,54 @@ def read(spec)
...
+        stream = stream.joins(:event).where(event_store_events: {event_type: spec.with_types}) if spec.with_types?
+        stream = stream.order(position: order(spec), id: order(spec))
+        stream = stream.limit(spec.limit) if spec.limit?
```
joelvh: Shared logic, can move outside of if/else

rails_event_store_active_record/lib/rails_event_store_active_record/generators/templates/no_global_stream_entries_template.rb
pawelpacana: I would expect so but it did not appear to be needed

joelvh:
Hey guys, wanted to check in with the plans for this and/or v1.0 release milestones and schedule? Is there somewhere I should look to see the plan?

I'm hoping to get ROM inline with AR soon.

Thanks!

mpraglowski:
> I'm hoping to get ROM inline with AR soon.

@joelvh any news here? We would like to "resurrect" this PR and go with it but I'm not sure how to process with changes in ROM.

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -50,47 +49,37 @@ def count(specification)
 
     def update_messages(messages)
       hashes = messages.map(&:to_h)
-      hashes.each{|h| h[:id] = h.delete(:event_id) }
       for_update = messages.map(&:event_id)
       start_transaction do
-        existing = Event.where(id: for_update).pluck(:id)
-        (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
+        existing = Event.where(event_id: for_update).pluck(:event_id, :id).to_h
+        (for_update - existing.keys).each{|id| raise RubyEventStore::EventNotFound.new(id) }
```
pawelpacana, line 55: id -> event_id

rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -50,47 +49,37 @@ def count(specification)
 
     def update_messages(messages)
       hashes = messages.map(&:to_h)
-      hashes.each{|h| h[:id] = h.delete(:event_id) }
       for_update = messages.map(&:event_id)
       start_transaction do
-        existing = Event.where(id: for_update).pluck(:id)
-        (for_update - existing).each{|id| raise RubyEventStore::EventNotFound.new(id) }
+        existing = Event.where(event_id: for_update).pluck(:event_id, :id).to_h
```
pawelpacana, line 54: renaming would help in understanding

ruby_event_store/spec/specification_spec.rb
```diff
@@ -38,7 +38,6 @@ module RubyEventStore
 
     specify { expect{specification.from(nil)}.to raise_error(InvalidPageStart) }
     specify { expect{specification.from('')}.to raise_error(InvalidPageStart) }
-    specify { expect{specification.from(:head)}.to raise_error(EventNotFound, /head/) }
```
pawelpacana, line 41: seems unrelated to this commit

pawelpacana:
Merging to avoid yet another rebase. Failing mutations and specs (Rails 4.2) to be dealt with on master.

PR #515, opened 2018-12-08
Drop support for legacy repository

mpraglowski:


pawelpacana:
Good one but I'm considering _not yet_. If #514 makes it, it is much easier to jump from legacy to it.

PR #524, opened 2018-12-29
JSON Support in ROM Adapter

joelvh:
Updated migrations to allow specifying JSON or JSONB for `data` and  `metadata`. This required some updating of ROM schema types and also cleaning up the test data uses for `SerializedRecord`. The test data in RSpec tests worked as arbitrary strings with YAML but not JSON (e.g. a UUID string can be deserialized by YAML but needs quotes in  JSON). Minor changes to make the data compatible with both YAML and  JSON was sufficient since tests use serialized data rather than a specific RES serializer.


joelvh:
@paneq @pawelpacana can you guys please help me understand how to address mutant? I need help here and also with  #463 to get them merged because I mostly see results that don't point to anything, such as this:

image

Thank you!

PR #527, opened 2019-01-03
Partial move from Travis CI to Circle CI

pawelpacana:


.circleci/config.rb
```diff
@@ -12,34 +12,75 @@
...
+
+dependent_job = ->(name, requires) do
+  { name => { 'requires' => Array(requires) } }
+end
 
 check_config = {
   "check_config" => {
     "docker" => [
-      { "image" => "circleci/ruby:2.6.0-node-browsers" }
+      { "image" => "pawelpacana/res:2.6.0" }
```
joelvh: @pawelpacana curious if this should be linking to your profile/account?

.travis.yml
```diff
@@ -83,42 +59,6 @@ matrix:
     env: TARGET=ruby_event_store-rom SCRIPT=test DATABASE_URL=postgres://localhost/rails_event_store_active_record?pool=5 DATA_TYPE=json
```
joelvh: @pawelpacana looks like the variations with `DATA_TYPE` arg are only specified for Ruby  2.5.3... should they  be replicated everywhere for other versions as well?

pawelpacana: I think it is sufficient to check them on "current" Ruby only. I'd now consider 2.6 to be the "current" Ruby. I haven't managed to moved those entries yet to CircleCI. I'd appreciate help with it. 

2.5 should be considered a "compatibility" Ruby, that is we don't want to introduce any syntax-breaking changes by using newer language additions. And if we do for some reason, there a polyfill or a different compatibility mechanism in place. The last compatibility Ruby is 2.3 but I'd opt for removing it around March (reaching end-of-life).

The further plan with CircleCI is to introduce:
- Postgresql workflow which checks several gems with pg in `DATABASE_URL` and `DATA_TYPE` combinations (currently only ROM supports it but it is a great idea)
- Mysql workflow to check several gems with that specific `DATABASE_URL`
- Rails workflow to check gems depending on Rails how they interact with different Rails versions 

The reasoning for such partitioning is that when you make a mysql-braking change, you'd expect such workflow to fail and I can see a benefit of looking at such "components" of a complete test suite.

Btw. I have briefly checked https://buildkite.com/RailsEventStore as alternative to speed things a bit more but it requires more work. Jumping on CircleCI with Docker makes a good intermediate step and already made builds a bit faster.

joelvh: @pawelpacana I'm interested in helping, but don't have any experience setting all these things up (configs, docker). What's the best way for me to get things running locally to  test the configs and so on?

joelvh: Also, what you  describe makes sense in terms of separating based on variations in database or framework. MySQL is actually  a bit problematic with the ROM adapter because sometimes there are deadlocks when running  the test suite, and I'm not sure why.

Issue #469, opened 2018-10-10
Introduce ProcessManager mixing similar to AggregateRoot in order to help with event sourced process managers


pawelpacana:
Possible implementation in: https://blog.arkency.com/process-managers-revisited/

pawelpacana:
Possibly related to #335

tomaszwro:
after discussing with @pawelpacana: let's make something like https://github.com/arkency/aggregates, but for process managers

good example of a process manager is in https://github.com/RailsEventStore/cqrs-es-sample-with-res

PR #538, opened 2019-01-25
Map & Reduce operations

mpraglowski:


pawelpacana:
How much do we gain over delegating {map, reduce} to each?

mpraglowski:
Syntactic sugar :) 
The trigger for this PR was an #469 issue - I've started to experiment with possible implementations and `reduce` was my first thought. Not as a solution but rather as a base for further experiments.

aggregate_root/lib/aggregate_root.rb
```diff
@@ -37,10 +37,8 @@ def apply(*events)
 
   def load(stream_name, event_store: default_event_store)
     @loaded_from_stream_name = stream_name
-    events_enumerator(event_store, stream_name).with_index do |event, index|
-      apply(event)
-      @version = index
-    end
+    event_store.read.stream(stream_name).reduce {|_, ev| apply(ev) }
```
joelvh: @mpraglowski should  this still be in batches?

ruby_event_store/lib/ruby_event_store/specification_reader.rb
```diff
@@ -23,6 +23,25 @@ def each(specification_result)
       end
     end
 
+    # @api private
+    # @private
+    def map(specification_result, &block)
+      result = []
```
joelvh: @mpraglowski style opinion - could use `each_with_object` instead of an extra variable?

aggregate_root/lib/aggregate_root.rb
mpraglowski: Since https://github.com/RailsEventStore/rails_event_store/commit/6d0d204c46039775e86fcf72c142b6ee809fadcf it is already in batches.

PR #541, opened 2019-01-29
Improved have_published output (with data and metadata structural diff)

andrzejsliwa:
WIP

- [x] Prepare prove of concept
- [ ] Discuss and clarify output of `have_published` with @pawelpacana 
- [ ] Provide spec for SuperDiffStructuralDiffer adapter
- [ ] Make mutant happy 
- [ ] Refactor


pawelpacana:
I like how that raw `#<FooEvent:0x000000010aa67010 ...>` is gone from the output

pawelpacana:
Output messages look good. I wonder if we could make the reason for non-match more specific, i.e. instead of _differing arrays_ have something like _missing FooEvent_ or _not expecting BarEvent, BazEvent and BazBarEvent_.

PR #542, opened 2019-01-31
of_type revisited

mpraglowski:


pawelpacana:
Can you please remind me what was the reason to use normalize_to_array in first place and its relation to mutant ignore?

I'm a bit on the fence with this change

paneq:
@pawelpacana the reason was a bug in mutant

pawelpacana:
> the reason was a bug in mutant
worth checking if it still remains

PR #543, opened 2019-02-01
Fix normalize_to_array: make sure append_to_stream & link_to_stream accepts Enumerable

mpraglowski:


pawelpacana:
Next step: `s/normalize_to_array/Array/`? That is https://ruby-doc.org/core-2.6/Kernel.html#method-i-Array?

pawelpacana:
I've gradually backported suggested changes, thanks! I've skipped enum-to-array tests. 

I don't believe we've been supporting this in signature of write methods in a client. At some point we've introduced a bug in `aggregate_root` related to that (ade8f84a010b9645d7dd4afbd353e2ca0ec57a3b).

Convince me that we should support it and let's have a dedicated commit introducing it then.

pawelpacana:
Closing now in a rebased state.

PR #544, opened 2019-02-01
Stop skipping tests - link to stream & :auto expected versions

mpraglowski:
No more legacy event repository. All other have support for it. All new ones should have it also.


pawelpacana:
Good question whether we actually want to do that. This test is a lint that can be used when developing your own adapter. If we expect all future adapters to support the features of removed skips that it is probably fine.

Can mongo and dynamodb adapters accept this?

mpraglowski:
> If we expect all future adapters to support the features of removed skips that it is probably fine.

Yeah - I would expect all event repositories to suport linking and :auto expected version.

PR #547, opened 2019-02-04
Restore AggregateRoot::Repository

mpraglowski:
It have been killed long time ago https://github.com/RailsEventStore/rails_event_store/commit/54b93a72372314c4a6dbdecf8b3cfc8aa645f577#diff-1d5e211055c3d9a6b2b9e4314668a3dc
Now it is going back. The promisses from `Redesign of aggregate root
module` commit still holds true:
* still no magic
* still no restrictions on how aggregate class should look like, etc

This is a breaking change. Instead of:

```
order = Order.new.load("OrderStreamHere")
...
order.store
```

you need to have code:

```
repository = AggregateRoot::Repository.new
order = repository.load(Order.new, "OrderStreamHere")
...
repository.store(order, "OrderStreamHere")
```

### DIFF
--- a/aggregate_root/lib/aggregate_root.rb
+++ b/aggregate_root/lib/aggregate_root.rb
@@ -2,6 +2,7 @@ require 'aggregate_root/version'
 require 'aggregate_root/configuration'
 require 'aggregate_root/transform'
 require 'aggregate_root/default_apply_strategy'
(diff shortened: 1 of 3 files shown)


mpraglowski:
This PR should be a discussion starting point. Especially regarding maintaining backward compatibility.

PR #554, opened 2019-02-15
Make projection to use event's type to define & run handlers

mpraglowski:


ruby_event_store/lib/ruby_event_store/projection.rb
```diff
@@ -42,7 +42,7 @@ def current_state
     end
 
     def call(event)
-      handlers.fetch(event.class).(current_state, event)
+      handlers.fetch(event.type).(current_state, event)
```
pawelpacana: 👍

ruby_event_store/lib/ruby_event_store/projection.rb
```diff
@@ -27,7 +27,7 @@ def init(handler)
 
     def when(events, handler)
       Array(events).each do |event|
-        handlers[event] = handler
+        handlers[event.to_s] = handler
```
pawelpacana: why `to_s`?

ruby_event_store/lib/ruby_event_store/projection.rb
```diff
@@ -85,7 +85,7 @@ def start_events(start)
     end
 
     def transition(state, event)
-      handlers[event.class].(state, event)
+      handlers[event.type].(state, event)
```
pawelpacana: why no `fetch` here?

ruby_event_store/lib/ruby_event_store/projection.rb
mpraglowski: Because I still want to setup projection with:

```
...
.when(SomeEventClass, -> { ... })
```

paneq: What syntax do you imagine with `Proto`

ruby_event_store/lib/ruby_event_store/projection.rb
mpraglowski: Good question - but should it be? or maybe we want to ignore events for which there is no defined handlers?

Currently when you do not specify any stream to process projection works on all events in all streams (`reduce_from_all_streams` method). Current implementation forces you to provide handlers for all domain event types :( Good luck on some production systems with reasonable number of domain events ;)

ruby_event_store/lib/ruby_event_store/projection.rb
mpraglowski: AFAIK `Proto` implements `type` - so you could use anything that matches Proto's events types.

mpraglowski: We could not use `event.class` anymore as we want to allow use self defined Event base classes.

ruby_event_store/lib/ruby_event_store/projection.rb
mpraglowski:
```
      @handlers = Hash.new { ->(_, _) {} }
```

That's why it's not fetch. I was wrong ... no need to define all kind of events - I've missed that.
`handlers[event.type]` for not existing event type will return "empty" lambda.

ruby_event_store/lib/ruby_event_store/projection.rb
pawelpacana: `to_s` actually allows Proto to be used here

PR #556, opened 2019-02-22
non-uuid positions are no more

pawelpacana:


ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -45,18 +45,11 @@ def from(start)
     # Limits the query to events before or after another event.
     # {http://railseventstore.org/docs/read/ Find out more}.
     #
-    # @param start [:end, String] id of event to start reading from.
-    #   :end can mean the end or beginning of the stream, depending on the
-    #   #direction
+    # @param start [String] id of event to start reading from.
```
pawelpacana: _to start reading from_?

PR #563, opened 2019-03-13
Ruby 2.3 EOL

pawelpacana:
https://www.ruby-lang.org/en/downloads/branches/


pawelpacana:
Failure independent of PR. I need to address it anyway from `master`.

PR #564, opened 2019-03-14
Event type remapped

mpraglowski:
Event builder instead of hash with class remapping


paneq:
Long time, no see :) 

* Don't you think that if you need custom `event_builder` then you should just substitute the whole mapper instead?
* `TypeToClass` should not be in `Mappers` namespace as it is not a mapper, it's a dependency of some mappers

pawelpacana:
> I was reviewing the encryption mapper implementation yesterday and saw its using get_const instead of allowing remapping as well

The only reason for that was it originated from a codebase where remapping is not used that way. I'd accept addition of remapping there.

joelvh:
There was another thread recently  about how to configure remappings. I think it is something that should be standard in mappers or have it in a pipeline. (What's the state of pipelines? 😄)

mpraglowski:
> What's the state of pipelines? 

Maybe this is better approach here - the pipeline could be collection of transformation classes implementing:

```
  dump: f(SerializedRecord) -> SerializedRecord
  load: f(SerializedRecord) -> SerializedRecord
```

and a single builder class where methods will be slightly different:

```
  dump: f(DomainEvent) -> SerializedRecord
  load: f(SerializedRecord) -> DomainEvent
```

and the mapping pipeline could be build by:

```
RubyeEventStore::MappingPipeline.new(
  mapper: DefaultMapper.new,
  transformations: [Encrypt.new, RemapEventType.new, DoAnythingYouWantHere.new, ...]
)
```

mpraglowski:
Maybe `SerializedRecord` won't be a good fit to transform data during transformations - as `data` & `metadata` will be already serialized to to YAML/JSON/...

We probably need another class here, similar to `SerializedRecord` but with `data` & `metadata` not transformed to serialization format yet.

So transformations should work on the new class:

```
dump: f(NamingIsHardRecord) -> NamingIsHardRecord
load: f(NamingIsHardRecord) -> NamingIsHardRecord
```

and we will need 2 mapper classes that implement:

```
dump: f(DomainEvent) -> NamingIsHardRecord
load: f(NamingIsHardRecord) -> DomainEvent
```


```
dump: f(NamingIsHardRecord) -> SerializedRecord
load: f(SerializedRecord) -> NamingIsHardRecord
```

so the complete pipeline could be initialised as:

```
RubyeEventStore::Mappers::Pipeline.new([
  DomainEventMapper.new,
  Encrypt.new,
  RemapEventType.new,
  DoAnythingYouWantHere.new,
  ...,
  SerializedRecordMapper.new,
])
```

and current implementations could be replaced by:

```
RubyeEventStore::Mappers::DefaultMapper = RubyeEventStore::Mappers::Pipeline.new([
  DomainEventMapper.new,
  SerializedRecordMapper.new,
])

RubyeEventStore::Mappers::EncryptionMapper = RubyeEventStore::Mappers::Pipeline.new([
  DomainEventMapper.new,
  Encryption.new(...),
  SerializedRecordMapper.new,
])

RubyeEventStore::Mappers::Protobuf = RubyeEventStore::Mappers::Pipeline.new([
  DomainEventProtobufMapper.new,
  SerializedRecordMapper.new,
])
```

joelvh:
Could you create a diagram of how the conversions currently happen end-to-end? I think that would be a huge help to keep the proper mental model in mind as we attempt the pipelines.

mpraglowski:
Let's talk in code ;) https://github.com/RailsEventStore/rails_event_store/pull/572

Issue #567, opened 2019-03-29
Unreliable decryption with OpenSSL


pawelpacana:
In one of the `EncryptionMapper` [tests](https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/spec/mappers/encryption_mapper_spec.rb#L132-L152) we verify that decrypting a cryptogram with incorrect key (not the key used to encrypt it) results in ForgottenData. 

In order to achieve that, we [rely on the fact](https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/mappers/encryption_mapper.rb#L213) that decrypting with incorrect key would result in an `OpenSSL::Cipher::CipherError` error being raised. There are some cases however when decrypt happily returns garbage data instead:

```
irb(#<RubyEventStore::Mappers::EncryptionMapper:0x00000001140960e0>):014:0> cryptogram
=> "\x16)/\xF0\xF2\xE8W\xC8\xA49.V\xB6\x8C\xEF\xA5\xA7\xE6(\ro\xD8\x84N\x15\xD0\"\xBFb\xF3\x96\xF4"
irb(#<RubyEventStore::Mappers::EncryptionMapper:0x00000001140960e0>):015:0> encryption_key.decrypt(cryptogram, meta.fetch(:iv))
=> "BL\xC7\xCCZ<\u007FAg\x8DRLC4\x88\xD8ڼ\x9A\x8B\u0014i\xF1\u0012\xC1\u0019\xA2X\xEF\xF8\xD9"
```

Is it something expected within OpenSSL or are we using it wrong?

fidel:
> Yes, that's correct, decrypting broken data or data with the wrong key may result in the wrong plaintext without producing an error. Actually, the chance is slightly more than 1/256.
More: https://crypto.stackexchange.com/a/63418

fidel:
Discussion suggest using gcm. Apperently, my openssl version doesn't support AEAD for `aes-256-gcm` algo, so it's hard to check if it would solve our issue.

fidel:
Ok, so after deeper research, the only viable solution atm is to sign the encrypted value with HMAC and then verify it while decrypting.

We should go with Encrypt-then-MAC approach, probably. Good read: https://crypto.stackexchange.com/a/205

I started working on a PoC, but will probably try to address this throughout the weekend or on Monday.

pawelpacana:
[...]

PR #569, opened 2019-04-06
Make projections faster by only reading Events that have handlers

Gordin:


mpraglowski:
I have mixed feelings about that PR. It looks like a nice improvement but how it will work if someone uses the event type remapping feature we have implemented in `DefaultMapper`? 
The remapping of event type (event's class) is performed while reading from event repository by mapper. I have no problem when someone explicte uses the `of_type` method because it's his/hers responsibility to provide correct types... but here it will be easy to miss the use case when some events won't be ready due to remapping feature.
I.e. class `A` given in `when` method, but it's stored as `X` in `event_type` and is only remapped to `A` while reading from event repository. Now it will work correctly because `X` will be remapped to `A` before it is processed by projection. With filtering by event type events of type `X` won't be read and won't be processed.

Gordin:
I didn't know about the remapping feature, I'll look into it ^^. Maybe I can just also include the events that will be mapped.

I was gonna ask if you have any idea why the mutation testing for this fails fails in the EncryptionMapper, but I'm guessing it could have something to do with the remapping as well. I'll try to fix this as well.

pawelpacana:
> I was gonna ask if you have any idea why the mutation testing for this fails fails in the EncryptionMapper

Sorry for the trouble with EncryptionMapper, the reason for this is https://github.com/RailsEventStore/rails_event_store/issues/567

pawelpacana:
@Gordin that issue with EncryptionMapper should be gone if you rebase with master

pawelpacana:
Thank you! I've rebased your branch in #709 and this is now merged. 

Remapping is yet to be done, but we did not support it in Projection previously either (#710).

PR #572, opened 2019-04-12
Mappers pipeline

mpraglowski:


mpraglowski:
This is breaking change https://github.com/RailsEventStore/rails_event_store/pull/572/commits/7457c9f8ee60dd223608fcf2824b2672cd04e0c4 - currently we only symbolize metadata - I could not recall what was the reason for that.

ruby_event_store/lib/ruby_event_store/mappers/pipeline.rb
```diff
@@ -0,0 +1,24 @@
+module RubyEventStore
+  module Mappers
+    class Pipeline
+      def initialize(transformations)
```
joelvh: @mpraglowski `initialize(*transformation)`?

joelvh:
This is great @mpraglowski!

I tend to forget why `SerializedRecord` requires each field to be a string. Why  does it care about the type?

pawelpacana:
> I tend to forget why SerializedRecord requires each field to be a string. Why does it care about the type?

Database does not care about non-strings. Requiring strings already at SerializedRecord level enforces  explicit (as oppposed to accidental) conversion to those strings by mapper.

mpraglowski:
> You've chosen Hash for to pass item around transformations. I'd prefer an object with API surface smaller than Hash, delegating to it (maybe simillar to Metadata).

TransformationItem it is.

ruby_event_store/lib/ruby_event_store/mappers/pipeline.rb
mpraglowski: Now with explicit edge transformations it could be defined as:

```ruby
def initialize(*transformations, to_domain_event: DomainEventMapper.new, to_serialized_record: SerializedRecordMapper.new)
  @transformations = ...
end
```

And it will works fine when no edge mappers will be given like:

```ruby
Pipeline.new(Transformation1.new, Transformation2.new)
```

It will work event when we add edge mappers at the end of arguments:

```ruby
Pipeline.new(Transformation1.new, Transformation2.new, to_domain_event: SomeOtherStrategy.new)
```

by it fails with `SyntaxError ((irb):40: syntax error, unexpected ',', expecting =>)` (at least for my tests) when we have definition like:

```ruby
Pipeline.new(to_domain_event: SomeOtherStrategy.new, Transformation1.new, Transformation2.new)
```

Maybe I'm missing something.

pawelpacana:
> > I tend to forget why SerializedRecord requires each field to be a string. Why does it care about the type?
> 
> Database does not care about non-strings. Requiring strings already at SerializedRecord level enforces explicit (as oppposed to accidental) conversion to those strings by mapper.

That however reminded me this is problem for `NullMapper`. After all not all use cases require strings and that should be up to serializer that processes `data` and `metadata`.

joelvh:
Yeah, with the ROM adapter, it uses Sequel as the driver and for a JSON column, it comes back as a Hash, which I need to serialize to a string to get to work with SerializedRecord. I believe we briefly chatted about it at the time I added the JSON column capability.

pawelpacana:
Is the pipeline always symmetrical or is there a chance to have different chain of transformations depending on direction?

mpraglowski:
Currently is always symmetrical however you could implement your transformations to be one way only and bypass transformation item for the other way.

ruby_event_store/lib/ruby_event_store/mappers/event_class_remapper.rb
```diff
@@ -0,0 +1,20 @@
...
+      def initialize(class_map)
+        @class_map = class_map
+      end
+
+      def dump(item)
+        item
+      end
+
+      def load(item)
+        item.merge(event_type: class_map.fetch(item.fetch(:event_type)) { item.fetch(:event_type) })
```
pawelpacana: Following might be simpler than requiring `fetch` on class map that we don't make use of really:
```
class_map[item.fetch(:event_type)] || item.fetch(:event_type)
```

ruby_event_store/lib/ruby_event_store/mappers/pipeline.rb
pawelpacana: Keyword arguments go after positional ones.

ruby_event_store/lib/ruby_event_store/mappers/default.rb
```diff
@@ -3,6 +3,8 @@
 module RubyEventStore
   module Mappers
     class Default
+      include PipelineMapper
```
pawelpacana: Maybe that's already in next commits: doesn't feel right to include PipelineMapper and depend on a presence of an ivar.

Either `include PipelineMapper.new(...)` with all pipeline stuff or `Default = PipelineMapper.new(...)`.

mpraglowski: Would be great but it will break backward compatibility - we need some arguments given in mapper initializer to build a pipeline.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -81,7 +101,11 @@ def initialize(events_class_remapping: {})
         require_optional_dependency
         @pipeline = Pipeline.new([
           DomainEventProtoMapper.new,
+          ProtoMapper.new,
           EventClassRemapper.new(events_class_remapping),
+          SymbolizeMetadataKeys.new,
+          StringifyMetadataKeys.new,
```
pawelpacana: why both symbolize and stringify?

mpraglowski: to keep symmetric behaviour

pawelpacana: sounds like a weak argument for different pipelines ;)

mpraglowski: This is because edge mappers have different needs - `to_domain_event` mapper needs symbolized metadata keys, but `to_serialized_record` needs here stringified keys. Using both mappers here in that order solved problem.

pawelpacana: Exactly, with separate pipelines (one for dump and one for load) Symbolize would be in one but not the other. Same with Stringify.

mpraglowski: Yeah... but that's outdated code ;) This is no more in final version.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
```diff
@@ -16,7 +16,6 @@ class Protobuf
       include PipelineMapper
 
       def initialize(events_class_remapping: {})
-        require_optional_dependency
```
pawelpacana: Not sure about this one, I'd rather like to see fail because of missing dependency before first use of mapper.

ruby_event_store/lib/ruby_event_store/mappers/encryption_mapper.rb
```diff
@@ -6,7 +6,7 @@ class EncryptionMapper
       def initialize(key_repository, serializer: YAML, forgotten_data: ForgottenData.new)
         @pipeline = Pipeline.new(
           transformations: [
-            Encryption.new(key_repository, forgotten_data: forgotten_data),
+            Encryption.new(key_repository, serializer: serializer, forgotten_data: forgotten_data),
```
pawelpacana: > Encryption transformation should use the same serializer as whole pipeline

Not really but we've discussed that already :)

ruby_event_store/lib/ruby_event_store/mappers/default.rb
pawelpacana: OK, no breaking compatibility. I'd however revert that `include PipelineMapper` module. Having mapper methods explicitly delegate to pipeline is fine composition. I know this basically happens in the included module but it is not instantly visible. 

What do we gain with that extraction?

ruby_event_store/lib/ruby_event_store/mappers/domain_event_mapper.rb
```diff
@@ -0,0 +1,22 @@
+module RubyEventStore
+  module Mappers
+    class DomainEventMapper
```
pawelpacana: Btw. shouldn't we call this transformations? It does not implement Mapper API after all and that might be misleading when we already have precise meaning for mapper in RES.

ruby_event_store/lib/ruby_event_store/mappers/encryption_mapper.rb
pawelpacana: On a second thought maybe that's good call to use the same. If you go with JSON and schema, you'll be covered for any loss in the transformation.

ruby_event_store/lib/ruby_event_store/mappers/domain_event_mapper.rb
mpraglowski: Yeah - naming is hard. Separate commit with renames shall be pushed soon.

ruby_event_store/lib/ruby_event_store/mappers/protobuf.rb
mpraglowski: Moved to https://github.com/RailsEventStore/rails_event_store/blob/1767bd70da4db24a39788833526c7d27decde79d/lib/ruby_event_store/mappers/protobuf_nested_struct_metadata_mapper.rb#L20

But that is still called in initalizer of `ProtobufNestedStructMetadataMapper` which is called in ctor of `Mappers::Protobuf`, and if you use that mapper in your client you initalize it when creating new client instance. For me that's before first use ;)

Issue #446, opened 2018-09-07
support changing messages on Sqlite


paneq:
* zdennis/activerecord-import#540

joelvh:
@paneq update the dependency version of AR Import to v0.26.0+ and close?

PR #574, opened 2019-04-18
Updating messages with sqlite

pawelpacana:
- [x] push new docker images for CircleCI with newer sqlite3

Closes: #446


rails_event_store_active_record/spec/event_repository_spec.rb
```diff
@@ -20,7 +20,7 @@ module RailsEventStoreActiveRecord
     let(:test_race_conditions_auto)  { !ENV['DATABASE_URL'].include?("sqlite") }
     let(:test_race_conditions_any)   { !ENV['DATABASE_URL'].include?("sqlite") }
     let(:test_binary)                { true }
-    let(:test_change)                { !ENV['DATABASE_URL'].include?("sqlite") }
+    let(:test_change)                { true }
```
joelvh: @pawelpacana I  believe ROM also has this built in - not sure if this test is enabled on ROM

pawelpacana: Right, I'll check this 👍

PR #575, opened 2019-04-23
NullMapper in form of Default + null serializer

### LINKED ISSUE #552: The problem with NullMapper
With [NullMapper](https://github.com/RailsEventStore/rails_event_store/issues/321) combined with `InMemoryRepository` one can have a decent and fast setup for testing. The speed comes from the fact that for in-memory storage no serialization is needed.

In fact `NullMapper` does not convert given event object to `SerializedRecord` instance. Yet the repositories assume such. It mostly worked given event instance responds to `event_id`, `data` and `metadata`.

With the introduction of `of_type` to `Specification` such implementation of `NullMapper` no longer fits assumptions — responding to `event_type` is required. What is more `Event` and `SerializedRecord` have differing `to_h` and `==` implementations making `NullMapper` behavior even more questionable.

Making `NullMapper` use `SerializedRecord` is currently not possible due to [requirement](https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/serialized_record.rb#L5) of `data` and `metadata` to be strings. But an object imitating `SerializedRecord` is still an option. I'm not sure how then `NullMapper` will fit into the scenario where one primarily uses `Protobuf` mapper (and possibly other sophisticated mappers).

Will introducing #400 change anything here? Is it that we only want a behavior where `data` and `metadata` remain unchanged?

pawelpacana:
Closes #552

### DIFF
--- a/rails_event_store-rspec/lib/rails_event_store/rspec/have_applied.rb
+++ b/rails_event_store-rspec/lib/rails_event_store/rspec/have_applied.rb
@@ -5,7 +5,7 @@ module RailsEventStore
         @expected  = [mandatory_expected, *optional_expected]
         @matcher   = ::RSpec::Matchers::BuiltIn::Include.new(*expected)
         @differ    = differ
-        @phraser    = phraser
+        @phraser   = phraser
       end
 
       def matches?(aggregate_root)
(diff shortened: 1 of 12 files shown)


pawelpacana:
Aside: in matchers for testing rspec-matchers we rely on the fact that NullMapper returns same object instance. Probably something we cannot depend on when taking route presented in this PR.

PR #576, opened 2019-04-29
Possible fix for in Migrating_Messages.html.md

Phenixer:
I think that the "Change event type" section would not work since it's overwriting with the same array as before. We would need to store the map in a variable, if I am not mistaken.


pawelpacana:
Good catch, thanks!

Issue #578, opened 2019-05-01
DefaultApplyStrategy for AggregateRoot fails when using protobuf


obahareth:
When you use Protobuf for your events along with AggregateRoot, the AggregateRoot default apply strategy tries to call a handler called `apply_proto` because it isn't reading the event class ([It's reading `event.class`](https://github.com/RailsEventStore/rails_event_store/blob/59f1fee25f7e8f32757f7079d49876bab385b6ff/aggregate_root/lib/aggregate_root/default_apply_strategy.rb#L22) instead of `event.data.class` in the case of protobuf). It seems like we need a different apply strategy for protobuf.

The examples below all work if you use the default events and YAML, but when switching to Protobuf they don't.

Sample output

```
AggregateRoot::MissingHandler (Missing handler method apply_proto on aggregate MyAggregate)
```

Sample aggregate

```rb
class MyAggregate
  include AggregateRoot

  def initialize
    @amount = 0
  end

  def add_amount(amount)
    @amount += amount
  end

  on BalanceAdded do |event|
    add_amount(event.data.fetch(:amount, 0))
  end


  private

  attr_reader :amount
end
```

paneq:
@obahareth Nice catch. This might fix it #579 , could you give it a try? It's just PoC.

obahareth:
#579 does fix it, thanks!

pawelpacana:
Continued in #582

pawelpacana:
Fix has been released in v0.40.0: https://github.com/RailsEventStore/rails_event_store/releases/tag/v0.40.0

PR #579, opened 2019-05-01
AggregateRoot + Protobuf

paneq:
Maybe fixes: #578 :)


obahareth:
@paneq Thank you so much for the quick response and this PR, it works!

:shipit:

pawelpacana:
Proto test in aggregate_root would be a good addition to this

pawelpacana:
@paneq does protobuf require namespacing with dot, that is for example: `res_testing.SpanishInquisition`? That one would break in current AR implementation as we split namespaces by `::`.

pawelpacana:
See #582

PR #580, opened 2019-05-01
Simple benchmark from RES instrumentation

pawelpacana:
Given a block of instrumented RES code, print aggregated time of each RES
metric with percentage of total block execution time.

Example:

```
  metric                  ms      %
  ─────────────────────────────────
  read                  0.02   0.13
  serialize            10.68  66.22
  append_to_stream      0.87   5.41

  total                16.13 100.00
```

Right now, it expects (hardcoded) `ActiveSupport::Notifications` but it
doesn't have to be this way.

Testing unfortunately relies on magic number of mocked Time.now
execution. Something that may unexpectedly change in the future.

### DIFF
--- a/contrib/measure/Gemfile
+++ b/contrib/measure/Gemfile
@@ -1,0 +1,15 @@
+source "https://rubygems.org"
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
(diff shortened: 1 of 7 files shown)


pawelpacana:
No mutation testing yet, as this is contrib. Also naming might change in the near future. Besides that, wdyt?

PR #582, opened 2019-05-03
Proto with AggregateRoot, continued.

pawelpacana:


aggregate_root/spec/aggregate_root_spec.rb
```diff
@@ -229,8 +229,8 @@
       order.apply(Orders::Events::OrderExpired.new)
       expect(order.status).to eq(:expired)
 
-      expect(order.private_methods).to include(:"on_Orders::Events::OrderCreated")
-      expect(order.private_methods).to include(:"on_Orders::Events::OrderExpired")
```
paneq: potential collision when same class name but different namespace ie `Orders::Created`, `Customers::Created` (not very domain-specific example but just to illustrate the problem).

paneq:
## Alternative approach

```
class EventType < String
  def unnamespaced
    split(%r{::|\.}).last
  end

  def normalized
    Transform.to_snake_case(unnamespaced)
  end
end
```

```
class Proto
  def type
    EventType.new(data.class.descriptor.name)
  end
end
```

```
class RES::Event
  def type
    EventType.new(self.class.name)
  end
end
```

Ah no, I just realized this would not help much as `on` gets a class as an argument... Scratch that...

## Thoughts

Ok, here is a problem that I see. Asymmetry. `on` method uses `event_klass.name` to make decisions. `handler_name` method uses `event.type`. One gets a class of an event (or event data in case of proto), the other  gets an instance of an event. This is not a symmetrical interface. 

* Maybe proto should use: `on 'res_testing.OrderPaid'` (passing type instead of a class) ?
* Maybe aggregate root should consult the mapper to understand how to map the input of the `on` method to an event type which might vary per mapper implementation `mapper.map_klass_to_type(klass)` ?

The current solution obviously works but it doesn't feel right to me. But as you can see, I am not clear how to improve it easily.

aggregate_root/spec/aggregate_root_spec.rb
pawelpacana: Possible. Should that happen in aggregate? Should `on` behave differently than `non-on` in the same default strategy?

pawelpacana:
Proponents of `on ClassName` favored it because some IDEs allow better code navigation as opposed to using strings.

pawelpacana:
> Maybe proto should use: on 'res_testing.OrderPaid' (passing type instead of a class) ?

As long as `Event#type` returns string, this is a sensible approach to me. There might be a room to even make passing class backwards-compatible.

pawelpacana:
Alternative approach: https://github.com/RailsEventStore/rails_event_store/pull/588

pawelpacana:
Going with alternative as it does break compatibility nor changes behaviour. Lets revisit when a better solution becomes clearer.

PR #584, opened 2019-05-06
Experiments with AggregateRoot API

pawelpacana:


pawelpacana:
Failed on #585, merging.

PR #591, opened 2019-05-17
AggregateRoot's repository instrumentation

mpraglowski:
Instruments `load` & `store` methods. Tracks aggregate class, stream and
for storing also current aggregate version & number of events stored.

### DIFF
--- a/aggregate_root/lib/aggregate_root/instrumented_repository.rb
+++ b/aggregate_root/lib/aggregate_root/instrumented_repository.rb
@@ -24,7 +24,8 @@ module AggregateRoot
     end
 
     def with_aggregate(aggregate, stream_name, &block)
-      repository.with_aggregate(aggregate, stream_name, &block)
+      block.call(load(aggregate, stream_name))
+      store(aggregate, stream_name)
     end
 
     private
--- a/aggregate_root/lib/aggregate_root/instrumented_repository.rb
+++ b/aggregate_root/lib/aggregate_root/instrumented_repository.rb
@@ -1,0 +1,33 @@
+module AggregateRoot
+  class InstrumentedRepository
+    def initialize(repository, instrumentation)
+      @repository = repository
+      @instrumentation = instrumentation
+    end
+
+    def load(aggregate, stream_name)
+      instrumentation.instrument("load.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
+                                 stream_name: stream_name) do
+        repository.load(aggregate, stream_name)
+      end
+    end
+
+    def store(aggregate, stream_name)
+      instrumentation.instrument("store.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
+                                 aggregate_version: aggregate.version,
+                                 stored_events: aggregate.unpublished_events.size,
+                                 stream_name: stream_name) do
+        repository.store(aggregate, stream_name)
+      end
+    end
+
+    def with_aggregate(aggregate, stream_name, &block)
+      repository.with_aggregate(aggregate, stream_name, &block)
+    end
+
+    private
+    attr_reader :instrumentation, :repository
+  end
+end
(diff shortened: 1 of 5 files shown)


aggregate_root/lib/aggregate_root/instrumented_repository.rb
```diff
@@ -0,0 +1,33 @@
...
+                                 stream_name: stream_name) do
+        repository.load(aggregate, stream_name)
+      end
+    end
+
+    def store(aggregate, stream_name)
+      instrumentation.instrument("store.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
+                                 aggregate_version: aggregate.version,
+                                 stored_events: aggregate.unpublished_events.size,
```
pawelpacana: Perhaps we should pass just `aggregate.unpublished_events` here. That would be in symmetry with `append_to_stream.repository.rails_event_store`.

aggregate_root/lib/aggregate_root/instrumented_repository.rb
```diff
@@ -0,0 +1,33 @@
...
+                                 stream_name: stream_name) do
```
pawelpacana: s/stream_name/stream, again following `append_to_stream.repository.rails_event_store`

aggregate_root/lib/aggregate_root/instrumented_repository.rb
```diff
@@ -0,0 +1,33 @@
...
+                                 aggregate_class: aggregate.class,
+                                 stream_name: stream_name) do
+        repository.load(aggregate, stream_name)
+      end
+    end
+
+    def store(aggregate, stream_name)
+      instrumentation.instrument("store.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
+                                 aggregate_version: aggregate.version,
```
pawelpacana: s/aggregate_version/version/

aggregate_root/lib/aggregate_root/instrumented_repository.rb
```diff
@@ -0,0 +1,33 @@
...
+      instrumentation.instrument("load.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
+                                 stream_name: stream_name) do
+        repository.load(aggregate, stream_name)
+      end
+    end
+
+    def store(aggregate, stream_name)
+      instrumentation.instrument("store.repository.aggregate_root",
+                                 aggregate_class: aggregate.class,
```
pawelpacana: I think passing the instance, labeled as aggregate would be fine (thus s/aggregate_class/aggregate/)

pawelpacana: Instrumentation is useful in logging context, being able to track particular instance is helping.

aggregate_root/spec/instrumented_repostory_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'spec_helper'
+require 'active_support/notifications'
+
+module AggregateRoot
+  RSpec.describe InstrumentedRepository do
+
+    describe "#load" do
+      specify "wraps around original implementation" do
+        some_repository = spy
```
pawelpacana: Is there a way for spy to check if it still follows an API that is mocked? I'm always wary of such tests. I briefly remember https://github.com/psyho/bogus implementing this check.

pawelpacana: https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/verifying-doubles/using-an-instance-double

aggregate_root/spec/instrumented_repostory_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        subscribe_to("store.repository.aggregate_root") do |notification_calls|
+          aggregate = Order.new
+          aggregate.create
+          aggregate.expire
+
+          instrumented_repository.store(aggregate, 'SomeStream')
+
+          expect(notification_calls).to eq([{
+            aggregate_class: Order,
+            aggregate_version: -1,
```
pawelpacana: Interesting question, are we more interested in version before or after storing events? Also what happens in context of instrumentation if we cannot store events?

mpraglowski: Usually repository gets a new instance of aggregate, so on load method call version does not matter. The version in store instrumentation is an aggregate version **before** store.

Issue #457, opened 2018-09-25
Debugging API - filter and order by event publishing date


swistak35:
Similar to #456, filtering and/or ordering by date of publishing an event is very useful for debugging.

pawelpacana:
Done #764

PR #593, opened 2019-05-17
Adds time filtering of events

masylum:
# Adds filtering of events by time

👋 Hello there!

This is my first PR to the project.
We have this use case where we want to filter events by their creation date. We implemented our own version of the event browser explorer but adapted to our domain. One of the requirements we have is to be able to filter out events by date, but right now there is no easy way to do so with rails event store.

Let me know what you think!

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/specification.rb
+++ b/ruby_event_store/lib/ruby_event_store/specification.rb
@@ -33,12 +33,34 @@ module RubyEventStore
     # Limits the query to events before or after another event.
     # {http://railseventstore.org/docs/read/ Find out more}.
     #
-    # @param start [String] id of event to start reading from.
+    # @param stop [String] id of event to start reading from.
     # @return [Specification]
     def to(stop)
       raise InvalidPageStop if stop.nil? || stop.empty?
       raise EventNotFound.new(stop) unless reader.has_event?(stop)
       Specification.new(reader, result.dup { |r| r.stop = stop })
+    end
+
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
+    # @param equals [Boolean] whether you want to include events on that date
+    # @return [Specification]
+    def older_than(date, equals = false)
+      raise InvalidPageStart if date.nil? || date.empty?
+      Specification.new(reader, result.dup { |r| r.older_than = [date, equals] })
+    end
+
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
+    # @param equals [Boolean] whether you want to include events on that date
+    # @return [Specification]
+    def newer_than(date, equals = false)
+      raise InvalidPageStop if date.nil? || date.empty?
+      Specification.new(reader, result.dup { |r| r.newer_than = [date, equals] })
     end
 
     # Sets the order of reading events to ascending (forward from the start).
--- a/ruby_event_store/lib/ruby_event_store/specification_result.rb
+++ b/ruby_event_store/lib/ruby_event_store/specification_result.rb
@@ -3,14 +3,16 @@ module RubyEventStore
     def initialize(direction: :forward,
                    start: nil,
                    stop: nil,
+                   older_than: nil,
+                   newer_than: nil,
                    count: nil,
                    stream: Stream.new(GLOBAL_STREAM),
                    read_as: :all,
                    batch_size: Specification::DEFAULT_BATCH_SIZE,
                    with_ids: nil,
                    with_types: nil)
-      @attributes = Struct.new(:direction, :start, :stop, :count, :stream, :read_as, :batch_size, :with_ids, :with_types)
-        .new(direction, start, stop, count, stream, read_as, batch_size, with_ids, with_types)
+      @attributes = Struct.new(:direction, :start, :stop, :older_than, :newer_than, :count, :stream, :read_as, :batch_size, :with_ids, :with_types)
+        .new(direction, start, stop, older_than, newer_than, count, stream, read_as, batch_size, with_ids, with_types)
       freeze
     end
 
@@ -52,6 +54,22 @@ module RubyEventStore
     # @return [String|Symbol]
     def stop
       attributes.stop
+    end
+
+    # Ending date.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @return [Array<Date, Boolean>]
+    def older_than
+      attributes.older_than
+    end
+
+    # Starting date.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @return [Array<Date, Boolean>]
+    def newer_than
+      attributes.newer_than
     end
 
     # Read direction. True is reading forward
@@ -177,6 +195,8 @@ module RubyEventStore
     # * direction
     # * start
     # * stop
+    # * older_than,
+    # * newer_than
     # * count
     # * stream
     # * read_as
@@ -191,6 +211,8 @@ module RubyEventStore
         get_direction,
         start,
         stop,
+        older_than.join(''),
+        newer_than.join(''),
         limit,
         stream,
         attributes.read_as,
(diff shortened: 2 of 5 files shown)


pawelpacana:
Hello :wave:

Thank you for the contribution. There's an open issue (#457) very briefly suggesting what you implemented. I'll have a look at this PR soon :)

masylum:
Awesome! Let me know if you think the API/approach matches the library philosophy and I will work to make those tests pass (I just replicated the `from/to` tests to start with)

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -33,14 +33,36 @@ def from(start)
...
     end
 
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
+    # @param equals [Boolean] whether you want to include events on that date
+    # @return [Specification]
+    def older_than(date, equals = false)
+      raise InvalidPageStart if date.nil? || date.empty?
```
mpraglowski: I think the error raised here might be misleading.

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -33,14 +33,36 @@ def from(start)
...
+    end
+
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
+    # @param equals [Boolean] whether you want to include events on that date
+    # @return [Specification]
+    def newer_than(date, equals = false)
+      raise InvalidPageStop if date.nil? || date.empty?
```
mpraglowski: The same here. Misleading error. Probably we need a new one or just use `ArgumentErrror`

ruby_event_store/lib/ruby_event_store/specification_result.rb
```diff
@@ -191,6 +211,8 @@ def hash
         get_direction,
         start,
         stop,
+        older_than.join(''),
```
mpraglowski: Will fail if no dates set. Beware of nils.

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -33,14 +33,36 @@ def from(start)
...
     def to(stop)
       raise InvalidPageStop if stop.nil? || stop.empty?
       raise EventNotFound.new(stop) unless reader.has_event?(stop)
       Specification.new(reader, result.dup { |r| r.stop = stop })
     end
 
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
```
mpraglowski: Why `Date`? Why not `Time` ? It will be a common scenario to filter events only from a very short amount of time (minutes or hours) - not always full days are the range we want to get while debugging some problems.

ruby_event_store/lib/ruby_event_store/specification.rb
```diff
@@ -33,14 +33,36 @@ def from(start)
...
+      Specification.new(reader, result.dup { |r| r.older_than = [date, equals] })
+    end
+
+    # Limits the query to events before or after another event.
+    # {http://railseventstore.org/docs/read/ Find out more}.
+    #
+    # @param date [Date]
+    # @param equals [Boolean] whether you want to include events on that date
+    # @return [Specification]
+    def newer_than(date, equals = false)
```
mpraglowski: `newer_than(date, equals = false)` - I don't like that 2nd argument - it adds additional complexity here, we need to store tuple `<Date, Boolean>` instead of just `Date` (or maybe `Time` as mentioned above), then `specification.newer_than` returns that tuple instead of just time. Let's decide on some way of selecting events (>& < as method names suggests) and let developers to pass appropriate time range.

masylum: that's what I wanted input on. Not very happy with that API. Another option was to expose 4 methods to map the `<`, `<=`, `>` and `>=` operators. Also not sure about the method name. Maybe we could go with something like arel: `lt`, `lteq`, `gt`, `gteq`?

ruby_event_store/lib/ruby_event_store/specification.rb
masylum: Good point. I will change. This is the kind of feedback I wanted before diving deeper.

ruby_event_store/lib/ruby_event_store/specification.rb
masylum: yep, I just wanted to prototype an initial idea so I did it by copy/pasting `from/to` methods. Will translate this exceptions to meaningful ones.

masylum:
Let's discuss some of the big-picture issues with this PR:

> 1st:
> RES sets timestamp of domain event at the moment of its publication (here: /ruby_event_store/lib/ruby_event_store/client.rb@master#L294  ). But that timestamp stored in metadata is not the same timestamp as set in event's table created_at attribute :(
> That's why using created_at for filtering might be misleading. In some cases metadata's timestamp might be set by developer some time before domain event publication and then that value will be used - which might be quite different than created_at.

Very good point. There is this blog post that talks about the *transaction* and *valid* time difference in detail:
https://juxt.pro/blog/posts/value-of-bitemporality.html

The only problem is that the current API for storing *valid time* in RES is through metadata. That means serialization, which makes it very hard to be queried against.

By using `created_at` we can only query against the *transaction time* (when did we persist that event) which can be misleading indeed. So the only way users could query the *valid time* (when did the event happened?) without loading them all in memory would be to "lift" it to the events table instead of being serialized. But I guess that's outside of the scope of this proposal.

> There are 2 created_at stored:
> 1st in event_store_events table - and that is a timestamp set when domain event is stored,
> 2nd in event_store_events_in_streams table - and that is a timestamp set when domain event is stored OR linked to some stream.
> I'm not sure how the filtering should work - which timestamp will be valid to use here.
> This might be affected by #514 - then we either read all streams (and use created_at from event's table) or we read from specific stream (and then use created_at from stream's table) - but still that must be documented and explained to avoid confusion.

I'm not sure what use case would want to know when a certain event has been linked to a stream. I think "when did the event become part of a given stream" has very different semantics that "when was the event persisted"" or "when did the event happen"" as discussed in the prior point. I think those are 3 different questions.

Summarizing, the way I see it:

  - *When was the event persisted*: We can query this easily and there is value adding this into ES. Confusion could be minimized if documented properly.
  - *When did the event happen?*: This would require changes on the API. Not worth it yet, but I would definitely consider it in the future!
  - *When did the event become part of a given stream*. There may be some value on answering the question, but I don't think it matters that much yet.

ruby_event_store/lib/ruby_event_store/specification.rb
mpraglowski: I would keep the naming and let developer to provide valid arguments for `>` & `<` operators.

pawelpacana: WIth Ruby 2.7 endless ranges it could be interesting to consider `between(Range)` API.

mpraglowski: Yes, I though about it but we want to have open ranges and we still support Ruby 2.4 AFAIR :)

mpraglowski:
> When was the event persisted: We can query this easily and there is value adding this into ES. Confusion could be minimized if documented properly.

That simplest solution possible now. If documented properly might be good enough...

> When did the event happen?: This would require changes on the API. Not worth it yet, but I would definitely consider it in the future!

... however if we want (and I think we should) have event timestamp (_valid time_) outside of the metadata then first way of implementing this might be confusing and changing it would introduce a breaking change. Maybe we should go for that now - outside of this PR but then this PR should use _valid time_ to filter events.

> When did the event become part of a given stream. There may be some value on answering the question, but I don't think it matters that much yet.

Again here use of _valid time_ instead of _transaction time_ will help here - the _valid time_ is independent on stream as event is persisted only once.

pawelpacana:
Added #627 to address the discrepancy between timestamps.



> There are 2 created_at stored:
> ...
> 2nd in event_store_events_in_streams table

Timestamp when the event was linked is the least interesting to me. It does not say when something happened. I'd focus on `event_store_events`.`created_at` once the discrepancy is resolved.

masylum:
closing until there is progress upstream

PR #600, opened 2019-05-28
Mappers & transformations separated

mpraglowski:
Trying to introduce some structure here, and separation between mappers
& transformations.


ruby_event_store/lib/ruby_event_store/mappers/transformation/encryption.rb
```diff
@@ -0,0 +1,125 @@
+module RubyEventStore
+  module Mappers
```
swistak35: Maybe `EncryptionKey` and `ForgottenData` could be here as well?

pawelpacana: The reason for it probably was:
- no change of already published (released) consts
- no divergence in class name and its location in filesystem structure

PR #601, opened 2019-05-31
Return `self` to make `correlate_with` chainable

joelvh:


mpraglowski:
Makes sense. @joelvh would you add missing test?

joelvh:
@mpraglowski added a test

PR #610, opened 2019-06-06
Browser caused events

swistak35:


swistak35:
Screenshot_20190606_221051

Issue #439, opened 2018-08-24
AfterCommit AsyncProxyStrategy and Transactional Testing Issue


mpraglowski:
Transactional testing, whether by RSpec's use_transactional_fixtures or DatabaseCleaner's transactional cleaning strategy, is incompatible with the AfterCommit AsyncProxyStrategy.

The relevant code is as follows:

```
module AsyncProxyStrategy
  class AfterCommit
    def call(klass, event)
      if ActiveRecord::Base.connection.transaction_open?
        ActiveRecord::Base.
          connection.
          current_transaction.
          add_record(AsyncRecord.new(klass, event))
      else
        klass.perform_later(YAML.dump(event))
      end
    end
  end
end
```

Transactional testing always leaves an open transaction, and any after_commit hooks associated with this outermost transaction are never executed, instead of being scheduled immediately.

Please advise.

Regards,
Joseph

paneq:
As far as I know it works fine since Rails 5:

* https://github.com/rails/rails/pull/18458
* https://github.com/grosser/test_after_commit

paneq:
Clarification: AfterCommit callbacks are executed in transactions tests when there is 1 transaction opened  (the one wrapping whole test) but does that work with our code? I don't know. I believe it should. It would be nice to know more info (rails version) and have some code to reproduce.

PR #616, opened 2019-06-21
Ensure AfterCommitAsyncDispatcher runs handlers in tests

scottbartell:
I was using the `AfterCommitAsyncDispatcher` along with Sidekiq workers and I couldn't get integration tests working (with inlined Sidekiq workers and Database Cleaner's transaction strategy).

This solution follows a similar pattern from change to Rails https://github.com/rails/rails/pull/18458

### DIFF
--- a/rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
+++ b/rails_event_store/lib/rails_event_store/after_commit_async_dispatcher.rb
@@ -13,11 +13,12 @@ module RailsEventStore
     end
 
     def run(&schedule_proc)
-      if ActiveRecord::Base.connection.transaction_open?
-        ActiveRecord::Base
-          .connection
-          .current_transaction
-          .add_record(AsyncRecord.new(self, schedule_proc))
+      transaction = ActiveRecord::Base.connection.current_transaction
+
+      if transaction.joinable?
+        transaction.add_record(
+          AsyncRecord.new(self, schedule_proc)
+        )
       else
         yield
       end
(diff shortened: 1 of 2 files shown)


pawelpacana:
Interesting, I'll have a look at this!

pawelpacana:
For the record from the mentioned commit:
>    after_commit callbacks run after committing a transaction whose parent
> 	    is not `joinable?`: un-nested transactions, transactions within test cases,
> 	    and transactions in `console --sandbox`.

pawelpacana:
Thank you, it looks like a solid improvement!

Before I'd do something like this:

```ruby
class CleanerStrategy
  attr_reader :strategy, :options

  def initialize
    @strategy = :transaction
    @options  = {}
  end

  def call(metadata)
    if metadata[:type] == :feature && metadata[:strategy].nil?
      @strategy = :truncation
      @options  = {pre_count: true}
    elsif metadata[:strategy].present?
      @strategy = metadata[:strategy]
      @options  = metadata[:strategy] == :truncation ? {pre_count: true} : {}
    end
    self
  end

  def test_transaction?
    strategy == :transaction
  end
end
```

```ruby
require 'database_cleaner'
require 'cleaner_strategy'

RSpec.configure do |config|
  config.around(:each) do |example|
    cleaner = CleanerStrategy.new.(example.metadata)
    DatabaseCleaner[:active_record].strategy = cleaner.strategy, cleaner.options
    DatabaseCleaner.cleaning do
      example.run
    end
  end
end
```

```ruby
RSpec.configure do |config|
  config.before(:each) do |example|
    dispatcher = if CleanerStrategy.new.(example.metadata).test_transaction?
                   EventStore::ImmediateEventDispatcher
                 else
                   EventStore::AsyncEventDispatcher
                 end

    event_store =
      EventStoreAdapter.new(
        RailsEventStore::Client.new(
          repository: RubyEventStore::InMemoryRepository.new,
          dispatcher: dispatcher,
          ....
        )
      )
```

pawelpacana:
Btw. it seems to have been discussed in #439 previously

pawelpacana:
Btw. released in https://github.com/RailsEventStore/rails_event_store/releases/tag/v0.41.0

Issue #637, opened 2019-07-01
Load historical database schema from SQL dumps


pawelpacana:
At the moment we use subprocesses with several gemfiles to load database schema as published in recent versions on the gem.

We use this schema to verify that migrations between versions work as expected.

It is no surprise this way of testing is slow. Loading database schema from SQL may bring some improvements. It would be possible to have database dumps with some initial data in them (to verify data stays intact after such migration).

Where to look: 
- https://github.com/RailsEventStore/rails_event_store/blob/409339adbe094dabd439e15e46daaf145be7f647/support/helpers/subprocess_helper.rb
- https://github.com/RailsEventStore/rails_event_store/blob/409339adbe094dabd439e15e46daaf145be7f647/support/helpers/schema_helper.rb

pawelpacana:
Apparently we'd be dealing with different SQL dialects and besides that we have slightly different schemas per data store (i.e. uuid column on postgres, blob column on sqlite, etc.).

I'm thinking now that extracting migration files from published gems (like `support/helpers/0_18_2_migration_template.rb`) paired with `run_support_migration` is a better idea.

pawelpacana:
As long as we'll need old gemfiles and subprocesses to actually run older RES code, there's no benefit of changing this loading process.

PR #645, opened 2019-07-11
Test RES on Rails 6.0.0.rc1

### LINKED ISSUE #632: Test on upcoming Rails 6 version
https://weblog.rubyonrails.org/2018/12/20/timeline-for-the-release-of-Rails-6-0/
https://weblog.rubyonrails.org/2019/4/24/Rails-6-0-rc1-released/

pawelpacana:
Closes #632

Checklist:
- [x] conditional sqlite gem version for Rails 6, that is ~> 1.4.1
- [x] add dummy app for Rails 6
- [x] decide how to approach `activerecord-import` version conflict when prefilling historical data for migrations with older gem version


pawelpacana:
#637 would help

pawelpacana:
We cannot change past dependencies. If we require certain version of `activerecord-import` to work with Rails 6 that is incompatible with past release constraints, it just means we will not test migrations specs with Rails 6. At least not on sqlite3. I think that's fine.

PR #658, opened 2019-07-25
Add register unique subscriptions

alejanderl:
Would make sense to ensure the uniqueness in subscriptions?

I like the idea of registering subscribers in the event class but not in an initializer but seems like some jobs get executed multiple times when following this pattern.


pawelpacana:
Let me guess — there's Rails code reloading involved (in development) and each class reload resubscribes handlers when they're defined on such class.

How would you ensure idempotence of the simplest handler? That is `->(event) { }`

alejanderl:
I guess you mean uniqueness but I see your point. Then it makes more sense to take care of that in a wrapper class for my use case.
Thanks.

PR #662, opened 2019-08-08
Go To Stream field

swistak35:
This PR adds a Go To Stream input field in top-right corner of a Browser.


swistak35:
Screenshot_20190808_210134

Issue #290, opened 2018-04-12
Reusing aggregate instances leads to WrongExpectedEventVersion error


andrzejkrzywda:
```

**Comment by paneq** [comment]
Point 1: Circular sync handlers is rarely a good idea imho so I am leaning towards _won't fix_.

**Comment by paneq** [comment]
Point2: I was thinking about more optimistic approach.

Current code:

```ruby
  def store(stream_name = @loaded_from_stream_name, event_store: default_event_store)
    event_store.publish_events(unpublished_events, stream_name: stream_name, expected_version: version)
    @version += unpublished_events.size
    @unpublished_events = nil
  end
```

Potential new code (might have bugs, just an idea)

```ruby
  def store(stream_name = @loaded_from_stream_name, event_store: default_event_store)
    publish_those_events = unpublished_events.to_a
    version = @version
    @version += unpublished_events.size
    @unpublished_events = nil
    event_store.publish_events(publish_those_events, stream_name: stream_name, expected_version: version)
  rescue
    # restore state and bubble up
    @version -= publish_those_events.size
    @unpublished_events = publish_those_events
    raise
  end
```

however that's not really good enough as the whole state reached by _applying_ events is not longer correct so you can still shoot yourself in the foot if you try to reuse that cached aggregate instance. On any exception the instance should be discarded from cache. To incentivize it we could use `freeze` and make the instance unusable.

```ruby
  def store(stream_name = @loaded_from_stream_name, event_store: default_event_store)
    publish_those_events = unpublished_events.to_a
    version = @version
    @version += unpublished_events.size
    @unpublished_events = nil
    event_store.publish_events(publish_those_events, stream_name: stream_name, expected_version: version)
  rescue
    # restore state, freeze and bubble up
    @version -= publish_those_events.size
    @unpublished_events = publish_those_events
    freeze
    raise
  end
```

This however might be conflicting with #107 #140 ideas which wanted to react to concurrency issues

![](https://memegenerator.net/img/instances/59082792/freeze-motherfucker.jpg)

cc @andrzejkrzywda

swistak35:
More thoughts from discussions with @andrzejkrzywda @fidel and @pawelpacana today:

1) We should not care for now, about aggregate being possibly inconsistent when some error happens, because this inconsistency is possible even without scenario mentioned in the first post:

```
class Order
  include AggregateRoot

  def some_action
    @something = 42
    raise SomeInvariantViolated if some_invariant_violated?
[...]

PR #663, opened 2019-08-08
Reusing in-memory aggregate instances

swistak35:
This is WIP.

- [ ] add the same argument to RailsEventStore::Client
- [ ] decide on the naming of `after_store_callback`

About the second point, options were enlisted here: https://github.com/RailsEventStore/rails_event_store/issues/290#issuecomment-505631048

There was no other suggestions than proposed by me in that comment, so:
1) yield
2) callback named argument (of course we can name it differently, if someone has better idea. I also thought that maybe we could name it `_after_store_callback` and NOT mention it in documention to sign that this is experimental/private feature which may change?)

### DIFF
--- a/ruby_event_store/spec/client_spec.rb
+++ b/ruby_event_store/spec/client_spec.rb
@@ -75,6 +75,14 @@ module RubyEventStore
       expect(client.read.stream(stream).to_a).to eq([first_event, second_event])
     end
 
+    specify 'publishing calls after_store_callback if provided' do
+      some_var = 42
+
+      client.publish(TestEvent.new, after_store_callback: ->() { some_var = 13 })
+
+      expect(some_var).to eq(13)
+    end
+
     specify 'append many events' do
       client = RubyEventStore::Client.new(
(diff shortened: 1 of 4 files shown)


ruby_event_store/spec/client_spec.rb
```diff
@@ -72,6 +72,14 @@ module RubyEventStore
       expect(client.read.stream(stream).to_a).to eq([first_event, second_event])
     end
 
+    specify 'publishing calls after_store_callback if provided' do
+      some_var = 42
+
+      client.publish(TestEvent.new, after_store_callback: ->() { some_var = 13 })
+
+      expect(some_var).to eq(13)
+    end
```
paneq: This test fails to specify why

```
client.publish(TestEvent.new)
some_var = 13 
expect(some_var).to eq(13)
```

is not a valid solution and why `after_store_callback` needs to be invoked after storing but before calling a broker.

pawelpacana:
Not on the radar in near future

PR #664, opened 2019-08-13
Pagination buttons

swistak35:
Little changes to pagination buttons.
Previously only the active buttons were shown, which meant that after few times you've clicked "next" (and got to the last page), you had "previous" button in the same place, which was pretty frustrating experience. Now, the inactive buttons are just disabled. Also, the order of buttons has changed, from `first / last / next / previous` to more commonly used `first / previous / next / last`. 

Screenshot_20190813_204113


joelvh:
Thank you :-) It was very confusing otherwise.

Issue #641, opened 2019-07-01
Extract Rack metadata middleware from RailsEventStore gem


pawelpacana:
It is possible to use Rack middleware that enhances recorded events metadata outside of Rails. Yet we ship it in the Rails version of the gem (`rails_event_store`): https://github.com/RailsEventStore/rails_event_store/blob/409339adbe094dabd439e15e46daaf145be7f647/rails_event_store/lib/rails_event_store/middleware.rb

A wider audience could benefit from it if that was present in `ruby_event_store`. Still, `rails_event_store` may continue enabling it by default: 
https://github.com/RailsEventStore/rails_event_store/blob/409339adbe094dabd439e15e46daaf145be7f647/rails_event_store/lib/rails_event_store/middleware.rb

joelvh:
What are your thoughts about putting much more into “contrib”? Things like Rack and workers and other plumbing to integrate in many ways? All the modular “plugins”, essentially. I saw your other ticket about documenting how to use RES with Hanami and Sidekiq, and also to separate AR from Rails — put all the pieces in contrib and then have a place that helps pull the pieces together for different integrations, etc.

pawelpacana:
@joelvh Feel free to put more intro `contrib/`. So far the common thing about it is that code there does  not get released along "official" RES gems and does not necessarily has to follow 100% mutation coverage nor versioning rules.

I'd be glad to see there more stuff ✨

pawelpacana:
Related: https://www.loom.com/share/49caaa09ded0452db024f199e20eb9ae

Issue #135, opened 2017-10-05
Respect subscribers defined order


mostlyobvious:
_From @gottfrois on June 20, 2016 11:36_

When defining global and per event subscribers, we should respect in what order they are defined.

Today we trigger global subscribers after per event subscribers:

``` ruby
def all_subscribers_for(event_type)
  subscribers[event_type] + @global_subscribers
end
```

https://github.com/arkency/ruby_event_store/blob/master/lib/ruby_event_store/pub_sub/broker.rb#L50

My use case is the following:
- For all events, define a subscriber that will broadcast the event into rabbitmq broker
- For specific events, define subscribers that are basically Sagas (to trigger other events based on conditions)

``` ruby
::RailsEventStore::Client.new.tap do |es|
  es.subscribe_to_all_events(::Broadcaster.new)
  es.subscribe(::Consumers::Foo.new, [::Events::SomeEvent])
  es.subscribe(::Consumers::Bar.new, [::Events::SomeEvent])
end
```

I expected `Broadcaster` to be called first but it is call at the end, after `Foo` and `Bar` consumers.

_Copied from original issue: RailsEventStore/ruby_event_store#22_

pawelpacana:
_From @mpraglowski on June 20, 2016 11:42_

I would not rely on order of events. I would make the dependencies between handlers more explicit.

pawelpacana:
_From @mpraglowski on June 20, 2016 11:44_

BTW could you share implementation of RabbitMQ broker ? 

And please check PRs #20 & #21 - this might affect your code soon ;)

pawelpacana:
_From @gottfrois on June 20, 2016 11:54_

Thanks for the PRs, dynamic subscription might affect me in deed. Regarding the broker, I have not implemented a different broker than the one defined in ruby event store.

Since I still want to be able to do event subscriptions locally, and simple broadcast in the wild using rabbit. The easiest way was to have a `subscribe_all`.

The thing is, order is important. In my case, different services (on different hosts) subscribes to those events and one of the latest handler might trigger a new event that needs to happen after the one currently in the loop.

pawelpacana:
_From @gottfrois on November 23, 2016 9:13_

For futur references, here is the approach I choosed to handle event/consumer mapping:

```ruby
module EventHandlers
  class Foo < ActiveJob::Base
    prepend EventConsumer

    consume Events::Something::Happened

    def perform(event)
      # do something with the event
    end
  end
end
```

```ruby
[...]

PR #665, opened 2019-08-16
Subscriptions redesigned

mpraglowski:


joelvh:
I like this direction. What issues are you addressing with this?

mpraglowski:
I've started with instrumentation of subscriptions. In current codebase it would be easy to just add instrumentation when you add subscription but unsubscribe is not that easy. Later I've have other ideas, so current implementation aims at:

* allow to instrument subscribe & unsubscribe
* define Subscription object to encapsulate all operations on subscription (no more lambda to unsubscribe, just find your subscription & call `unsubscribe` method)
* easy access to information about subscriptions
* allow to provide your own subscription store (will allow to implement persisted subscriptions)

joelvh:
Ok thanks!

joelvh:
We created a class-based DSL to make it easy to create event handlers for read models. We have an  `async` option that builds a Sidekiq worker class with the given proc. (It uses a naming convention for the Sidekiq worker process to resolve the proper worker class.)

I wonder if this new object model might make it possible to enable some `async` convention as well? The main challenge is to identify  the correct proc between the web and worker processes.

Thoughts?

Example DSL we use:

```ruby
class UserReadModelDenormalizer < EventHandler
  on Events::UserRegistered, async: true do |event|
    Mailers::RegistrationConfirmation.deliver(event.data[:email])
  end
end
```

This would create a class such as `UserReadModelDenormalizer::Worker_HandleUserRegistration` that gets enqueued and picked up  by Sidekiq.

joelvh:
I should note, to  make  sure naming is consistent, I think we actually  explicitly specify a name for the worker currently using `async: 'HandleUserRegistration'` which creates `UserReadModelDenormalizer::Worker_HandleUserRegistration`

joelvh:
cc @paneq @pawelpacana

joelvh:
@paneq @pawelpacana I saw something about "contrib" flying by a little while ago. I have developed various DSLs to organize things in our projects and wonder if some of that might be good to go in contrib? It's mostly base classes with DSLs similar to what I commented with here, to define handlers and also commands and some other things. Would that be a fit for contrib?

joelvh:
One more thought - would RailsEventStore ever use things from "contrib", or is contrib a superset of Rails?

joelvh:
Ah, I was thinking of #641

mpraglowski:
@joelvh 

> I saw something about "contrib" flying by a little while ago.

Contrib is free for all to experiment - not released yet, don't know if it should be as a single gem or each idea as a separate gem under common namespace - to be decided yet. Best ideas could be incorporated into RES.

> I have developed various DSLs to organize things in our projects and wonder if some of that might be good to go in contrib?

Sure! It will be good to have it there. Fee free to create new folder and gem inside contrib - similar to my `transformations` gem https://github.com/RailsEventStore/rails_event_store/tree/master/contrib/transformations

ruby_event_store/lib/ruby_event_store/global_subscription.rb
```diff
@@ -0,0 +1,59 @@
...
+      @store.add(self)
+    end
+
+    def call(event)
+      (Class === subscriber ? subscriber.new : subscriber).call(event)
+    end
+
+    def unsubscribe
+      @store.delete(self)
+    end
```
pawelpacana: with `#call` aliased to `#unsubscribe` this would not be a breaking change, but `#call` is already taken...

ruby_event_store/lib/ruby_event_store/global_subscription.rb
```diff
@@ -0,0 +1,59 @@
...
+      EOS
+    end
+
+    # Two subscriptions are equal if:
+    # * they are of the same class
+    # * have identical subscriber (verified with eql? method)
+    #
+    # @param other [GlobalSubscription, Object] object to compare
+    #
+    # Event equality ignores metadata!
```
pawelpacana: Event... copy paste?

ruby_event_store/lib/ruby_event_store/subscription.rb
```diff
@@ -0,0 +1,65 @@
...
+    # This hash is based on
+    # * class
+    # * event types
+    # * subscriber objecy id
+    def hash
+      # We don't use metadata because == does not use metadata
+      [
+        self.class,
+        event_types,
+        subscriber.object_id
```
pawelpacana: why `object_id`?

ruby_event_store/lib/ruby_event_store/subscriptions.rb
```diff
@@ -33,6 +33,42 @@ def all_for(event_type)
...
+      end
+
+      def all_for(event_type)
+        @subscriptions[event_type]
+      end
+    end
+
+    class GlobalStore
+      def initialize
+        @subscriptions = []
```
pawelpacana: Wouldn't it make more sense to have a `Set` here?

ruby_event_store/lib/ruby_event_store/subscriptions.rb
```diff
@@ -33,6 +33,42 @@ def all_for(event_type)
     private
     attr_reader :local, :global, :thread
 
+    class Store
+      def initialize
+        @subscriptions = Hash.new {|hsh, key| hsh[key] = [] }
```
pawelpacana: If we don't intend to store same subscription multiple times, I'd rather see `Set` here

ruby_event_store/lib/ruby_event_store/subscriptions.rb
```diff
@@ -67,52 +71,35 @@ def delete(subscription)
...
-      def add(subscription)
-        GlobalSubscription.new(subscription, @store)
-      end
+      private
 
-      def all_for(_event_type)
-        @store.all
+      def build_store(klass)
+        var = Concurrent::ThreadLocalVar.new(klass.new)
+        var.value = klass.new
```
pawelpacana: Is `var.value=` call needed right after constructing `ThreadLocalVar`?

pawelpacana:
```
v = ThreadLocalVar.new(14)
v.value #=> 14
v.value = 2
v.value #=> 2
```

ruby_event_store/lib/ruby_event_store/global_subscription.rb
```diff
@@ -1,17 +1,17 @@
 module RubyEventStore
   class GlobalSubscription
-    def initialize(subscriber, store)
+    def initialize(subscriber, store: nil)
```
pawelpacana: I don't yet see the purpose, checking next commits.

ruby_event_store/lib/ruby_event_store/in_memory_subscriptions_store.rb
```diff
@@ -0,0 +1,27 @@
+module RubyEventStore
+  class InMemorySubscriptionsStore
+    def initialize
+      @subscriptions = Hash.new {|hsh, key| hsh[key] = [] }
+    end
+
+    def add(subscription, type = nil)
```
pawelpacana: https://blog.arkency.com/using-singleton-objects-as-default-arguments/

ruby_event_store/lib/ruby_event_store/in_memory_subscriptions_store.rb
```diff
@@ -0,0 +1,27 @@
...
+      @subscriptions[type.to_s] << subscription
```
pawelpacana: Why `to_s`? Aren't we expecting `Event#type` after all?

ruby_event_store/lib/ruby_event_store/in_memory_subscriptions_store.rb
```diff
@@ -4,22 +4,18 @@ def initialize
       @subscriptions = Hash.new {|hsh, key| hsh[key] = [] }
     end
 
-    def add(subscription, type = nil)
+    def add(subscription, type = GLOBAL_SUBSCRIPTION)
```
pawelpacana: 👍 

I'd rather call it `ANY_EVENT`

ruby_event_store/lib/ruby_event_store/subscription.rb
```diff
@@ -1,6 +1,8 @@
 module RubyEventStore
   class Subscription
     def initialize(subscriber, event_types = [GLOBAL_SUBSCRIPTION], store: nil)
+      raise SubscriberNotExist, "subscriber must exists" unless subscriber
```
pawelpacana: 🇬🇧

ruby_event_store/lib/ruby_event_store/subscription.rb
```diff
@@ -15,6 +17,10 @@ def unsubscribe
       event_types.each{ |type| @store.delete(self, type) } if @store
     end
 
+    def global?
+      event_types.include?(GLOBAL_SUBSCRIPTION)
```
pawelpacana: Is it always `event_types == [GLOBAL_SUBSCRIPTION]` or do we allow a `Subscription` to be both global and for some particular type?

ruby_event_store/lib/ruby_event_store/spec/subscription_store_lint.rb
```diff
@@ -0,0 +1,40 @@
+RSpec.shared_examples :subscription_store do |subscription_store|
+  FirstEvent = Class.new(RubyEventStore::Event)
+  SecondEvent = Class.new(RubyEventStore::Event)
+
+  specify do
+    expect(subscription_store.all).to eq []
+    expect(subscription_store.all_for(FirstEvent)).to eq []
```
pawelpacana: mostly ok but since we allow different event instances I think I'd prefer 

```ruby
def first_type
  FirstEvent.new.type
end
```

ruby_event_store/lib/ruby_event_store/subscription.rb
```diff
@@ -1,32 +1,56 @@
...
+    # Instantiates a new subscription
+    #
+    # @param subscriber - a callable object / class or lambda execured when subscription is triggered
+    # @param event_types [Array<Class, String>] - types of events for which subscription should be triggered. When no event types are given a global subscription is created.
+    # @param store - subscription store where subscription is persisted. When subscription store is nil a subscription is not persisted (usefull in tests).
+    #
+    # @return [Subscription]
     def initialize(subscriber, event_types = [GLOBAL_SUBSCRIPTION], store: nil)
-      raise SubscriberNotExist, "subscriber must exists" unless subscriber
+      raise SubscriberNotExist, 'subscriber must exists' unless subscriber
```
pawelpacana: 🇬🇧

ruby_event_store/lib/ruby_event_store/subscription.rb
```diff
@@ -41,7 +65,7 @@ def inspect
     def ==(other)
       other.instance_of?(self.class) &&
         other.event_types.eql?(event_types) &&
-        other.subscriber.eql?(subscriber)
+        other.subscriber.equal?(subscriber)
```
pawelpacana: Are you sure we want to check if it is the same instance in memory?

pawelpacana:
Please consider https://github.com/RailsEventStore/rails_event_store/issues/135

Issue #627, opened 2019-06-25
Unify storage of event timestamp


pawelpacana:
RES sets timestamp of domain event at the moment of its publication (here: https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/client.rb#L294 ). 

But that timestamp stored in metadata is not the same timestamp as set in event's table `created_at` attribute. In RESAR adapter, it is set by ActiveRecord specifically.

It would be great to be able to rely on `created_at` column for filtering (see #593).

In fact we can maintain current API in which there's a timestamp in event object metadata. We don't need to continue persisting it in metadata column in database table. Mapper, or specifically a Transformation can be responsible for repackaging the timestamp between an `Event` and a `SerializedRecord`. It might require to extend `SerializedRecord` with a `timestamp` attribute.

pawelpacana:
Done #764

PR #674, opened 2019-10-04
Unify timestamp storage

pawelpacana:
[#627]

Breaking changes so far:
- format of serialized message for async jobs (problematic when deploying changes — workers need strategy to handle both formats for a while, also RES client must be able to deserialize both)
- ROM timestamps changed to local time as Sequel expects, were previously put as UTC interpreted as a local time — mostlī harmless if you acknowledge time skew before certain point in time

Doubts:
- SerializedRecord#timestamp as a Time object or just an ISO8601 string? The latter better for async jobs, the former suits adapters better and requires less instantiations — do we miss some other kind of SerializedRecord or need a different internal Record?
- repositories use SerializedRecord#timestamp as a source for persisted timestamp (created_at in schema)

Future extensions on in this PR:
- exposing link time (created_at in event_store_events_in_streams)
- bi-temporal modeling, need one (or more) additional timestamp
- filtering by timestamp


rails_event_store_active_record/lib/rails_event_store_active_record/event_repository.rb
```diff
@@ -49,7 +49,11 @@ def count(specification)
     end
 
     def update_messages(messages)
-      hashes = messages.map(&:to_h)
+      hashes = messages.map do |m|
+        m_ = m.to_h
+        m_[:created_at] = m_.delete(:timestamp)
```
swistak35, line 54: Maybe do that in line below? seems to do similar job

pawelpacana:
Probably needs database migrations for mysql to upgrade timestamp resolution.

contrib/transformations/lib/transformations/with_indifferent_access.rb
```diff
@@ -8,7 +8,8 @@ def dump(item)
         event_id:   item.event_id,
         metadata:   ActiveSupport::HashWithIndifferentAccess.new(item.metadata).deep_symbolize_keys,
         data:       ActiveSupport::HashWithIndifferentAccess.new(item.data).deep_symbolize_keys,
-        event_type: item.event_type
+        event_type: item.event_type,
```
mpraglowski, line 11: We should exclude this change in contrib project. It depends on already released RubyEventStore gem version and this will break the tests as timestamp changes has not yet been released. Also IMO we should avoid mixing contrib projects changes into PR with main gems code.

PR #682, opened 2019-10-31
Only fetch handled events in projections

### LINKED ISSUE #681: Speed up projections by only fetching projected events
We noticed that running a projection from all streams was prohibitively slow and we were going to hand-roll our own projections then filter on event type.

After playing around I found that we could use the existing ruby event store projection with a small change:

```ruby
class FilteredProjection < RubyEventStore::Projection
  def read_scope(event_store, stream, count, start)
    super(event_store, stream, count, start).of_type(*handlers.keys.uniq)
  end
end
```

The `when` methods registering handlers for the projection should tell you exactly which events to pull out.

The above isn't ideal because we are overriding methods that aren't part of the public interface. I can submit a PR when I get a chance later.

brentsnook:
Speeds up projections immensely. Resolves #681

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/projection.rb
+++ b/ruby_event_store/lib/ruby_event_store/projection.rb
@@ -82,7 +82,7 @@ module RubyEventStore
     end
 
     def read_scope(event_store, stream, count, start)
-      scope = event_store.read.in_batches(count)
+      scope = event_store.read.in_batches(count).of_type(*handlers.keys.uniq)
       scope = scope.stream(stream) if stream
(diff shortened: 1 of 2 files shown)


adriancb:
👏

brentsnook:
Can anyone help me understand why these mutation tests are broken? I checked and main repo is also busted when I run `make mutate` locally.

pawelpacana:
Thank you for coming up with an idea and implementation! I've rebased previous, similar work that addressed this problem in #709 and this is now merged.

PR #693, opened 2019-12-26
Add missing end in the docs

ashkan18:
Putting an `end` to on a super minor issue in the docs 😄. Setting request metadata example was missing `end`.


pawelpacana:
Thank you 🖤

PR #695, opened 2020-01-17
Allow missing events

casiodk:
Proposal for allowing missing events ( for example in local dev env)

Often when developing I change my mind about naming of events and I rename them. It then annoys me that I cant see any events in ruby event store browser.

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/mappers/transformation/domain_event.rb
+++ b/ruby_event_store/lib/ruby_event_store/mappers/transformation/domain_event.rb
@@ -19,6 +19,8 @@ module RubyEventStore
             metadata: item.metadata,
             data:     item.data
           )
+        rescue NameError => e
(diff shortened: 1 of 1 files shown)


casiodk:
I didn't test this, so just view it as a proposal, thanks :)

andrzejsliwa:
@casiodk what about `RubyEventStore::Mappers::Default.new(events_class_remapping: events_class_remapping)`? You are able to define class remapping in such cases:

Screenshot 2020-02-05 at 09 45 50

But frankly speaking I would delete just whole stream instead. Skipping events can be really error prone in case when state of aggregate will be inconsistent due to some logic constraints who depends on specific applying events.

casiodk:
Thanks for the tips, yes i ended up deleting the stream, but it would be nice to have some fallback behaviour in development, so it just works even if events are missing

PR #711, opened 2020-02-26
Failing test case for event class remapping

pawelpacana:


mpraglowski:
Nope https://github.com/RailsEventStore/rails_event_store/pull/711/commits/b6b0b3c77126765663fec2c8351304a812202317 - it will not work, just missing rebase :( - the problem is that `Projection` build read scope based on classes declared in `when` methods, but this classes are not the ones that has been written in data store.

mpraglowski:
This is not really the problem with the `Projection` class but with `of_type` method that skips remapped events.

Issue #702, opened 2020-01-31
Utilise Ruby 2.7 deprecation warnings


fidel:
As a followup to #696 we should deal with Ruby 2.7 deprecation warnings. Most common issue is:
`Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`.

Fairly easy to fix, I suppose.

pawelpacana:
Is there a rubocop autofix for that?

dneilroth:
@fidel spun up a PR here https://github.com/RailsEventStore/rails_event_store/pull/735

mpraglowski:
Some deprecation warning still exists in gem we have in dependencies.

PR #735, opened 2020-05-25
Do not use keyword args as last parameter in method calls

dneilroth:
Issue: https://github.com/RailsEventStore/rails_event_store/issues/702

the remaining deprecation warnings in the tests are coming from rspec-mocks and the sqlite3 gem. Fixes for rspec-mocks appear to be in the works [here](https://github.com/rspec/rspec-mocks/pull/1324).

there are a ton of `gems/ruby-2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/sqlite.rb:114: warning: rb_check_safe_obj will be removed in Ruby 3.0` warnings related to the sqlite3 gem, which is fixed in version '1.4.2', so I _think_ these warnings happen in the tests that use version '1.3.13' but not rails 6.  I am not sure tbh, happy to update if anyone has a recommendation. 

@mpraglowski would love a review here 🙏


RAILS_VERSION
```diff
@@ -1 +1 @@
-6.0.2
\ No newline at end of file
+6.0.3
```
dneilroth, line 1: most of the warnings are solved by bumping to this version of rails

dneilroth, line 1: this solves the many warnings that are coming from rails dependencies https://weblog.rubyonrails.org/2020/5/6/Rails-6-0-3-has-been-released/

rails_event_store/spec/active_job_scheduler_spec.rb
```diff
@@ -58,19 +60,21 @@ module RailsEventStore
...
-              "event_id" => "83c3187f-84f6-4da7-8206-73af5aca7cc8",
-              "event_type" => "RubyEventStore::Event",
-              "data" => "--- {}\n",
-              "metadata" => "---\n:timestamp: 2019-09-30 00:00:00.000000000 Z\n",
-              "_aj_symbol_keys" => ["event_id", "data", "metadata", "event_type"]
-            }
-          ],
-          queue: "default"
-        })
+        expect(enqueued_jobs[0]).to include(
```
dneilroth, line 63: the enqueued_job now includes a bunch of extra metadata, for example:
```
'job_class' => 'RailsEventStore::MyAsyncHandler',
'job_id' => random uuid,
'provider_job_id' => nil,
'queue_name' => 'default',
'priority' => nil,
'arguments' => [
  {
    'event_id'        => '83c3187f-84f6-4da7-8206-73af5aca7cc8',
    'data'            => "--- {}\n",
    'metadata'        => "---\n:timestamp: 2019-09-30 00:00:00.000000000 Z\n",
    'event_type'      => 'RubyEventStore::Event',
    '_aj_symbol_keys' => %w[event_id data metadata event_type],
  },
],
'executions' => 0,
'exception_executions' => {},
'locale' => 'en',
'timezone' => 'UTC',
'enqueued_at' => timestamp string,
```
I am having trouble tracking down exactly why that is getting added now, presumably due to one of the dependency bumps. I changed the test to use `include` instead of `eq` so that it will check if those keys + values exist in the job, but it will not check if it is an exact match (since it no longer is).

fidel:
Thanks for your contribution @dneilroth!

PR #737, opened 2020-06-05
Outbox

swistak35:


joelvh:
I see -- so it's in effect an asynchronous message broker?

PR #749, opened 2020-07-27
Use Event.import instead of making N save calls

mpraglowski:
This will lower number of SQL calls from N+2
to 3 (in best case) (where N is number of events).


pawelpacana:
For a brief moment I was worried about atomicity of adding events after these changes but there is a transaction block hidden inside `add_to_stream`.

pawelpacana:
I also realized that `to_event_id.call(element)` might have been a relict of database-generated uuids which we do not or want utilize in any way but maybe an author had this on thier mind.

pawelpacana:
Getting rid of `to_event_id.call(element)` first would have been much appreciated for review comfort ;)

PR #740, opened 2020-06-18
Multiple databases repository

mpraglowski:
Described here https://blog.arkency.com/rails-multiple-databases-support-in-rails-event-store/

PR #753, opened 2020-08-18
Multiple databases repository support

mpraglowski:
Extracted changes in RES required by sample app from #740


mpraglowski:
I've run some benchmarks using https://github.com/mlomnicki/res-benchmarks

The results:

```
./run.sh benchmarks/read_stream.rb

Comparison:
master:      119.2 i/s
multiple_database:      114.8 i/s - same-ish: difference falls within error
```

```
./run.sh benchmarks/publish.rb

Comparison:
master:      159.8 i/s - same-ish: difference falls within error
multiple_database:      161.9 i/s
```

PR #837, opened 2020-11-10
Allow to specify event type resolver for subscriptions

mpraglowski:
Make implicit explicit. Type resolver is reponsible
for converting event type from anything (class, string, etc)
to event type (string preffered here).

Default implementation (backward compatible) is just
lambda using to_s method to convert event type
given as class or string to string.

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/spec/subscriptions_lint.rb
+++ b/ruby_event_store/lib/ruby_event_store/spec/subscriptions_lint.rb
@@ -114,4 +114,23 @@ RSpec.shared_examples :subscriptions do |subscriptions_class|
 
     expect(subscriptions.all_for('Test1DomainEvent')).to eq([handler, handler])
   end
+
+  it 'subscribes by type of event which is a class' do
+    handler         = TestHandler.new
+    subscriptions.add_subscription(handler, [Test1DomainEvent])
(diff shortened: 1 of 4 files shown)


ruby_event_store/lib/ruby_event_store/spec/subscriptions_lint.rb
```diff
@@ -114,4 +114,23 @@ def call(event)
...
+  it 'subscribes by type of event which is a class' do
+    handler         = TestHandler.new
+    subscriptions.add_subscription(handler, [Test1DomainEvent])
+    subscriptions.add_thread_subscription(handler, [Test1DomainEvent])
+
+    expect(subscriptions.all_for('Test1DomainEvent')).to eq([handler, handler])
+    expect(subscriptions.all_for(Test1DomainEvent)).to eq([handler, handler])
+  end
+
+  it 'subscribes by type of event which is a class or strong with custom type resolver' do
```
swistak35, line 127:
```suggestion
  it 'subscribes by type of event which is a class or string with custom type resolver' do
```

swistak35 (approved):

pawelpacana:
@mpraglowski this needs to cover aggregate_root as well

PR #840, opened 2020-11-16
Failed mutations on timeouts and crashes

pawelpacana:


aggregate_root/.mutant.yml
```diff
@@ -0,0 +1,11 @@
+---
+coverage_criteria:
+  # Control the timeout criteria, defaults to `false`:
+   # * `true` - mutant will consider timeouts as covering mutations.
+   # * `false` mutant will ignore timeouts when determining mutation coverage.
+   timeout: true
```
mbj, line 6: @pawelpacana Make sure to also configure a timeout, else this setting has no effect.

pawelpacana: https://github.com/RailsEventStore/rails_event_store/pull/840/commits/23ac8ea6bc95ee7460c1af1b8263bf0cab1e3f76

mbj: thanks. BTW I see you run a full pass on CI, this is not recommended, normally you should only run incremental passes.

Also 3/s is a very low speed. I'll have a look at your setup.

pawelpacana: How does incremental work when mutant is getting new or improved transformations?

mbj: > How does incremental work when mutant is getting new or improved transformations?

it would only select changed code, even if mutant gained new operators. Incremental does not replace running full runs once in a while.

Still for bigger projects its recommended. But: As I meanwhile discussed this project has very fast tests, hence you should get away with full runs.

I'll do some optimiaztions in PRs, thanks for getting me access so I can try with the CI integration.

support/make/install.mk
```diff
@@ -1,5 +1,6 @@
 install: ## Install gem dependencies
 	@echo "Installing gem dependencies"
+	@gem install bundler -v '~> 1'
```
mbj, line 3: @pawelpacana This may be controversial but allows to build consistently.

Our constraints are:

* A `Gemfile.lock` should exist
* Should we bundle with bundler `~> 2` we end up in a dependency conflict on the Rails 4.2 facet of your build.
* The `Gemfile.lock` only matters for RES developers, not users of releases.
* Bundler `2.x` can if `1.x` is installed fall back to 1.x.

So unless you are willing to drop the build for Rails 4.2, this seems to be the common denominator. 

If this where my project, I'd try to install the bundler at the correct version directly in the project specific docker image. But for now I'm going path of least resistance.

OT: Adding the `Gemfile.lock` makes the odd mutations I was trying to debug disappear. I've to assume at this point it was dependency indeterminism which `Gemfile.lock` are made to prevent.

pawelpacana: Rails 4.2 support might be quite possible to drop on `master` branch

mbj: I'd love that. But for the moment I can reproduce an odd result without the Gemfile.lock here also.

I'm happy to extract the gemfile lock addition in a separate PR.

But my main focus is to reproduce the odd result here. I can confirm the new coverage criteria improve the situation, but I'm not 100% happy (yet).

PR #853, opened 2020-12-04
Bump to mutant 0.10.14

mbj:
* This release changes to consider mutations that make rspec crash as
  covered as long `process_abort: true` is active as a coverage criteria.


mbj:
@pawelpacana With that change `rails_event_store-rspec` passes locally for me.

PR #858, opened 2020-12-07
Reproducible builds via lockfiles

pawelpacana:
Another attempt of:
023c477


mbj:
@pawelpacana Thanks for your effort here. Contributing / testing mutant improvements was always hard for me given the absence of lockfiles.

Context: Mutants licensing mechanism "assumes lockfiles" and it has bad edge case behavior on absent lockfiles.

PR #860, opened 2020-12-07
Evaled Gemfile does not work with dependabot

pawelpacana:
I'd very much like to use dependabot to bump dev dependencies,
especially with regard to rails and mutant versions used.

### DIFF
--- a/aggregate_root/Gemfile
+++ b/aggregate_root/Gemfile
@@ -4,7 +4,11 @@ source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 gemspec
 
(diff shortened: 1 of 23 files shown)


mbj:
@pawelpacana Evaled dependencies work with Dependabot if you use `eval_gemfile 'literal-string-here'`.

mbj:
You just have to avoid expressions. OT: I'd be happy to talk to the dependabot team to make a partial ruby evaluator that allows to use the common ideoms.

pawelpacana:
I'll check that — easy to revert if it works as advertised ;)

pawelpacana:
Seems to works, thanks @mbj!

PR #873, opened 2020-12-08
Full daily mutation runs

pawelpacana:


pawelpacana:
Merging to see if it is enough to trigger schedules

PR #880, opened 2020-12-08
The reason why bundler/inline exists

pawelpacana:
* integration test when subprocess-executed previous gem version
  generates data later migrated and read by newer version
* no need for Gemfile files and tedious hooking into bundle install
  process
* reusing bundler cache for single Gemfile.lock in CI possible again

https://bundler.io/guides/bundler_in_a_single_file_ruby_script.html
https://www.rubydoc.info/github/bundler/bundler/Bundler%2Ewith_unbundled_env


mbj:
@pawelpacana I like this solution. I'll likely copy it for mutants integration tests.

pawelpacana:
Needs more work around this bundle ceremony: https://github.com/ruby/setup-ruby/blob/master/index.js#L252

```
/opt/hostedtoolcache/Ruby/2.7.2/x64/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/lib/bundler/definition.rb:506:in `validate_platforms!': Your bundle only supports platforms [] but your local platforms are ["ruby", "x86_64-linux"], and there's no compatible match between those two lists. (Bundler::ProductionError)
```

pawelpacana:
Possibly related: https://github.com/rubygems/bundler/issues/7114

PR #884, opened 2020-12-09
Outbox cleanup

swistak35:


contrib/ruby_event_store-outbox/Gemfile
```diff
@@ -13,11 +13,7 @@ gem 'influxdb', '~> 0.8.0', require: false
 gem 'childprocess'
 gem 'rails', ENV['RAILS_VERSION']
```
pawelpacana, line 14: RAILS_VERSION is more, multiple Gemfiles are preferred (each having a lockfile)

swistak35: I will start with one version working and expand later

pawelpacana: s/ENV['RAILS_VERSION']/6.0.3.4/ then

.github/workflows/ruby_event_store-outbox.yml
```diff
@@ -26,3 +26,30 @@ jobs:
...
         working-directory: ${{ env.GEM_LOCATION }}
+
+  mutate:
+    runs-on: ubuntu-20.04
+    env:
+      GEM_NAME: ruby_event_store-outbox
+      BUNDLE_GEMFILE: ${{ matrix.gemfile }}
+      DATABASE_URL: ${{ matrix.database }}
+      GEM_LOCATION: contrib/ruby_event_store-outbox
+      MUTANT_JOBS: 1
```
pawelpacana, line 42: should not be needed when `sqlite3::memory:`

PR #887, opened 2020-12-09
EHLO

pawelpacana:


pawelpacana:
Zrzut ekranu 2020-12-10 o 00 31 10

Path filtering seems to work after git repo was re-linked: https://www.netlify.com/blog/2019/10/09/launching-monorepo-support-for-netlify-sites/

PR #913, opened 2020-12-16
Upgrade to mutant-0.10.20

mbj:
See: https://github.com/mbj/mutant/blob/master/Changelog.md#v01020-2020-12-16

TLDR: It may be much faster on your big passes.


mbj:
@pawelpacana I'm self merging this one. Intentionally to make sure the next daily full pass gets it.

PR #914, opened 2020-12-18
Move ruby_event_store-rom to contrib

### LINKED ISSUE #834: Phase out ROM event repository to contrib
Splitting work across two quite similar event repositories is time consuming and takes a lot of effort. It would have been easier with a contributor well-invested in ROM ecosystem.

Short list of things missing in ROM on upcoming v2.0.0 release:
* there is new indexed `valid_at` column for bi-temporal event sourcing, changes querying and sorting a bit and requires migration
* we're dropping global stream entries in `event_store_events_in_streams`, this again requires migration
* we're making more use of `created_at` columns which uncovered some bugs — wrong time being passed to sequel (utc vs local) and `update_messages` messing with `created_at`
* we're increasing timestamp precision, which again requires migration on mysql/sqlite
* we're adding time filtering on `created_at`, which changes querying a bit again

Mutation coverage is something to improve as well.

Looking forward to bring it back anytime, at the moment and without a maintainer it grabs too much attention.

mpraglowski:
Closes #834


pawelpacana:
To reconsider if we want to include ROM (and Outbox) in `coverage.yml`. I'd prefer to keep their scheduled jobs in (their) separate workflows

pawelpacana (approved):

PR #916, opened 2020-12-20
Introduce public API to return known handlers for given event type and  have_subscribed_to_events matcher

### LINKED ISSUE #842: How can I check that my handler subscribed to an event?
I can access current subscriptions, but it's a pretty ugly approach
```ruby
event_store
  .send(:broker)
  .instance_variable_get(:@subscriptions)
  .send(:local)
  .instance_variable_get(:@subscriptions)
```
Do we need a public interface to get current subscriptions or a method to be sure that a handler subscribed to an event?

StanisLove:
Resolve #842


ruby_event_store/lib/ruby_event_store/broker.rb
```diff
@@ -2,6 +2,8 @@
 
 module RubyEventStore
   class Broker
+    attr_reader :subscriptions
```
mpraglowski, line 5: I think this exposes too much of subscriptions internals. The subscriptions object has methods to add & remove subscriptions & I would live to avoid manipulating it outside ot RES.

Maybe RES should publish some "system" events when subscription is added and removed? Than you could build the read model with list of current subscriptions or even have more sophisticated logic based on it? WDYT?

mpraglowski:
Merged but I will make some minor changes.

ruby_event_store/lib/ruby_event_store/broker.rb
StanisLove: We just make `attr_reader` public so we still can't make changes in subscriptions outside RES.
I don't see any problem here.

StanisLove:
I will open PR to update `ruby_event_store-rspec` documentation soon.

ruby_event_store/lib/ruby_event_store/broker.rb
mpraglowski: But `attr_reader :subscriptions` exposes `Subscription` class instance, with methods to manipulate (add & remove) subscriptions. You can read only the object itself but you are still allowed to run any method on it.

PR #921, opened 2020-12-23
Outbox mutants

swistak35:


contrib/ruby_event_store-outbox/lib/ruby_event_store/outbox/sidekiq_producer.rb
```diff
@@ -15,13 +15,16 @@ def call(klass, args)
         normalized_item = sidekiq_client.__send__(:normalize_item, item)
         payload = sidekiq_client.__send__(:process_single, normalized_item.fetch('class'), normalized_item)
         if payload
-          Record.create!(
+          Repository::Record.create!(
```
mpraglowski, line 18: `Repository::Record` & `Repository::Lock` should be internals of `Repository` implementation. Should not be uses anywhere outside of it. `repository` instance method should be used to create records & manipulate locks.

swistak35: Yes, I wanted to do this immediately but this is a little tricky. This repository is used in two places, one (here) in producer, so the client's app, only to create records in outbox table. Second, consumer process. The tricky part is that for example there is `def intialize(database_url)` which is required for consumer process, but is not relevant for producer process (because the connection with db is already established). So I'm now thinking that there should be two repositories, `Repository::Producer` and `Repository::Consumer` :)

PR #936, opened 2020-12-30
Ruby 3

mpraglowski:


mbj:
@pawelpacana Thats the first time in many years you do not have to ask: When will mutant work on a new Ruby release ;)

PR #943, opened 2020-12-31
Drop unused fill_ids method

mpraglowski:
It has been used bt DRES, but seems to be not there anymore


pawelpacana:
Does that mean you're decommissioning DRES?

pawelpacana:
Probably ok given last RES version known to work with DRES was 0.28.

There will be more incompatibilities by now. When we decide to bring DRES back to usable state, we can revert this change if proved still needed.

pawelpacana:
The only place that could have used this was `PgLinearizedRepository` but is not.

pawelpacana:
Also 7ce94679a4a0169ff44d13ecd5e7b8134022349c

PR #948, opened 2021-01-01
Up-to-date mutant

pawelpacana:


pawelpacana:
Fails on `aggregate_root`:
- mutant is now better on regexes
- regex covers code that was extracted to separate gem (protobuf)
- event_type simplification in aggregate_root should use event_type resolver, perhaps this helps

PR #968, opened 2021-01-05
Mutant configs

pawelpacana:


mbj (commented):
@pawelpacana I think this is more clean compared to turning the makefile into a "sort of" mutant config file.

There will be more additions to the config file, some that do not even fit into the CLI.

Also: When you now run mutant manually (without the makefile) you can do this:

```
mutant --since master 'RubyEventStore::SomeClassYouRightNowFocusOn*'
```

in your local CLI, and it would pick up the config file. Unlike the previous makefile that was "lost" for manual operation.

pawelpacana:
```
mutant --since master 'RubyEventStore::SomeClassYouRightNowFocusOn*'
```

is actually equivalent to

```
make mutate-changes SUBJECT='RubyEventStore::SomeClassYouRightNowFocusOn*'
```

pawelpacana:
Because mutant keeps improving, and `matcher` in config shipped in 0.10.25, I have to deal with upgrade blockers first. But the move to config file looks promising.

mbj:
> Because mutant keeps improving, and `matcher` in config shipped in 0.10.25, I have to deal with upgrade blockers first. But the move to config file looks promising.

nice.

BTW:

> is actually equivalent to

I'm aware you can proxy any of mutants features to the makefile. The problem is that this may not be a very efficient way to spend your time. You'll always be "behind". Plus you create a nonstandard interface for your contributors that invalidates mutants documentation for them.

pawelpacana:
> The problem is that this may not be a very efficient way to spend your time. You'll always be "behind". Plus you create a nonstandard interface for your contributors that invalidates mutants documentation for them.

On the hand, proxying shields from inevitable upstream changes and contributors have unified interface to do certain things. Other side of the same coin ;)

For "modern" mutant moving to configs seems like a good bet. Ability to easily eject from makefiles is also valuable. Not having to escape ignore/subjects is the king.

mbj:
> On the hand, proxying shields from inevitable upstream changes and contributors have unified interface to do certain things. Other side of the same coin ;)

True and in the "old days" of mutant this possibly was the correct choice. I always also had these kinds of overlays in my own projects. Before I had the time (through making mutant commerical) to provide a better user experience natively.

> For "modern" mutant moving to configs seems like a good bet. Ability to easily eject from makefiles is also valuable. Not having to escape ignore/subjects is the king.

I suspect that its wise to not duplicate mutants config, given mutant is more actively developed recently.

PR #969, opened 2021-01-05
Event type resolver in aggregate root

mpraglowski:


aggregate_root/lib/aggregate_root.rb
```diff
@@ -85,6 +94,7 @@ def self.included(host_class)
   end
 
   def self.included(host_class)
-    host_class.include with_default_apply_strategy
+    host_class.extend  OnDSL
+    host_class.include with
```
pawelpacana, line 98: that reads horrible :/

mpraglowski: yeah... naming is hard ;)

PR #1015, opened 2021-03-02
Fix wrong generator command in start.html.md

keang:
This is a typo fix I found when following the guide


pawelpacana:
That is not a typo fix. There's a different generator name for v1 an v2 releases:

https://railseventstore.org/docs/v1/start/
https://railseventstore.org/docs/v2/start/

keang:
Oh I see! 
That is a surprise because I just do `gem install rails_event_store`, and then clicking "Getting started" on https://railseventstore.org/ brought me to the command that didn't work.
Happy to close this though 👍

pawelpacana:
> That is a surprise because I just do gem install rails_event_store, and then clicking "Getting started" on https://railseventstore.org/ brought me to the command that didn't work.

That's a great find! I wasn't aware of it and just fixed it: https://github.com/RailsEventStore/rails_event_store/commit/9567563383c421f5f7eb3b98460a94f42ff11d1c

Thank you!

PR #1025, opened 2021-03-16
Spike for persistent projections toolkit

swistak35:
For now only a process which tracks up to which identifier, all transactions has ended.


leifg:
I love it!

Is there any chance this will get merged or implemented as a feature of RES?

swistak35:
@leifg Yes, definitely, but there are a few different paths we can take from the current state of the spike -- we need to do slightly more research and decide which route we want to take.

mostlyobvious:
Timed out, might re-use of next spike if that happens.

PR #1103, opened 2021-08-04
Preserve types transform

mpraglowski:


ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,132 @@
...
+        private
+
+        SYMBOL_KEYS_KEY = '_res_symbol_keys'
+        RESERVED_KEYS = [
+          SYMBOL_KEYS_KEY, SYMBOL_KEYS_KEY.to_sym,
+        ]
+        PASS_THROUGH = ->(v) { v }
+
+        def transform_data(argument)
+          argument.each_with_object({}) do |(key, value), hash|
```
swistak35, line 57: Did we have an assumption that the data is always a hash?

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,132 @@
...
+
+        def load(record)
+          types = record.metadata.delete(:types)
+          data = types ? restore_types(record.data, types) : record.data
+
+          Record.new(
+            event_id:   record.event_id,
+            event_type: record.event_type,
+            data:       data,
+            metadata:   record.metadata,
```
swistak35, line 42: Why metadata isn't transformed? Timestamps and simple value objects (`RequestUuid`, `UserAgent` or just DateTime objects) may exist there too

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,132 @@
...
+
+        private
+
+        SYMBOL_KEYS_KEY = '_res_symbol_keys'
+        RESERVED_KEYS = [
+          SYMBOL_KEYS_KEY, SYMBOL_KEYS_KEY.to_sym,
+        ]
+        PASS_THROUGH = ->(v) { v }
+
+        def transform_data(argument)
```
swistak35, line 56: `transform_hash` maybe? argument may be a not be a top-level data, but a lower-level hash.

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,132 @@
...
+        end
+
+        def transform_argument(argument)
+          case argument
+          when Hash
+            transform_data(argument)
+          when Array
+            argument.map{|i| transform_argument(i)}
+          else
+            serializer_of(argument.class.name).call(argument)
```
swistak35, line 69: `.class.name` could be a passed dependency of finding out the serializer to resolve (but doesn't have to be, the programmer can easily rename a class so it's fine for me)

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
mpraglowski: Good catch... we have no such assumption

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,132 @@
...
+          end
+        end
+
+        def store_types(argument)
+          argument.each_with_object({}) do |(key, value), hash|
+            hash[serialize_hash_key(key)] = store_type(value)
+          end
+          .merge(SYMBOL_KEYS_KEY => argument.each_key.grep(Symbol).map!(&:to_s))
```
swistak35, line 77: I'd remove that logic and raise an gtfo error when someone gives us hashes with mixed strings and symbols, or at least with the same symbol defined as string (like `{ foo: 10, "foo" => 10 }`) and then just do `foo[key.to_s] || foo[key.to_sym]` when reading to try both.

swistak35: In especially that we ignore hashes with keys other than strings or symbols. So from one hand RES would be very "smart" in terms of data persistence (differentiating between symbols and strings) and on the other hand very stupid (forbidding all other kinds of hashes, i.e. hash with keys being numbers, which is perfectly valid usecase for sparse arrays). It does not click for me and is inconsistent in the approach.

swistak35: Also, the performance both in terms of CPU (data reading here is complex) and payload size, some events are big and have lot of keys, and serialization/deserialization takes time.

swistak35:
I like that :+1:

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
swistak35: Also, the serializer of the record's data (YAML, JSON, ...) does not necessarily preserve strings vs symbols differences. So:
* if we assume that strings vs. symbols are preserved, then let's make the `types` also keeping symbols and strings
* otherwise, let's just ignore it in both.

As far as I understand how it would work, we do assume that `data` could keep symbols and strings (which is not true in most cases), but it cannot keep symbols and strings in metadata (so we do the "remember the list of keys which were symbols so we can make these strings symbols again after deserialization").

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
mpraglowski: Right... adding that.

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
mpraglowski: > In especially that we ignore hashes with keys other than strings or symbols.
This we could change - keys also could be transformed using provided set of serializers/deserializers - then it could be anything as long as you provide a way to dumo/load it.

mpraglowski: > if we assume that strings vs. symbols are preserved, then let's make the types also keeping symbols and strings

That's just another serializer for `Symbol` type.

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
mpraglowski: There is no such assumption anymore.

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+module RubyEventStore
+  module Mappers
+    module Transformation
+      SerializationError = Class.new(ArgumentError)
```
pawelpacana, line 6: SerializationError for all Transformation kind? Or specific to PreserveTypes?

ruby_event_store/lib/ruby_event_store/mappers/transformation/preserve_types.rb
```diff
@@ -0,0 +1,143 @@
...
+            serializer: serializer,
+            deserializer: deserializer
+          }
+          self
+        end
+
+        def dump(record)
+          data = transform(record.data)
+          metadata = transform(record.metadata)
+          if (metadata.respond_to?(:[]))
```
pawelpacana, line 25: can we make an assumption on Metadata being always in an Event?

mpraglowski: With `RubyEventStore::Event` this assumption is always true because of its initializer implementation

```ruby
    def initialize(event_id: SecureRandom.uuid, metadata: nil, data: {})
      @event_id = event_id.to_s
      @metadata = Metadata.new(metadata.to_h)
      @data     = data
    end
 ```
 
 If we prepare event with nil metadata (or use another event class that allows that) we will get:
 
 ```ruby
 > e = RubyEventStore::Event.new
=> #<RubyEventStore::Event:0x00007fc630389408 @event_id="2755b447-78d1-4dad-aec3-29792a9b6567", @metad...

> e.metadata
=> #<RubyEventStore::Metadata:0x00007fc630388eb8 @h={}>

> e.instance_variable_set("@metadata", nil)
=> nil
> e.metadata
=> nil

> c = RubyEventStore::Client.new(repository: RubyEventStore::InMemoryRepository.new)
=> #<RubyEventStore::Client:0xb4>

> c.publish(e)
Traceback (most recent call last):
        9: from /Users/mpraglowski/.rbenv/versions/2.7.2/bin/irb:23:in `<main>'
        8: from /Users/mpraglowski/.rbenv/versions/2.7.2/bin/irb:23:in `load'
        7: from /Users/mpraglowski/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        6: from (irb):13
        5: from /Users/mpraglowski/arkency/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:32:in `publish'
        4: from /Users/mpraglowski/arkency/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:341:in `enrich_events_metadata'
        3: from /Users/mpraglowski/arkency/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:341:in `each'
        2: from /Users/mpraglowski/arkency/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:341:in `block in enrich_events_metadata'
        1: from /Users/mpraglowski/arkency/rails_event_store/ruby_event_store/lib/ruby_event_store/client.rb:347:in `enrich_event_metadata'
NoMethodError (undefined method `[]' for nil:NilClass)
```

So here in transformations I could assume metadata are present. The question is if I should make such assumptions for both data & metadata ;)

mpraglowski: BTW this works because:

```ruby
> nil.respond_to?(:[])
=> false
```

PR #1182, opened 2021-09-07
Almost working Parcel 1 -> 2 conversion

pawelpacana:
* allows postcss 8 and dropping legacy tailwind/postcss/purgecss workarounds
* no more security issues from legacy packages


pawelpacana:
I experienced some issues with watching file changes — seems not to work correctly. Perhaps it is worth exploring esbuild too — which promises same improvements (and even less dependencies). Bundling speed in our case make very little difference.

Issue #1062, opened 2021-06-01
Browser requires unsafe Content-Security-Policy


piotrekbator:
At the moment if I set up a CSP limiting `script-src` the browser won't work and the only way to make it work (AFAIK) is to add the `'unsafe-inline'` option to the CSP. Which is... unsafe :D

The reason for that is that there's an inline script in the HTML served by the browser:

````html
<script type="text/javascript">
  RubyEventStore.Browser.Elm.Main.init({
    flags: {
      /* ... */
    }
  });
</script>
````

I think that the solution would be to:

1. Provide the configuration hash (what's currently under `flags`) to a `data` attribute on `<body />` for example; probably just dumped as JSON.
2. Move the `init` invocation to the main JS file: just add some code that looks for the `data` attribute, loads the settings from there and if they're found and sound, pass them to the `init` function.

I'm not sure if the current JS file that's used by the browser is ready for that or if that's even a good idea. But I'm quite confident that supporting a safe CSP should be handled.

And I'd be happy to do some work on that, just don't want to poke around blindly. Let me know if the path I'm suggesting is fine from your perspective and I'll try to prepare a PR.

pawelpacana:
To recap, that I understand it correctly:

* adding global CSP policy (for the whole site with the browser mounted in routes) to limit inline scripts prevents browser from working (there's inline script to init the browser)
* removing inline script and moving its contents to separate script file would pass global CSP policy that limits inline scripts

Out of curiosity, is there a middle ground, i.e. overriding CSP policy for some parts of the site, possibly doable with meta tags from browser-rendered HTML or from Rack middleware? (setting header just for the browser responses)

The solution you've proposed is sound, that is:

* moving configuration to data attributes on a dom element in main browser HTML document 
* reading data attributes and passing them to Elm application flags in a separate JS file (i.e. https://github.com/RailsEventStore/rails_event_store/blob/0b653f60cee3ae6b73626ce09e2258c217329c25/ruby_event_store-browser/elm/src/index.js), where `Elm.Main.init()` will be called

It also makes exposing Elm application under `window.RubyEventStore.Browser` obsolete, which is a good thing.

For the reference: https://guide.elm-lang.org/webapps/

piotrekbator:
Yeah, that's correct.

[...]

PR #1185, opened 2021-09-09
Browser compatibility with Content Security Policy

piotrekbator:
I finally got around to working on a fix for the problem I've described in #1062 and here's what I've come up with.

First of all I added a test to ui tests that serves the browser together with a reasonable CSP header. What's important is that this header does not include a `script-src 'unsafe-inline'` which would be the only way to make the inline script initiating the `Browser` work with CSP. So this test failed before I've implemented the changes.

And the rest is what we've discussed in the issue: I've moved the settings from the inline JS to a `meta` tag and dumped them as JSON. Then the JSON is retrieved in `index.js` and used to initiate the `Browser` there.

I've also removed the code exposing the `Browser` via `window.RubyEventStore.Browser` as @pawelpacana suggested in #1062.

### DIFF
--- a/ruby_event_store-browser/elm/src/index.js
+++ b/ruby_event_store-browser/elm/src/index.js
@@ -1,4 +1,8 @@
 require("./style/style.css");
 
-window.RubyEventStore = {};
-window.RubyEventStore.Browser = require("./Main.elm");
+var Browser = require("./Main.elm");
+
+var settings = document.querySelector("meta[name='ruby-event-store-browser-settings']").getAttribute("content");
+console.debug("settings", settings);
+settings = JSON.parse(settings);
+Browser.Elm.Main.init(settings);
--- a/ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
+++ b/ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
@@ -65,21 +65,13 @@ module RubyEventStore
             <head>
               <title>RubyEventStore::Browser</title>
               <link type="text/css" rel="stylesheet" href="<%= path %>/ruby_event_store_browser.css">
+              <meta name="ruby-event-store-browser-settings" content='<%= browser_settings %>'>
             </head>
             <body>
               <script type="text/javascript" src="<%= path %>/ruby_event_store_browser.js"></script>
-              <script type="text/javascript">
-                RubyEventStore.Browser.Elm.Main.init({
-                  flags: {
-                    rootUrl:    "<%= routing.root_url %>",
-                    apiUrl:     "<%= api_url %>",
-                    resVersion: "<%= RubyEventStore::VERSION %>"
-                  }
-                });
-              </script>
             </body>
           </html>
-        }, locals: { path: settings.root_path || request.script_name, api_url: settings.api_url || routing.api_url }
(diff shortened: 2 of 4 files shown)


ruby_event_store-browser/elm/src/index.js
```diff
@@ -1,4 +1,8 @@
 require("./style/style.css");
 
-window.RubyEventStore = {};
-window.RubyEventStore.Browser = require("./Main.elm");
+var Browser = require("./Main.elm");
+
+var settings = document.querySelector("meta[name='ruby-event-store-browser-settings']").getAttribute("content");
+console.debug("settings", settings);
```
pawelpacana, line 6: Is it needed?

ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
```diff
@@ -65,21 +65,13 @@ def self.for(event_store_locator:, host: nil, path: nil, api_url: nil, environme
             <head>
               <title>RubyEventStore::Browser</title>
               <link type="text/css" rel="stylesheet" href="<%= path %>/ruby_event_store_browser.css">
+              <meta name="ruby-event-store-browser-settings" content='<%= browser_settings %>'>
```
pawelpacana, line 68: `<meta>` surprised me a bit here, I'd have expected `data-*`:
https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes

ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
```diff
@@ -94,6 +86,16 @@ def routing
           )
         end
 
+        def browser_settings
+          JSON.dump({
+            flags: {
+              rootUrl:    routing.root_url,
+              apiUrl:     settings.api_url || routing.api_url,
+              resVersion: RubyEventStore::VERSION
```
pawelpacana, line 94: I'd prefer flat structure here and `Browser.Elm.Main.init(flags: settings)`

ruby_event_store-browser/elm/src/index.js
piotrekbator: whoops

pawelpacana:
Thank you for the contribution! The essence is there, the details can be tweaked after merge 🙂

ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
piotrekbator: Yeah, that's what I was planning on doing but then, when I started thinking about where to put this attribute (on what element) I realized that there's no "root" element here. So I went with how e.g. Rails includes the csrf token (https://api.rubyonrails.org/classes/ActionView/Helpers/CsrfHelper.html#method-i-csrf_meta_tags).

Seemed reasonable, but I can change it if you'd prefer to have it other way.

piotrekbator:
I'll have the issues fixed in a moment, can I just update this PR?

pawelpacana:
> I'll have the issues fixed in a moment, can I just update this PR?

 Whatever works. I've invited you to maintainers team — things should come easier with write access

Issue #583, opened 2019-05-05
Browser (web interface) doesn't work due to missing assets


obahareth:
I have a new Rails app and I followed the instructions in [the tutorial](https://railseventstore.org/docs/browser/), and I get these errors:

```
ActionController::RoutingError (No route matches [GET] "/res/ruby_event_store_browser.css"):
```

```
ActionController::RoutingError (No route matches [GET] "/res/ruby_event_store_browser.js"):
```

pawelpacana:
Do you perhaps use git dependency in `Gemfile`? Assets are compiled and packaged in released gems.

obahareth:
Oh yes, you're absolutely right. I had to do that to try out the fix for #578. Is that fix released yet?

pawelpacana:
Fix hasn't been released yet but I guess that will eventually happen this week

PR #1187, opened 2021-09-11
Browser working from git source

pawelpacana:
It was never possible to use `rails_event_store` from git source and have a functioning browser:
#583. 

```ruby
gem 'rails_event_store', github: 'RailsEventStore/rails_event_store', branch: 'store-ci-assets'
```

We only version sources. Assets are bundled and baked  into released browser gem. 

Yet, with little help of:
- CI building assets out of every pushed commit
- S3 storing each of built assets and making them addressable by sha/name
- Cloudfront CDN making fetching assets quick
- Bundler letting us know that activated gem comes from git 

we're finally able to make this surprise thing of a past. Also unblock some early adopters 😀

That's mostly proof of concept. It works but needs rewrite, and actually building assets of all pushed commits, not only the last one.

Seeing is believing:

```
 $LOAD_PATH.select { |x| x.end_with? "ruby_event_store-browser/lib" }
=> ["/Users/mostlyobvious/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/rails_event_store-151d0dfbec24/ruby_event_store-browser/lib"]
```

```
http://d3iay4bmfswobf.cloudfront.net/151d0dfbec24/ruby_event_store_browser.css
```

Zrzut ekranu 2021-09-12 o 22 38 08

Zrzut ekranu 2021-09-12 o 22 54 06

What do you think?


jandudulski:
What's stopping us here?

pawelpacana:
> What's stopping us here?

Probably one more rebase since I had to resurrect separate CSS assets for [#1346] 🙃

I'd love to have it to experiment with some browser ideas on more production-like data!

pawelpacana:
not to myself: alias CDN domain

PR #1204, opened 2021-09-28
Experimenting with esbuild

pawelpacana:
Pros:
* less dependencies and up-to-date (postcss) dependencies — less problems with
  `npm audit` and hopefully lesser need to update them

Cons:
* watch mode seems to not work with elm source files (detects changes
  only in JS entrypoint) thus watch not ported :/
* bundle size 20kB larger (previously it was 60kB minified and so it is using [elm + uglifyjs](https://github.com/elm/compiler/blob/master/hints/optimize.md#scripts))
  https://discourse.elm-lang.org/t/elm-minification-with-esbuild/7143
* locking on quite old esbuild version since plugins were not updated


pawelpacana:
See #1208

PR #1237, opened 2021-11-12
Self-hosted runners

swistak35:


pawelpacana:
Closing — maybe we'll revisit the AWS part in the future

PR #1238, opened 2021-11-17
Require MFA for releases

jandudulski:
Read more: https://guides.rubygems.org/mfa-requirement-opt-in/


pawelpacana:
:+1:  if this enforces the practice we've been following for quite some time already

PR #1249, opened 2021-12-17
Add GAT to companies using RES section

gat-bryszard:
We're also using RES and we're not afraid to admit it publicly ;).


pawelpacana:
> We're also using RES and we're not afraid to admit it publicly ;).

I'd love to see it listed in your recruitment offers ;)

PR #1262, opened 2022-01-25
Fix initialization for ROM

katafrakt:
Currently initialization for ROM does not work because of number of things:

* `::RubyEventStore::ROM.configure` having been removed - I replaced it with a call to `::RubyEventStore::ROM.rom_container`, which I made public, but maybe a public proxy `.configure` would be better
* Wrong source migration path
* `ROM::SQL::RakeSupport.create_migration` trying to create output migration in a really weird path (in Ruby 3): 
```
rake aborted!
Errno::ENOENT: No such file or directory @ rb_sysopen - db/migrate/{:path=>"/Users/katafrakt/dev/palaver/db/migrate"}_create_ruby_event_store_tables.rb
/Users/katafrakt/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/rom-sql-3.5.0/lib/rom/sql/migration/migrator.rb:54:in `write'
```
* Non-existing require path in instructions

This PR fixes all four, making it possible to create migrations for ROM setup.

### DIFF
--- a/contrib/ruby_event_store-rom/lib/ruby_event_store/rom.rb
+++ b/contrib/ruby_event_store-rom/lib/ruby_event_store/rom.rb
@@ -31,8 +31,6 @@ module RubyEventStore
         end
       end
 
-      private
-
       def rom_container(adapter_name, database_uri, &block)
         if adapter_name.is_a?(::ROM::Configuration)
           ::ROM.container(adapter_name.tap(&block), &block)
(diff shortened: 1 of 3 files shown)


jandudulski (approved):

pawelpacana:
Thank you 🖤

pawelpacana:
I've invited you to RES collaborators. Feel free to improve any rough edges that you experience 🙂

PR #1267, opened 2022-01-27
Fix request_metadata lambda example in docs

pepawel:
This fixes TypeError "no implicit conversion of nil into Hash" exception


swistak35:
Looks good, I've verified that indeed this method needs to return a hash

PR #1294, opened 2022-03-23
Add RubyEventStore::Serializers::YAML as default serializer

### LINKED ISSUE #1260: Default serialization does not work in Ruby 3.1 for non-primitive values
Ruby 3.1 introduced Psych 4.0, which forces use of `safe_load` when calling `YAML.load` for non-core classes. I stumbled upon it because I use `BigDecimal` to store prices, which the land in the event. This was working fine in earlier Ruby versions:

```ruby
record = RubyEventStore::Record.new(event_id: "Abc", event_type: "test", data: BigDecimal("12.5"), metadata: {}, timestamp: Time.now,  valid_at: Time.now)
serialized = record.serialize(YAML)
serialized.deserialize(YAML)
#=> #<RubyEventStore::Record:0x00007fe900b0c868 @event_id="Abc", @data=0.125e2, @metadata={}, @event_type="test", @timestamp=2022-01-12 23:58:30.419385 +0100, @valid_at=2022-01-12 23:58:30.419387 +0100, @serialized_records={}>
```

However, the same does not work in 3.1:

```ruby
record = RubyEventStore::Record.new(event_id: "Abc", event_type: "test", data: BigDecimal("12.5"), metadata: {}, timestamp: Time.now,  valid_at: Time.now)
serialized = record.serialize(YAML)
serialized.deserialize(YAML)

Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/class_loader.rb:99:in `find': Tried to load unspecified class: BigDecimal (Psych::DisallowedClass)                                                                             
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/class_loader.rb:28:in `load'
	from (eval):2:in `big_decimal'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:79:in `deserialize'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:128:in `visit_Psych_Nodes_Scalar'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/visitor.rb:30:in `visit'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/visitor.rb:6:in `accept'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:35:in `accept'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:318:in `visit_Psych_Nodes_Document'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/visitor.rb:30:in `visit'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/visitor.rb:6:in `accept'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:35:in `accept'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych.rb:335:in `safe_load'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych.rb:370:in `load'
	from /Users/katafrakt/dev/github/rails_event_store/ruby_event_store/lib/ruby_event_store/serialized_record.rb:57:in `deserialize'
	from (irb):6:in `<main>'
	from /Users/katafrakt/.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	... 17 levels...
```

This could be simplified as:

```ruby
YAML.load(YAML.dump(BigDecimal("10"))) #=> exception in 3.1, works up to 3.0
YAML.unsafe_load(YAML.dump(BigDecimal("10"))) #=> 0.1e2
```

So, I think default serializer should be some adapter/proxy for a `YAML` class, which would call `unsafe_load` in its `load`, not just `YAML`.

katafrakt:
Up until now regular `YAML` module shipped with Ruby was used as a default serializer. However, with release of Ruby 3.1 this module started to use Psych 4 instead of Psych 3. This, in turn, resulted in a breaking change when non-primitive values were serialized and deserializes, as since version 4 Psych does "safe loading" by default, which disallows deserializing objects such as `BigDecimal` or `Time`.

As this breaking change also leaks into RES (old events would no longer be correctly deserialized afer upgrading to Ruby 3.1), a proxy serializer is introduced. It checks if `unsafe_load` method of `YAML` is present (which indicates Psych 4) and uses it if available, falling back to plain old `YAML.load` if it's not present.

Fixes #1260


eric-at-carbodylab:
I ran into this problem today when attempting to update all the things.
THIS is definitely needed.

pawelpacana:
Merging — mutant failures are expected. The Ruby/Psych version that mutant runs on does not allow it to figure out that such changes are needed. I'll deal with those before releasing (#1268). 


Thank you!

Issue #651, opened 2019-07-15
Unexpected behavior when using JSON/JSONB data type with JSON serializer 


scottbartell:
If you are using JSON/JSONB data types for data/metadata, you probably shouldn’t be using the `JSON` serializer.

It turns out doing so saves your JSON (event data/metadata) as a top level string which technically is valid JSON but isn’t very useful because you can’t query the event data using PG JSON operators out of the box.

I fixed this by implementing a `NullSerializer` which just implements`.dump` and `.load` and just returns the passed in values.

## Here's an example

Configure Rails Event Store using JSONB data types for `data` and `metadata` fields:
```console
$ bin/rails g rails_event_store_active_record:migration --data-type=jsonb
$ bin/rails db:migrate
```

Publish an event using `JSON` serializer:
```rb
irb(main):001:0> event_store = RailsEventStore::Client.new(
  mapper: RubyEventStore::Mappers::Default.new(
    serializer: JSON,
  )
)
irb(main):002:0> event = RailsEventStore::Event.new(data: { order_id: 123 })
irb(main):003:0> event_store.publish(event)
```

The event was stored but `data` and `metadata` are saved as strings:
```rb
irb(main):004:0> pp RailsEventStoreActiveRecord::Event.all
[#<RailsEventStoreActiveRecord::Event:0x00007f84771cdb00
  id: "42f9d3d7-d8a0-416e-842f-33e80370c019",
  event_type: "RubyEventStore::Event",
  metadata: "{\"timestamp\":\"2019-07-15 15:07:38 UTC\"}",
  data: "{\"order_id\":123}",
  created_at: Mon, 15 Jul 2019 15:07:38 UTC +00:00>]

irb(main):018:0> RailsEventStoreActiveRecord::Event.where("data -> 'order_id' = '123'").to_a
  RailsEventStoreActiveRecord::Event Load (0.5ms)  SELECT "event_store_events".* FROM "event_store_events" WHERE (data -> 'order_id' = '123')
=> []
```

```sql
$ bin/rails db
psql (10.4)
Type "help" for help.

development=# select * from event_store_events;
                  id                  |      event_type       |                   metadata                    |         data         |         created_at
--------------------------------------+-----------------------+-----------------------------------------------+----------------------+----------------------------
[...]

PR #1297, opened 2022-04-05
Additional value mapping (Hanami 2 integration)

swilgosz:
# Overview

Tweaks for Hanami 2 integration.

# Background

During configuring of Hanami 2 project to work with RES-ROM, I have encountered sequel problem. Command below accepts named arguments, so there is no way to call the command handler directly passing invalid parameters to it.

```ruby
cmd = Crm::RegisterCustomer.new(customer_id: customer_id, name: name)
command_bus.(cmd)
```

However, calling command bus with this command  returns an Sequel Error: Column not found (details below).

I have found that `CreateEvents` changeset receives Hash as a value for JSONB columns (data and metadata) in case of Hanami 2 integration, and String in Rails or tests.

To solve it, I added two additional MAP_VALUE rules.

```
module RubyEventStore
  module ROM
    module Changesets
      class CreateEvents < ::ROM::Changeset::Create
        relation :events

        map(&:to_h)
        map do
          # ...
          map_value   :data, ->(value) { value.class.name == "Hash" ? value.to_json : value }
          map_value   :metadata, ->(value) { value.class.name == "Hash" ? value.to_json : value }
          # ...
        end

        def commit
          relation.multi_insert(to_a)
        end
      end
    end
  end
end
```

It sems to solve the problem, but I am not happy with this as I believe, I did not find the source of the problem. I expect that some kind of value transformation was not triggered.

### Summary

I keep this PR as a monkey patch for now, until the proper fix is found, to make Hanami 2 integration possible. Help welcome!

### DIFF
--- a/contrib/ruby_event_store-rom/Gemfile
+++ b/contrib/ruby_event_store-rom/Gemfile
@@ -7,4 +7,7 @@ gem "mysql2", ">= 0.5.3"
 gem "pg", ">= 1.2.2"
 gem "ruby_event_store", path: "../../ruby_event_store"
 gem "sqlite3", "1.4.2"
-gem "activesupport"
+
+group :test do
+  gem "activesupport"
+end
--- a/contrib/ruby_event_store-rom/lib/ruby_event_store/rom/changesets/create_events.rb
+++ b/contrib/ruby_event_store-rom/lib/ruby_event_store/rom/changesets/create_events.rb
@@ -11,6 +11,8 @@ module RubyEventStore
           rename_keys timestamp: :created_at
           map_value   :created_at, ->(time) { Time.iso8601(time).localtime }
           map_value   :valid_at,   ->(time) { Time.iso8601(time).localtime }
+          map_value   :data, ->(value) { value.class.name == "Hash" ? value.to_json : value }
+          map_value   :metadata, ->(value) { value.class.name == "Hash" ? value.to_json : value }
           accept_keys %i[event_id data metadata event_type created_at valid_at]
         end
 


contrib/ruby_event_store-rom/Gemfile
```diff
@@ -7,4 +7,7 @@ gem "mysql2", ">= 0.5.3"
 gem "pg", ">= 1.2.2"
 gem "ruby_event_store", path: "../../ruby_event_store"
 gem "sqlite3", "1.4.2"
-gem "activesupport"
```
pawelpacana, line 10: Gemfile is purely for test. We're not using bundler groups at all anywhere in the  RES codebase. 

You can have a look on ROM dependencies at:
https://github.com/RailsEventStore/rails_event_store/blob/be89cf85a50461b2a4df5835e5b80c2672267b9f/contrib/ruby_event_store-rom/ruby_event_store-rom.gemspec#L29-L36

contrib/ruby_event_store-rom/lib/ruby_event_store/rom/changesets/create_events.rb
```diff
@@ -11,6 +11,8 @@ class CreateEvents < ::ROM::Changeset::Create
           rename_keys timestamp: :created_at
           map_value   :created_at, ->(time) { Time.iso8601(time).localtime }
           map_value   :valid_at,   ->(time) { Time.iso8601(time).localtime }
+          map_value   :data, ->(value) { value.class.name == "Hash" ? value.to_json : value }
+          map_value   :metadata, ->(value) { value.class.name == "Hash" ? value.to_json : value }
           accept_keys %i[event_id data metadata event_type created_at valid_at]
```
pawelpacana, line 16: Perhaps worth adding a test recreating the situation you've encountered?

pawelpacana:
Long story short:
* in RES we do not insist on one particular data/metadata column type — it can be binary (for YAML and Protobuf), json or jsonb
* this column type choice can decided on initial migration
* database engine must support this choice (i.e. default binary works on all: sqlite, mysql, postgres)
* AR repository is checked against all datatypes where RDBMS permits: https://github.com/RailsEventStore/rails_event_store/blob/master/.github/workflows/rails_event_store_active_record.yml#L86-L100
* each choice needs to backed by a serializer that transforms event payload into something appropriate for the column type in storage
* by default each data/metadata payload is turned into string (works well for binary columns)
* json/jsonb columns, when accessed via AR, need something different — Ruby hashes https://github.com/RailsEventStore/rails_event_store/issues/651 and in order to provide them a [NULL](https://github.com/RailsEventStore/rails_event_store/blob/d88bcd85baca97d110fdba89ee7e7662d2b84692/ruby_event_store/lib/ruby_event_store/null.rb) serializer [is used](https://github.com/RailsEventStore/ecommerce/blob/590685b4546ecfefe98fa0082ee910ddbe7af78d/rails_application/config/initializers/rails_event_store.rb#L20) 
* ROM adapter tests are parametrized against data types too: https://github.com/RailsEventStore/rails_event_store/blob/master/.github/workflows/ruby_event_store-rom.yml#L72-L86 and we seem to explicitly pass `JSON` serializer [there in tests](https://github.com/RailsEventStore/rails_event_store/blob/7f07ee3bf06aaa66ddac7d82325c353779bd461b/contrib/ruby_event_store-rom/spec/rom/event_repository_spec.rb#L8-L17) (turning hashes into strings), could be that Sequel behaviour is different than ActiveRecord's [one](https://github.com/RailsEventStore/rails_event_store/issues/651) and the advice we give in [docs](https://github.com/RailsEventStore/rails_event_store/blob/7f07ee3bf06aaa66ddac7d82325c353779bd461b/railseventstore.org/source/docs/v2/mapping_serialization.html.md.erb#L65-L76) would not apply

pawelpacana:
Can you first check that switching https://github.com/RailsEventStore/ecommerce/blob/6de6008878decfb4332a01e052202fb4a473bb4f/hanami_application/config/providers/event_repository.rb#L9 `::RubyEventStore::NULL to `::JSON` works as expected?

pawelpacana:
Loudly thinking — if there's a way to pass ActiveRecord data/metadata as strings for json/jsonb columns, that do not get additional to-string serialization, that would greatly reduce complexity around serailizers for JSON across all known repositories.

pawelpacana:
Btw. that SQL payload was the most valuable part of the exception as it shows data in raw shape as we insert it. Too bad it was trimmed.

```
Sequel::DatabaseError: PG::UndefinedColumn: ERROR:  column "customer_id" does not exist
LINE 1: ...VALUES ('d20c4621-3874-45d3-8a06-338dc6c969d8', (("customer_...
````

pawelpacana:
Verified: https://github.com/RailsEventStore/ecommerce/pull/110#discussion_r870638026

PR #1313, opened 2022-04-27
Replace ActiveRecord::Import#import for Rails 6 bulk insert

deepj:
Rails 5.2's EOL is 1.6.2022 so it would be nice to remove `ActiveRecord::Import` and replace it for Rails 6 bulk insert in the next RES. Yes, it's a breaking change so it maybe suitable for RES 3.x.

Tests passed locally (`make test`).


pawelpacana:
In fact this change was discussed internally and will be welcomed addition to breaking 3.x release :+1:

contrib/ruby_event_store-outbox/Gemfile.sidekiq_5_2
```diff
@@ -1,14 +0,0 @@
...
-
-
-gem 'ruby_event_store', path: '../../ruby_event_store'
-gem 'sidekiq', '5.2.9'
-gem 'pg', '1.2.2'
-gem 'mysql2', '0.5.3'
-gem 'influxdb', '~> 0.8.1', require: false
-gem 'childprocess'
-gem 'rails', '6.1.0'
-gem 'sqlite3', '1.4.2'
```
pawelpacana, line 14: Perhaps one gemfile too far?

deepj: Restored

pawelpacana:
> @pawelpacana haven't we agreed that removing support for EOLed rubies/rails won't need a major release? In that case we could merge that PR just after rails gets EOLed

We did, we may want to release 3.0 before 5.2 gets EOLed.

PR #1335, opened 2022-05-24
Support ActiveJob adapters with pre-deserialized events and NULL serializer

### LINKED ISSUE #1334: Decouple serialisation-related logic from AsyncHandler
Trying to use the async handler in my ActiveJob class like so: 

```
class MyEventProcessingJob < ApplicationJob
  prepend RailsEventStore::CorrelatedHandler
  prepend RailsEventStore::AsyncHandler.with(serializer: RubyEventStore::NULL)

  def perform(the_event)
  end
end
```

This works flawlessly in a regular setup: the `perform` method receives instances of the ActiveRecord-based event model.

But now it's not possible to use the job manually anymore: calling `MyEventProcessingJob.perform_now(some_event)` will raise `undefined method `symbolize_keys'` because that's not a method of the Rails model class.

The current implementation reads

```
    def self.with(event_store: Rails.configuration.event_store, serializer: YAML)
      Module.new do
        define_method :perform do |payload|
          super(event_store.deserialize(serializer: serializer, **payload.symbolize_keys))
        end
      end
    end
```

I propose moving the `**` and `symbolize_keys` somewhere else, probably into the serializer.

That way, when using the `NULL` serializer and calling perform with an event instance, the same event instance will end up in the job class.

milgner:
Depending on the configuration of the ActiveJob backend and the serializer used - i.e. the `NULL` serializer -, it is possible that
the argument to `perform` is already a full `Event` record. In that scenario, it can be used as-is without prior processing.

Closes #1334, even if the approach is different from how I understood the problem at first glance. But when taking a closer look at `Record`, `DeserializedRecord` and the transformation pipeline I understood better the way things are related and decided on a more straightforward solution and just detecting that specific situation.

### DIFF
--- a/rails_event_store/lib/rails_event_store/async_handler_helpers.rb
+++ b/rails_event_store/lib/rails_event_store/async_handler_helpers.rb
@@ -2,21 +2,32 @@
 
 module RailsEventStore
   module AsyncHandler
-    def self.with_defaults
-      with
-    end
+    class << self
+      def with_defaults
+        with
+      end
 
-    def self.with(event_store: Rails.configuration.event_store, event_store_locator: nil, serializer: RubyEventStore::Serializers::YAML)
-      Module.new do
-        define_method :perform do |payload|
-          event_store = event_store_locator.call if event_store_locator
-          super(event_store.deserialize(serializer: serializer, **payload.symbolize_keys))
+      def with(event_store: Rails.configuration.event_store, event_store_locator: nil, serializer: RubyEventStore::Serializers::YAML)
+        Module.new do
+          define_method :perform do |payload|
+            event_store = event_store_locator.call if event_store_locator
+            record = AsyncHandler.event_from_payload(payload, event_store, serializer)
+            super(record)
+          end
         end
       end
-    end
 
-    def self.prepended(host_class)
-      host_class.prepend with_defaults
+      def prepended(host_class)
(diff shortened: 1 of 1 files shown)


rails_event_store/lib/rails_event_store/async_handler_helpers.rb
```diff
@@ -2,21 +2,32 @@
 
 module RailsEventStore
   module AsyncHandler
-    def self.with_defaults
-      with
-    end
+    class << self
```
pawelpacana, line 5: I don't think this was necessary change. Changing formatting made it harder to notice what really changed.

rails_event_store/lib/rails_event_store/async_handler_helpers.rb
```diff
@@ -2,21 +2,32 @@
...
 
-    def self.prepended(host_class)
-      host_class.prepend with_defaults
+      def prepended(host_class)
+        host_class.prepend with_defaults
+      end
+
+      # @return [Event] an event as deserialized from the payload
+      def event_from_payload(payload, event_store, serializer)
+        return payload if payload.kind_of?(RailsEventStore::Event)
```
pawelpacana, line 26: We don't require Event objects in RES to be of this kind. It's likely they are but we accept anything that has the same interface as `RES::Event`.

Issue #958, opened 2021-01-03
Make browser dependent on Rack only


pawelpacana:
There's very little benefit of `sinatra` dependency and it may introduce dependency conflicts.

pawelpacana:
https://github.com/RailsEventStore/rails_event_store/issues/715

pawelpacana:
Scope: https://github.com/RailsEventStore/rails_event_store/blob/765a3ce6cb2af18d3f5a805c1512a1afbc657b88/ruby_event_store-browser/lib/ruby_event_store/browser/app.rb

pawelpacana:
Unfortunately this would involve reinventing some kind of router on top for `Rack::URLMap`.

pawelpacana:
https://github.com/mperham/sidekiq/commit/9ea167db16bf7f4cc9ee94967d0ea0b4dee6a7c5

PR #1341, opened 2022-06-04
Browser — drop sinatra dependency

pawelpacana:
https://github.com/RailsEventStore/rails_event_store/issues/958

* Rack::Static to serve compiled JS from public path
* Rack::Request for nice facade over Rack env
* lacking Rack::Protection, which can be re-added if neeeded
* case pattern matching seems simple and sufficient for such a tiny app that Browser is,
  good thing that it uses === and that plays nicely with Regexp,
* one drawback is that it is also quite verbose without that fancy DSL for regex-matching
  URL patterns behind the scenes
* a few rough edges like unnamed and a bit too wide (thus order dependent) regexes,
  decoding params and endless string-vs-symbol whack-a-mole

Examples of router implementation:
https://github.com/pjb3/rack-router
https://github.com/mperham/sidekiq/commit/9ea167db16bf7f4cc9ee94967d0ea0b4dee6a7c5

### DIFF
--- a/ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
+++ b/ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
@@ -63,8 +63,7 @@ module RubyEventStore
                  page: request.params["page"]
                )
         in "GET", %r{/(events/.*|streams/.*)?}
-          render_html(
-            ERB.new(<<~HTML).result_with_hash(path: routing.root_path, browser_settings: browser_settings(routing))
+          erb <<~HTML, path: routing.root_path, browser_settings: browser_settings(routing)
               <!DOCTYPE html>
               <html>
                 <head>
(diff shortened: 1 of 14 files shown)


ruby_event_store-browser/lib/ruby_event_store/browser/app.rb
```diff
@@ -1,98 +1,122 @@
...
-      configure { set :protection, except: :path_traversal }
+      def call(env)
+        request = Rack::Request.new(env)
+        routing = Routing.new(host || request.base_url, root_path || request.script_name)
+        event_store = event_store_locator.call
 
-      get "/api/events/:id" do
-        json Event.new(event_store: settings.event_store_locator, event_id: params["id"])
+        case [request.request_method, request.path]
+        in "GET", %r{/api/events/([^/]+)$}
```
jandudulski, line 50: Love it :heart_eyes:

PR #1349, opened 2022-06-20
Revert mutant downgrade to check if still an issue

pawelpacana:


mbj:
@pawelpacana This should be fixed by: https://github.com/mbj/mutant/pull/1342

PR #1354, opened 2022-07-01
Raise ArgumentError when trying to publish nil as event

jandudulski:
Relates to #1269

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/client.rb
+++ b/ruby_event_store/lib/ruby_event_store/client.rb
@@ -339,12 +339,16 @@ module RubyEventStore
     end
 
     def enrich_events_metadata(events)
+      raise ArgumentError, "Event cannot be `nil`" if events.nil?
+
       events = Array(events)
       events.each { |event| enrich_event_metadata(event) }
       events
     end
 
     def enrich_event_metadata(event)
+      raise ArgumentError, "Event cannot be `nil`" if event.nil?
+
       metadata.each { |key, value| event.metadata[key] ||= value }
       event.metadata[:timestamp] ||= clock.call
       event.metadata[:valid_at] ||= event.metadata.fetch(:timestamp)
--- a/ruby_event_store/lib/ruby_event_store/client.rb
+++ b/ruby_event_store/lib/ruby_event_store/client.rb
@@ -28,6 +28,8 @@ module RubyEventStore
     # @param expected_version [:any, :auto, :none, Integer] controls optimistic locking strategy. {http://railseventstore.org/docs/expected_version/ Read more}
     # @return [self]
     def publish(events, stream_name: GLOBAL_STREAM, expected_version: :any)
+      assert_nil_events(events)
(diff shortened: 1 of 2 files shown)


ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -339,12 +339,16 @@ def transform(events)
     end
 
     def enrich_events_metadata(events)
+      raise ArgumentError, "Event cannot be `nil`" if events.nil?
```
swistak35, line 342: Why here instead of at the top of `#publish` / `#append`?

jandudulski: Because here is the guard coercing to array. I'm not sure if this is the best place to all this operations (probably not). Do you have any specific arguments for moving assertions to publish/append? (the only that I have in mind is the rule to keep assertions as close public methods as possible but I'm fine with breaking it here)

If that makes sense I can think about refactoring in separate PR

swistak35: `the only that I have in mind is the rule to keep assertions as close public methods as possible`

And that's the one I agree with :) Especially that we are raising an `ArgumentError` which is raised when arguments passed to the method are wrong. Any benefit of keeping these in `enrich_` methods?

ruby_event_store/lib/ruby_event_store/client.rb
```diff
@@ -28,6 +28,8 @@ def initialize(
     # @param expected_version [:any, :auto, :none, Integer] controls optimistic locking strategy. {http://railseventstore.org/docs/expected_version/ Read more}
     # @return [self]
     def publish(events, stream_name: GLOBAL_STREAM, expected_version: :any)
+      assert_nil_events(events)
```
mbj, line 31: @pawelpacana, @jandudulski Why not change it around, instead to explicitly check some negative examples, why not ask `events` *is* an `Array`?

jandudulski: Because it can be just an event or array of events

pawelpacana:
Thank you!

ruby_event_store/lib/ruby_event_store/client.rb
mbj: Than check for is array or is an event. But not explicitly do a partial negative check. There are so many other invalid values you do not check for.

`false`, `true`, `Object.new`, the list of "incompatible" inputs is endless, the list of compatible ones finite.

pawelpacana: I would prefer allow list too but it is not that straightforward. I’m fine accepting Object.new if it responds to right methods.

jandudulski: Ah, now I got your point @mbj. I like this point of view, the issue is that `event` might be many things, but I'll figure it out.

PR #1411, opened 2022-09-29
Update pubsub.html.md

Dirichi:
CancelOrder should be CancelOrderService


pawelpacana:
Thank you for noticing this!

PR #1434, opened 2022-11-02
Outbox: automatically retry once on connection errors

swistak35:
I am well aware that the tests are crap for these, i.e. there are no unit tests covering specific sidekiq processor (mostly because there is only a sidekiq processor and it's hard to abstract). I plan to improve tests later. Currently the behaviour is similar, but accidental: the consumer get the error (like timeout error), logs it, and then process next batch, starting from the same point. That will improve "performance" though, meaning the data won't be unnecessarily fetched again (we don't expect it to change, ever, so it doesn't make sense).


swistak35:
Merging, but please review @porbas or @mpraglowski

PR #1435, opened 2022-11-02
Outbox: Add Tempo dynamically controlling the batch size

swistak35:
Rationale: The static batch size (let's say 100) is a problem when we sometimes have very very big messages. We fetch 100 messages, few are big ones, we crash (OOM). We start again, we fetch 100 messages, we crash again. Whereas we could (just after starting the outbox process fetch only one message, assuming that this could be the "very big one" because of which we crashed.


swistak35:
Merging, but please review @porbas or @mpraglowski

contrib/ruby_event_store-outbox/spec/tempo_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+module RubyEventStore
+  module Outbox
+    RSpec.describe Tempo do
+      specify "starts with 1" do
+        tempo = Tempo.new(100)
+
+        expect(tempo.batch_size).to eq(1)
+      end
+
+      specify "each batch size is bigger" do
```
mpraglowski, line 12: `each batch size is bigger` but testing only 2 first values ;)

contrib/ruby_event_store-outbox/spec/tempo_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        expect(second_value).to be > first_value
+      end
+
+      specify "maximum batch size cap is respected" do
+        max_batch_size = 100
+        tempo = Tempo.new(max_batch_size)
+        100.times { tempo.batch_size }
+
+        value101 = tempo.batch_size
+        expect(value101).to eq(max_batch_size)
```
mpraglowski, line 26: `maximum batch size cap is respected` but what is tested here is only the value is the same at some point - no test if this really the value provided as maximum batch size

contrib/ruby_event_store-outbox/spec/tempo_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      specify "batch_size stabilize" do
```
mpraglowski, line 29: quite similar to `maximum batch size cap is respected` test

contrib/ruby_event_store-outbox/lib/ruby_event_store/outbox/tempo.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def batch_size
+        @batch_size = next_batch_size
+      end
+
+      private
+
+      def next_batch_size
+        return 1 if @batch_size.nil?
+        [@batch_size * 2, @max_batch_size].min
```
mpraglowski, line 17: Maybe the multiplier value could be also configurable, or this is just making this more complex than needed? 
2 is good default  :)

contrib/ruby_event_store-outbox/spec/tempo_spec.rb
swistak35: You're right, "many", is three :joy: I will change that

contrib/ruby_event_store-outbox/spec/tempo_spec.rb
swistak35: hmm, way I see it, this test ensures that the maximum batch size is reached "at some point", and the second test ensures that the value stabilizes "and doesn't go any higher"

swistak35: ...but, it could be that the maximum batch size value is reached but it stabilizes at some other (higher) point. Right. I will improve these

contrib/ruby_event_store-outbox/lib/ruby_event_store/outbox/tempo.rb
swistak35: Nah, I will make it a constant, but the key is to:
- start with 1
- go quickly (exponentially) to the desired batch size, with exponential, 2 or 3 doesn't make almost any difference

PR #1436, opened 2022-11-02
Outbox readability improvements

swistak35:


contrib/ruby_event_store-outbox/lib/ruby_event_store/outbox/runner.rb
```diff
@@ -0,0 +1,40 @@
...
+        prepare_traps
+      end
+
+      def init
+        consumer.init
+      end
+
+      def run
+        while !@gracefully_shutting_down
+          was_something_changed = consumer.one_loop
```
mpraglowski, line 18: Consider renaming `consumer.one_loop` to `consumer.call` or something similar - consumer does not care about the "loop" runs now, this is handled by runner.

swistak35: @mpraglowski yup, my plan was to name it `Consumer#process`

PR #1445, opened 2022-11-09
Self hosted runners experiment

pawelpacana:


pawelpacana:
Salvaging parts of this PR already, will revisit in the future

PR #1480, opened 2022-12-02
Snapshot repository

pjurewicz:


aggregate_root/lib/aggregate_root/snapshot_repository.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+require 'base64'
+require 'ruby_event_store/event'
+
+module AggregateRoot
+  class SnapshotRepository
+    def initialize(event_store, interval = 1)
```
pawelpacana, line 7: I'd set default interval to 100. That's arbitrary value but also a default fot batched reads. Then I'd explicitly pass 1 to SnapshotRepository in specs.

aggregate_root/lib/aggregate_root/snapshot_repository.rb
```diff
@@ -0,0 +1,75 @@
...
+      raise ArgumentError, 'interval must be greater than 0' unless interval > 0
```
pawelpacana, line 8: It probably makes sense to ensure it is `Integer(interval)`

aggregate_root/lib/aggregate_root/snapshot_repository.rb
```diff
@@ -0,0 +1,75 @@
...
+      Marshal.load(snpashot_event.data.fetch(:marshal))
+    end
+
+    def snapshot_stream_name(stream_name)
+      "#{stream_name}_snapshots"
+    end
+
+    def time_for_snapshot?(aggregate_version, published_events)
+      rest = (aggregate_version + 1) % interval
+      published_events > rest
```
pawelpacana, line 72: For some reason it is difficult to imagine how this works. Perhaps it can be written differently?

aggregate_root/spec/snapshot_repository_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+        include AggregateRoot
+
+        def initialize(uuid)
+          @status = :draft
+          @uuid = uuid
+        end
+
+        attr_accessor :status
+
+        def apply_order_created(_)
```
pawelpacana, line 50: `on` methods are preferred since we've scheduled dropping `apply_event_type` meta

aggregate_root/spec/snapshot_repository_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+      order.apply(order_expired)
+      repository.store(order, stream_name)
+      expect_snapshot(order_expired.event_id, 1, Marshal.dump(order))
+    end
+
+    specify "restoring snapshot" do
+      order = order_klass.new(uuid)
+      repository = AggregateRoot::SnapshotRepository.new(event_store)
+
+      apply_and_publish(order, order_created, order_canceled)
```
pawelpacana, line 104: Could we lean more on aggregate here over using apply+publish helper? Perhaps we'd need to adjust snapshot interval but it would be less brittle and closer to how people use it.

```
order.create
order.cancel
repository.store(order)
```

aggregate_root/spec/snapshot_repository_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    def apply_and_publish(order, *events)
+      events.each do |e|
+        order.apply(e)
+        event_store.publish(e, stream_name: stream_name)
+      end
+    end
+
+    def publish_snapshot(order, last_event_id, version)
+      event_store.publish(
+        AggregateRoot::SnapshotRepository::Snapshot.new(
```
pawelpacana, line 143: Works but seems brittle to changes in the aggregate itself. Shouldn't we expect a snapshot being recorded after interval and assume this is unecessary detail?

pawelpacana:
Cases to verify:

* What happens when an aggregate is renamed after snapshot is recorded? Not an unusual thing to refactor. Does it load, does it crash, does it fallback to non-snapshot load?

* What happens when we dump non-primitive values, i.e. more complex objects or objects from libraries? Does it mean fragility when loading marshalled data?

* Can we make the aggregate load even if loading last snapshot fails?

* Case of an aggregate producing more than a single event as a result of its action, i.e. one event in aggregate's stream, threshold of 2 and applying 2 events as a result of an action

casiodk:
Hey, great work. Is there some documentation on how to use this :)

lukaszreszke:
@casiodk the docs are not yet ready. For now we have a blog post https://blog.arkency.com/speed-up-aggregate-roots-loading-with-snapshot-events/ and repo with examples https://github.com/RailsEventStore/snapshots

casiodk:
Awesome, I will check it out :)

PR #1482, opened 2022-12-02
Slightly improve navigating around documentation

pawelpacana:
- sticky header across the whole site
- sticky search, always visible in the viewport
- sticky sidebar to scroll independently to main documentation content
- nuked footer, it was problematc when combined with sticky sidebar;
  parts of it that contain links to DDR book and DDD workshops will come back in different form


pawelpacana:
https://deploy-preview-1482--railseventstore.netlify.app/docs/v2/install/

Issue #1439, opened 2022-11-04
Rename RailsEventStoreActiveRecord into RubyEventStore::ActiveRecord


swistak35:
Creating an issue for something to track before 3.0. I'm not sure it makes sense to start working on the code before we decide we want to finally go into 3.x :)

pawelpacana:
https://github.com/RailsEventStore/rails_event_store/discussions/1147

pawelpacana:
Correct me if I'm wrong. As of now:
- `rails_event_store-2.6.0` depends on `rails_event_store-2.6.0`
- `rails_event_store-master` depends on `ruby_event_store-active_record-master`

We did it because we control this dependency and replaced internally all invocations of `RailsEventStoreActiveRecord` with `RubyEventStore::ActiveRecord`.

We're also providing `rails_event_store_active_record` gem that:
- has dependency to `ruby_event_store-active_record`
- aliases `RubyEventStore::ActiveRecord` as `RailsEventStoreActiveRecord`
- warns about rename  in post-install message and when requiring `rails_event_store_active_record`

The only reason this aliased gem exists is for someone directly depending on it (which we don't control).

There's also the case of documentation pointing to `RailsEventStoreActiveRecord`, which we don't want to change in 2.x. Also someone might have customized `RailsEventStoreActiveRecord` so this const being stil present is fair expectation.

Should we then keep `rails_event_store` to depend on `rails_event_store_active_record`, being a wrapper on `ruby_event_store-active_record`? Do we hide deprecation warnings? Do we need to mention this deprecation in the docs?

@swistak35 @lukaszreszke

swistak35:
Well, if we want to provide backwards compatibility, as far as I understand:
1) we should still have a dependency on `rails_event_store_active_record`
2) we should remove deprecation warnings
3) with 3.0 release, we change the dependency to `ruby_event_store-active_record`. It would be nice (but not mandatory) to also release 3.0 for both `ruby_event_store-active_record` and `rails_event_store_active_record` (and 3.0 being the last version of this one). Then people not using rails_event_store could make gradual change (since they have to explicitly specify `rails_event_store_active_record` if they use it)

I don't see other reason since we can't say "either that gem or this gem" in gemspec, or "if we have ruby_event_store-active_record, then we rails_event_store_active_record is no longer dependency".

[...]

PR #1493, opened 2022-12-08
RailsEventStore and contribs dependency on RubyEventStore::ActiveRecord

lukaszreszke:
https://github.com/RailsEventStore/rails_event_store/issues/1439


lukaszreszke:
rel: https://github.com/RailsEventStore/rails_event_store/issues/1439

PR #1511, opened 2022-12-23
Bump RubyEventStore::ActiveRecord migration version to 5

lukaszreszke:
* Generated reference schema dumps for sqlite, postgres, mysql5 and mysql8
* The schema dump is used for assertions
* Bumped defaults of RES::AR templates with the goal that the output schema stays the same
* Didn't touch contrib 
* I don't like the way the test is shaped. However, the reason for that is the fact that we're testing multiple configurations which output differs. Willing to take any advice to refactor this part.

### DIFF
--- a/rails_event_store/spec/rails_migration_generator_spec.rb
+++ b/rails_event_store/spec/rails_migration_generator_spec.rb
@@ -23,7 +23,7 @@ module RailsEventStore
     end
 
     it "uses particular migration version" do
-      expect(subject).to match(/ActiveRecord::Migration\[4\.2\]$/)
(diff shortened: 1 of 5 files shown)


pawelpacana (commented):
That's mostly useless for test purpose.

Schema is described with DSL. The meaning of this DSL is slightly changing with newer versions of ActiveRecord. Maybe it is good enough to assume we'll always compare against `ActiveRecord::Schema[7.0]`.

I'd rather compare against raw SQL structure. Which means we'll need to compare against 3 SQL structures. Which means we'll have to take a step back and implement such verification first.

lukaszreszke:
Ok. I'll try a different approach.

PR #1514, opened 2023-01-04
Event serialization formats refactor

lukaszreszke:
An attempt to improve serialization & mappers sections of documentation


railseventstore.org/source/docs/v2/mapping_serialization.html.md.erb
```diff
@@ -75,6 +77,8 @@ Rails.application.configure do
 end
 ```
 
+Using the `RubyEventStore::NULL` serializer will prevent the event store from serializing the event data and metadata. This is necessary because the Postgres database will handle the serialization.
```
fidel, line 80: To be more precise, ActiveRecord will handle the serialization before putting the data into the database. And will do otherwise when reading. Database itself expect data to be json already.

railseventstore.org/source/docs/v2/mappers.html.md.erb
```diff
@@ -0,0 +1,127 @@
...
+title: Mappers
+---
+
+Mapper is defined as a pipeline of transformations that transforms domain event object into record and back from record to domain event. Mappers are useful when you have to encrypt your events data, or you’d like to access events using both strings nad symbols.
+
+## Available mappers
+There is set of available mappers that you can use out of the box:
+
+- [`RubyEventStore::Mappers::Default`](https://github.com/RailsEventStore/rails_event_store/blob/master/ruby_event_store/lib/ruby_event_store/mappers/default.rb)
+  - works with `RubyEventStore::Event` or `RailsEventStore::Event`
```
fidel, line 11: It's repeated for every mapper. Maybe we could state that once, at the beginning.

lukaszreszke: It is repeated for every mapper besides the protobuf one. I am assuming that might be because it has slightly different `event_type` definition.

```    
class Proto < RubyEventStore::Event
      def event_type
        data.class.descriptor.name
      end
    end
 ```

fidel: Good point, so maybe a table would be a more readable solution, e.g. like here: https://blog.arkency.com/comparison-of-approaches-to-multitenancy-in-rails-apps/

lukaszreszke: I'll check this one, thanks 👍

PR #1515, opened 2023-01-04
Skip AR serialization for json(b) columns on postgres

### LINKED ISSUE #1478: ActiveRecord and the need of non-serialized data and metadata for JSONB
It was discovered that for ActiveRecord to fully utilize JSONB column, the content passed for `data` and `metadata` columns must not be serialized. When the content was first serialized in Mapper, then the Repository was further serializing it as a-string-in-json in the database: https://github.com/RailsEventStore/rails_event_store/issues/651

Thus we strongly advise in the docs to use `NULL` serializer for the repository in such cases.

I wonder if there's a way to tell ActiveRecord: here's the raw payload, do not process it. This will simplify a lot: we would uniformly use the serializer regardless of underlying datatype in the database and we will drop a lot of comments in the documentation. Making things easier for end users. The task is to research this.

fidel:
Resolves #1478


msencenb:
New RES user here (so far, very impressed!). Yesterday I ran the existing Rails app template:

`bin/rails app:template LOCATION=https://railseventstore.org/new`

That template outputs the event store to JSON Client:

`Rails.configuration.event_store = RailsEventStore::JSONClient.new`

Today, I happened to stumble on [this page in the docs](https://railseventstore.org/docs/v2/mapping_serialization/#configuring-for-postgres-json-b-data-type), which describes using a null serializer when using jsonb columns:

`RailsEventStoreActiveRecord::EventRepository.new(serializer: RubyEventStore::NULL)`

Is it still best practice to use the null serializer with postgres jsonb? Or did this commit make the JSONClient skip AR serialization by default and that page just has old documentation?

fidel:
Hello @msencenb! 

Thanks for reaching out. 

The template correctly came up with `JSONClient.new` which [sets up the repository with appropriate serializer](https://github.com/RailsEventStore/rails_event_store/blob/219d609cd8fd53b1570f13283c9a275840a171c8/rails_event_store/lib/rails_event_store/json_client.rb#L7) under the hood, as you already noticed in the different part of the documentation.

> Is it still best practice to use the null serializer with postgres jsonb?

Currently yes, but that's what `JSONClient` does for you, specifically. It's enriched client, designed specifically for users who want to use superpowers of json(b) in events.

Upcoming RES 3.0 will bring some changes in that area, but sticking with `JSONClient` will make it invisible to you. 

The `NULL` serializer trick was done since `ActiveRecord` does its own serialization based on type of the column and combining it with `JSON` serialization in repository gave unwanted results like double serialization.

If you're curious how it will be done in RES 3.0, [here's the PR](https://github.com/RailsEventStore/rails_event_store/pull/1515) which disables serialization on `ActiveRecord` model level and enables one in the repository.

But as I said before, no action is required from you, code created by the template is completely legit.

msencenb:
Ah perfect! I should have looked at that JSONClient code to see what was going on. Thanks for the quick response—all of that makes sense and sounds great.

PR #1518, opened 2023-01-12
fix(activerecord): use created_at for fallback in #as_of when valid_a…

alexdreher:
fix(activerecord): use created_at for fallback in #as_of when valid_at is nil

We encountered an issue when using bi-temporal and non-bi-temporal events with a client reader.

1. Event timestamp: 6.minutes.ago
2. Event timestamp: 3.minute.ago
3. Event valid_at: 5.minutes.ago
4. Event valid_at: 4.minutes.ago

`event_store.read.as_at` gives the expected order (`ORDER BY created_at`): 1, 2, 3, 4

But `event_store.read.as_of`

Expected result: 1, 3, 4, 2
Actual result: 3, 4, 1, 2

Why? Because `valid_at` of non-bi-temporal events 1 and 2 is `NULL` in database, and it only gets `ORDER BY valid_at`, so these events get either prepended or appended in result list.

The fix uses COALESCE, to fall back to `created_at`, when `valid_at` is null.

### DIFF
--- a/ruby_event_store-active_record/lib/ruby_event_store/active_record/event_repository_reader.rb
+++ b/ruby_event_store-active_record/lib/ruby_event_store/active_record/event_repository_reader.rb
@@ -116,7 +116,7 @@ module RubyEventStore
         when :as_at
           stream.order("#{@event_klass.table_name}.created_at #{order(spec)}")
         when :as_of
-          stream.order("#{@event_klass.table_name}.valid_at #{order(spec)}")
+          stream.order(Arel.sql("COALESCE(#{@event_klass.table_name}.valid_at, #{@event_klass.table_name}.created_at) #{order(spec)}"))
         else
           stream
         end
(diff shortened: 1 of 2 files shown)


fidel:
Hi @alexdreher! 

Thanks for the pull request, we've wrote a spec which reproduces the issue you've mentioned: fa484a5256c857832327b19f4e0e029a27c3bfbc

We'll follow–up on your PR tomorrow.

alexdreher:
Hi @fidel,

thanks a lot for adding the spec, because I was unsure where to place it exactly. 

Makes sense to put it in the scope of "valid-at storage optimization", since I was also wondering, why `valid_at` is dropped, when you save an event with `timestamp = valid_at`, which was my first naive attempt to bypass this issue. I only saw this by looking at the records in the database and afterwards in the specs.

Thank you for following up!

fidel:
Thanks for the fix @alexdreher, please expect new gem version to be released soon.

lukaszreszke:
Hi @alexdreher,

We have just released version 2.8.0 with your fix.

Thanks for contributing.

PR #1521, opened 2023-01-16
fix(activerecord): #as_of filter by time

alexdreher:
Using bi-temporal event sourcing with `valid_at` the order of the stream when using `#as_of` was fixed with https://github.com/RailsEventStore/rails_event_store/pull/1518 

I encountered another bug when using `#as_of` in combination with time filtering e.g. `older_than(time)`

`event_store.read.stream('stream').as_of.older_than(time).backward.limit(1).first`

Expected: Get the **valid** event before a given point in time.
Actual: The last event _created_ before that time.

Why? The following methods in `event_repository_reader.rb` use only `created_at` for filtering by time:

`older_than(time)`
`older_than_or_equal(time)`
`newer_than(time)`
`newer_than_or_equal(time)`

This fix uses `COALESCE(valid_at, created_at)` when filtering by time using `#as_of`.

### DIFF
--- a/ruby_event_store-active_record/lib/ruby_event_store/active_record/event_repository_reader.rb
+++ b/ruby_event_store-active_record/lib/ruby_event_store/active_record/event_repository_reader.rb
@@ -163,20 +163,28 @@ module RubyEventStore
         )
       end
 
+      def time_comparison_field(specification)
+        if specification.time_sort_by_as_of?
+          "COALESCE(#{@event_klass.table_name}.valid_at, #{@event_klass.table_name}.created_at)"
+        else
+          "#{@event_klass.table_name}.created_at"
+        end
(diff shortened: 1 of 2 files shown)


fidel:
Another nice catch @alexdreher, thanks!

We'll release `v2.8.1` today then ;)

fidel:
And there it is: [v2.8.1](https://github.com/RailsEventStore/rails_event_store/releases/tag/v2.8.1)

alexdreher:
Thank you very much, that was really fast 🎉

fidel:
@alexdreher I've invited you to the maintainers team, it should be easier to contribute in the future, e.g. by creating PRs directly in the RES repository.
Thanks again!

PR #1555, opened 2023-01-23
Fix deep stringify keys in serializer for sidekiq scheduler

alexdreher:
Deep stringify the keys for sidekiq scheduler, so that sidekiq does not complain when using `RubyEventStore::NULL` serializer.


fidel:
Hey @alexdreher, 

could elaborate a bit on why you want to use `RubyEventStore::NULL` within `SidekiqScheduler`?

From our perspective, it's better to rely on explicit serializer for `data` and `metadata`. So that we pass string values in a flat hash to sidekiq. Thus we suggest `RubyEventStore::Serializers::YAML` or simply `JSON` as a serializer, so 3rd party (sidekiq) won't decide on how to handle serialization/deserialization. 

There's an added benefit of doing so: serialization/deserialization would occur just once when using the same serializer for both scheduler and repository.

PR #1561, opened 2023-01-30
RES 3.0 integration

pawelpacana:


lukaszreszke:
> moving EventClassRemapper to ruby_event_store-transformations gem in case someone relied on it
done

Issue #1552, opened 2023-01-20
Improve RES projections


pawelpacana:
- https://arkency.slack.com/archives/C7B95EW3V/p1669453490523099
- https://arkency.slack.com/archives/C7B95EW3V/p1619529649198000

PR: https://github.com/RailsEventStore/rails_event_store/pull/1463

PR #1565, opened 2023-02-01
Fix projections with bi-temporal event stream

### LINKED ISSUE #1550: RubyEventStore::Projection does not yield events in deterministic order when using multiple streams
Given I have some events in my database:

```
{id: 1, data: {...}, timestamp: "2023-01-19Z01:01:01"} on stream "A"
{id: 2, data: {...}, timestamp: "2023-01-19Z01:01:02"} on stream "B"
{id: 3, data: {...}, timestamp: "2023-01-19Z01:01:03"} on stream "A"
```

When I run a projection like so: `RubyEventStore::Projection.from_stream(%w(A B))` the events will be returned in order: `1,3,2`.
However, if I switch the order of the streams: `RubyEventStore::Projection.from_stream(%w(B A))` the events will be returned in order: `2,1,3`.

My expectation is that the order of the events should be in the order of the data, regardless of streams. In other words, the order of the stream names in the array should not affect the order in which the events are returned.

Example test in test is added here: https://github.com/bertg/rails_event_store/commit/bcc2cabff5ffcf2af488181b3cb44eda4ef1181b, and included as example here:

```ruby
specify "reduce events from many streams in order" do
  event_store.with_metadata(timestamp: Time.utc(2018, 1, 1, 1, 1, 1)) {
    event_store.append(MoneyDeposited.new(data: { order: 1 }), stream_name: "Customer$1")
  }
  event_store.with_metadata(timestamp: Time.utc(2018, 1, 1, 1, 1, 2)) {
    event_store.append(MoneyDeposited.new(data: { order: 2 }), stream_name: "Customer$1")
  }
  event_store.with_metadata(timestamp: Time.utc(2018, 1, 1, 1, 1, 3)) {
    event_store.append(MoneyDeposited.new(data: { order: 3 }), stream_name: "Customer$2")
  }

  account_balance =
    Projection
      .from_stream(%w[Customer$2 Customer$1])
      .init(-> { { order: nil } })
      .when(MoneyDeposited, ->(state, event) {
        puts event.timestamp
        state[:order] = event.data[:order]
      })
      .run(event_store)
  expect(account_balance).to eq(order: 3)
end
```

This fails with:

```
  1) RubyEventStore::Projection reduce events from many streams in order
     Failure/Error:

       expected: {:order=>3}
            got: {:order=>2}
```

alexdreher:
Projections don't play with bi-temporal event sourcing at all.

I followed the discussions and tickets from RES 3.0 ( #1463 + #1552 + #1561) and I know, that this will become obsolete, when the "Projections redesigned" (https://github.com/RailsEventStore/rails_event_store/pull/1561/commits/f76c43dc4bc0ed1a08446c969dc257dfab16a8e4) is merged.

Then we can finally use:

```ruby
  Projection
    .new(0)
    .on(MoneyDeposited) { |state, event| state += event.data[:amount] }
    .on(MoneyWithdrawn) { |state, event| state -= event.data[:amount] }
    .call(event_store.read.stream(stream_name).from(starting.event_id).in_batches(2).as_of)
```

But for now, I would appreciate this fix, if there is nothing against it. 

Please have a look at the changed specs, as I don't know if we really need `time_sort_by=nil`.

When using non-bi-temporal events (without `valid_at`) the projection is now ordered by `timestamp` (but sadly, this doesn't fix #1550)

### DIFF
--- a/ruby_event_store/lib/ruby_event_store/projection.rb
+++ b/ruby_event_store/lib/ruby_event_store/projection.rb
@@ -76,7 +76,7 @@ module RubyEventStore
     end
 
     def read_scope(event_store, stream, count, start)
-      scope = event_store.read.in_batches(count)
+      scope = event_store.read.in_batches(count).as_of
       scope = scope.of_type(handled_events)
       scope = scope.stream(stream) if stream
       scope = scope.from(start) if start
--- a/ruby_event_store/spec/projection_spec.rb
+++ b/ruby_event_store/spec/projection_spec.rb
@@ -209,6 +209,28 @@ module RubyEventStore
       expect(balance).to eq(total: 6)
     end
 
+    specify "all events from the stream must be read in valid order (starting from given event)" do
+      event_store.append(
+        [
+          MoneyDeposited.new(data: { amount: 10 }, metadata: { valid_at: Time.new(2020, 1, 7) }), # this should be included
+          starting = MoneyWithdrawn.new(data: { amount: 2 }, metadata: { valid_at: Time.new(2020, 1, 4) }),
+          MoneyDeposited.new(data: { amount: 4 }, metadata: { valid_at: Time.new(2020, 1, 5) }),
+          MoneyWithdrawn.new(data: { amount: 3 }, metadata: { valid_at: Time.new(2020, 1, 6) }),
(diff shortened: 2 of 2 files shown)


alexdreher:
Makes sense and is another good reason for the redesign of projections.

Thanks @pawelpacana for the monkey patch, it is working as expected.

PR #1575, opened 2023-02-07
Extend profile to work for mysql

lukaszreszke:
Since mysql supports json format now, profile should take that into consideration and generate JSONClient instead of regular one.

Should be merged once we agree that JSONClient is a good default for MySQL with json columns

### DIFF
--- a/APP_TEMPLATE
+++ b/APP_TEMPLATE
@@ -7,11 +7,21 @@ class Profile
   private attr_reader :options, :location
 
   def client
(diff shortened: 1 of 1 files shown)


pawelpacana:
I keep forgetting why we haven't merged this. My biggest fear is that this PR sets a new default for new installs and we don't have enough data points from production use of MySQL + JSON + RES.

mostlyobvious:
We're now in position to deprecate JSONClient — blockers in ActiveRecord should be gone by now. This code will no longer be needed or nor helpful.

Issue #630, opened 2019-06-27
Consider rubocop to check for warnings 


pawelpacana:
At the moment we rely on particular test to check if the code does not generate any Ruby warning:

```ruby
module RubyEventStore
  RSpec.describe 'no warnings' do
    specify do
      expect(aggregate_root_warnings).to eq([])
    end

    def aggregate_root_warnings
      warnings.select { |w| w =~ %r{lib/aggregate_root} }
    end

    def warnings
      `ruby -Ilib -w lib/aggregate_root.rb 2>&1`.split("\n")
    end
  end
end
```

While it mostly works, it does not exercise some code paths. It came to my attention that running `rubocop -l` might be an alternative worth checking.

The code that triggered this debate was:
```ruby
raise(ArgumentError, "Anonymous class is missing name") if name.start_with? "#<Class:"
```
It results in:
```
/Users/pawelpacana/Code/rails_event_store/aggregate_root/lib/aggregate_root.rb:20: warning: string literal in condition
```

It appeared in test output while mutating. Neither our spec, nor rubocop found it.

The task is to evaluate usefulness of Rubocop in this aspect of its functionality and figure out how to best combine with our CI.

Related:
https://docs.rubocop.org/en/stable/basic_usage/#2-rubocop-as-a-replacement-for-ruby-w

PR #1576, opened 2023-02-07
Rubocop strictly as a linter

pawelpacana:
Some interesting results that make us consider using it.
First step is to resolve existing warnings, or mute them.
Next: add this lint as a step in the CI and Makefile tooling.

https://github.com/RailsEventStore/rails_event_store/issues/630

Co-authored-by: Łukasz Reszke <lukaszreszke93@gmail.com>
Co-authored-by: Szymon Fiedler <szymon.fiedler@gmail.com>

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,19 +1,19 @@
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.7
+  SuggestExtensions: false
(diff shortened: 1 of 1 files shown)


pawelpacana:
After reviewing reported offenses it was decided that that our current warning-detection mechanism works well. Bringing Rubocop introduces far more noise than signal. And currently even the signal is very weak — false negatives.

PR #1628, opened 2023-04-19
Add setup example for ruby_event_store-rom

leoarnold:


fidel:
Hey @leoarnold! 

Thanks for your contribution. We would prefer to have it in the official documentation, rather than in `README`. Did you consider improving the [docs here](https://railseventstore.org/docs/v2/without_rails/#rom-sequel) (if they miss anything)?

mostlyobvious:
Thank you 🙏

PR #1646, opened 2023-07-20
Add foreign keys

pjurewicz:


ruby_event_store-active_record/spec/migration_test_spec.rb
```diff
@@ -140,7 +144,7 @@ def data_type_to_mysql_type(data_type)
 
       def pg_show_create_table(table_name)
         db_config = ::ActiveRecord::Base.connection_db_config
-        <<~RESULT.strip
+        re = <<~RESULT.strip
```
lukaszreszke, line 151: should we keep the `re` variable? seems like it's unused

PR #1706, opened 2023-11-12
Fix snapshot_repository aggregate state loading from snapshot

DeVeLo:
Marshal.load creates new object with brand new object id. Current implementation breakes events publishing after first snapshot is created.

### DIFF
--- a/aggregate_root/lib/aggregate_root/snapshot_repository.rb
+++ b/aggregate_root/lib/aggregate_root/snapshot_repository.rb
@@ -26,7 +26,11 @@ module AggregateRoot
       query = event_store.read.stream(stream_name)
       if last_snapshot
         begin
-          aggregate = load_marshal(last_snapshot)
+          loaded_aggregate = load_marshal(last_snapshot)
+          loaded_aggregate.instance_variables.each do |var|
+            value = loaded_aggregate.instance_variable_get(var)
+            aggregate.instance_variable_set(var, value)
+          end
         rescue NotRestorableSnapshot => e
           error_handler.(e)
         else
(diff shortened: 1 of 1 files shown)


pjurewicz:
Hello @DeVeLo,
thank you for your contribution!
Can you document with a simple test the issue this PR is resolving?

DeVeLo:
Hello @pjurewicz, when I find a free moment. Meanwhile, I've found another minor error, which I will post here for correction when I have time. It is that if we emit 2 consecutive events in the aggregator (as part of handling one command), the next event will end with a 500 and a version mismatch error.

pjurewicz:
I'm also curious under what circumstances the JSON is malformed. Can you provide an example?
We cannot introduce base64 encoding just like that as it is a breaking change.
Our snapshot repository is still an experimental feature, and we value all feedback that we receive.

DeVeLo:
As it was said, snapshot implementation is an experimental feature and in my opinion is unusable in current form. Of course, maybe in some specific environment it works just fine. But, in my case, when I tried to emit snapshot it throws an exception "illegal/malformed utf-8". Which is a result of marshaling an object. Maybe this is related to db engine, I use PostgreSQL 15.

pjurewicz:
Right, marshal dumps would not work with JSON events serializer without introducing encoding. However, you can have any other [configured](https://railseventstore.org/docs/v2/mapping_serialization/) that accepts ASCII-8 encoded string

DeVeLo:
Anyway, the solution does not work out of the box for PostgreSQL and jsonb. Any constructive conclusions regarding this PR?

pjurewicz:
We won't introduce base64 encoding as a scope of this PR. We need a migration strategy for other users that is not covered by this PR.
According to the first commit, I miss the test helping understand what this change fixes and preventing regression.

PR #1710, opened 2023-12-06
Prefer includes over preload

pjurewicz:


pawelpacana:
I'm missing SQL query assertions for the optimization part and I think mutant misses them too.

Issue #1708, opened 2023-11-29
Projection does three (3) queries where it could be doing one (1)


Bertg:
Given I have the Following projection:

```ruby
RubyEventStore::Projection
  .from_stream("STREAM_NAME")
  .init(-> { { count: 0 } })
  .when(SomeEvent, ->(state, event) { state[:c] +=1 })
  .run(Rails.configuration.event_store)
```

Then I see the following queries:

```
RubyEventStore::ActiveRecord::EventInStream Load (2.7ms)  SELECT "event_store_events_in_streams".* FROM "event_store_events_in_streams" INNER JOIN "event_store_events" ON "event_store_events"."event_id" = "event_store_events_in_streams"."event_id" WHERE "event_store_events_in_streams"."stream" = $1 AND "event_store_events"."event_type" = $2 ORDER BY "event_store_events_in_streams"."id" ASC LIMIT $3  [["stream", "STREAM_NAME"], ["event_type", "SomeEvent"], ["LIMIT", 100]]
RubyEventStore::ActiveRecord::Event Load (0.5ms)  SELECT "event_store_events".* FROM "event_store_events" WHERE "event_store_events"."event_id" = $1  [["event_id", "9f75375d-3df3-410b-8f51-bdd33ba28912"]]
RubyEventStore::ActiveRecord::EventInStream Load (0.9ms)  SELECT "event_store_events_in_streams".* FROM "event_store_events_in_streams" INNER JOIN "event_store_events" ON "event_store_events"."event_id" = "event_store_events_in_streams"."event_id" WHERE "event_store_events_in_streams"."stream" = $1 AND "event_store_events"."event_type" = $2 AND (event_store_events_in_streams.id > 373) ORDER BY "event_store_events_in_streams"."id" ASC LIMIT $3  [["stream", "STREAM_NAME"], ["event_type", "SomeEvent"], ["LIMIT", 100]]
```

I identify these as follows:

1. Fetch events based on the Projection conditions
2. Fetch the events by id from the result of the previous condition
3. Fetch the second page of events based on the Projection conditions 

It is my contention that this should be 1 query.

The first two queries should be combined. There is no reason that this should be done in 2 steps. Either the data can be fetched in a the first query or the queries should be combined in some sort of subquery.

The third (last) query is not needed. We know the limit is 100, and we received less than 100 records from the database. There is no need to go to the database again. 

[...]

PR #1711, opened 2023-12-06
batch enumerator optimization

pjurewicz:
no need to fetch next batch if the previous one is smaller than specified batch size
(related to #1708)


ruby_event_store-active_record/spec/event_repository_spec.rb
```diff
@@ -239,7 +238,7 @@ module ActiveRecord
         )
 
         expect {
-          repository.read(specification.in_batches.result).to_a
+          repository.read(specification.in_batches(3).result).to_a
```
pawelpacana, line 241: Seems like you wanted an additional test with non-default batch size

PR #1738, opened 2024-02-07
Remove dependency on base64 gem

Earlopain:
In #1719 `base64`` was added as a dependency to `aggregate_root` to prevent warnings/errors in Ruby 3.3/3.4

However there is only one require and no usages of `base64` so it can just be removed.


pjurewicz:
Hey @Earlopain, good catch! Thank you for your contribution

PR #1742, opened 2024-02-10
Make browser history know about pagination

swistak35:


swistak35:
A few more design (as in: code) improvements could be made:
* Pagination.Specification could have another similar component, but knowing also what resource is being paginated, so that full specification can be passed over to Api, for example
* Pagination buttons could be extracted to separate component with it's own model?

I have a thought that instead of decomposing the model/flags and passing down only subset of fields, it would be better in most cases to pass whole model/flags, and decompose only if we are really on the bottom of displaying something. For example, I needed to pass down much more data into `displayPagination`, because there was assumption, that only `links` are enough, but it isn't. To that point it would still be good to pass everything from current component.

PR #1764, opened 2024-04-30
Remove `arkency-command_bus` dependency

TastyPi:
AFAICT this is not referenced anywhere in the gem, and we (currently) don't plan to use it, so it would be nice to not need to install it.


jandudulski:
This can come with a surprise for people that rely on the gem and do not specify it explicitly in the Gemfile. At least [docs](https://railseventstore.org/docs/v2/command_bus/) should also be updated that the gem has to be added manually.

PR #1789, opened 2024-07-16
Use record `created_at` as `enqueued_at` timestamp when pushing a job to Sidekiq

jakubkosinski:
This allows Sidekiq to calculate the latency including the time a job has spent in outbox


swistak35:
:+1:

Issue #1824, opened 2024-10-16
[deprecation warning]: `ostruct.rb` was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0


francois:
Just following instructions!

> /Users/francois/.rvm/gems/ruby-3.3.5/gems/ruby_event_store-2.15.0/lib/ruby_event_store.rb:6: warning: /Users/francois/.rvm/rubies/ruby-3.3.5/lib/ruby/3.3.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
> You can add ostruct to your Gemfile or gemspec to silence this warning.
> Also please contact the author of ruby_event_store-2.15.0 to request adding ostruct into its gemspec.

OpenStruct is known to not be super performant. I have not investigated how/where it is used. Can OpenStruct be replaced, without going to an outside library?

lukaszreszke:
Hi @francois. Thanks for reporting. I'll check that with the team and let you know.

fidel:
Hey @francois!

Fix for deprecation warning is not yet released, but you can use it by providing `github: "RailsEventStore/rails_event_store"` option in `Gemfile` along with this specific `ref:` https://github.com/RailsEventStore/rails_event_store/commit/291e0b0f65290092ca02b6827aa34861e85f78dd 

Happy New Year!

PR #1827, opened 2024-10-31
Remove unnecessary import of ostruct

leoarnold:
This line is not needed and triggers the error message

```
warning: ostruct was loaded from the standard library,
but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
```


fidel:
Hey @leoarnold!

Thanks for your contribution. Fix was a bit more sophisticated, you can find details in https://github.com/RailsEventStore/rails_event_store/issues/1824

Regards,
Szymon

PR #1854, opened 2025-03-28
feature: batch mapper

Bertg:
This PR is a follow-up of the conversation here: https://github.com/orgs/RailsEventStore/discussions/1847

A small chance is made to how event store invokes the mapper for reading events. By default it will still handle each event one by one, but it does that now inside the mapper, instead of the `Specification::Reader`. This provides a hook for developers to do some batch work on their mappers (eg: preloading encryption keys).

Thins to consider, improve:

* Mutation tests are failing, but they are failing on code not directly touched by this PR
* Naming of the `map_records_to_events` method
* The warning in `SpecificationReader` shows up only once in an application, but during tests it shows up a lot.


mpraglowski:
Moved to #1893

mpraglowski:
@Bertg thanks for your contribution - we will use it but the way of choosing batch strategy will be different. I'm working on it. It will be released as an experimental feature  https://railseventstore.org/docs/master/contributing/maintenance_policy#experimental-features

PR #1855, opened 2025-04-04
Event type resolver in message broker

mpraglowski:


mostlyobvious:
Broker should not be aware of event and this event type resolver. It should receive topic (formerly event_type).

PR #1859, opened 2025-04-26
Closes #1681. JSON client in serialization docs

### LINKED ISSUE #1681: Documentation on Postgres jsonb serialization is not consistent
My team recently changed our RES config to use PG jsonb serialization. We followed the docs here: https://railseventstore.org/docs/v2/mapping_serialization/#configuring-for-postgres-json-b-data-type

Only later on did we discover JSONClient that is mentioned at the bottom of this page: https://railseventstore.org/docs/v2/install/

The mapping serialization page should probably mention the JSONClient that solves the issue in blue text at the bottom of that page.

mlh758:
Saw the open issue, thought I would offer a PR for it since I went through a similar process integrating the library.

Caught an unrelated spelling error and reworded a sentence I thought was little odd but would revert that if you preferred the original phrasing.

### DIFF
--- a/railseventstore.org/docs/advanced-topics/event-serialization-formats.md
+++ b/railseventstore.org/docs/advanced-topics/event-serialization-formats.md
@@ -63,7 +63,13 @@ To generate migration containing event table schemas run
 $ rails generate rails_event_store_active_record:migration --data-type=jsonb
 ```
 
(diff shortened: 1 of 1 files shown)


mostlyobvious:
Thank you 🙏

PR #1875, opened 2025-05-15
Duplicate PR workflow trigger

mostlyobvious:
When opening PRs within this repository checks are duplicated — they run
both on push (to the branch) and as a result of opening a PR.


mostlyobvious:
Wrong solution for duplicate builds, #1878 and #1879 are the ones.

PR #1892, opened 2025-05-21
original intent

mostlyobvious:
- **Refactor**
  

- **Subscriptions don't transform class to event_type**
  Their scope is limited to event_type. Let facade od that heavy lifiting.


mpraglowski:
Already in codebase

PR #1909, opened 2025-07-09
Update browser time to follow ISO 8601 standard

rickychilcott:
I'm finding it difficult to read little-endian datetime formats in the events browser. ISO 8601 is the most universally recognized standard that is locale-agnostic.

This PR updates the formatting logic for Rails Events Browser to use ISO 8601.


mostlyobvious:
Looks pretty good!

zrzut 2025-07-9 o 16 38 17

PR #1914, opened 2025-09-10
Avoid infinite loop when upcast is a no-op

mpraglowski:
Equality check makes upcast to process again the record
even if the upcast itself made no changes (no-op).


fidel:
Closing, described here:
https://railseventstore.org/docs/master/advanced-topics/migrating-existing-events

fidel:
Solved in https://github.com/RailsEventStore/rails_event_store/commit/2a3b93d9fd1da0d582d7c0b2ce97ef076e37aaa1

PR #1920, opened 2025-11-07
Browser: Add view for list of event types

swistak35:
The PR adds the view for the list of event types available in the instance.

Inspired by https://github.com/orgs/RailsEventStore/discussions/1139

## Screenshots

Screenshot, but one can also just run browser's devserver and go to http://localhost:9393/types

image

When functionality not enabled:

image

## Caveats / challenges

1. There are few different ways that this could be approached. I went with the one which assumes that the events we want to display, are the ones defined by the `RubyEventStore::Event`. This allows that in the future additional information could be shared on this view (schema, event version, encryption properties, ...), but is a "convention over configuration" style. We have never declared that inheriting from the `RubyEventStore::Event` is mandatory, however I treat it as "if you do it, you get the most additional features in the ecosystem". Similarly, you can use DB connection through `ActiveRecord::Base` and do INSERT/UPDATE/whatever without an actual `ActiveRecord::Base` class, yet by doing that you lose some of the additional functionalities which rails brings to the table.

Another approach would be to fetch only the ones from the database. However going this way, I feel that actual possibilities are severely limited, since the database does not hold _that much_ interesting information as the event classes and event store instance. For example, extension of current view could be displaying the subscriptions of the instance, etc., relying purely on data from the database seems unnecessarily limiting providing what experience we could bring to the Browser. Also, we go here into the area of DB performance problems, whether the queries on the indexing fetching the event types are fast enough etc.

Another approach could be a hybrid one, fetching the data with both sources. Slightly more complex with more room for error. However it could nicely cover the cases where event versioning is used, or there are events which were in the codebase, but are not there anymore. It would be a shame to not display them in this case.

2. `experimental_event_types_query`, default implementation is provided, but with this keyword -- it needs to be enabled explicitly, something which would allow us to break compatibility of this new view in minor versions. Similarly as it was done when introducing `StepByStepFormatter` for RSpec.

3. The UI can get opinions, I have not thought thoroughily on this, but until there is a bit more data than just a link, it's hard to think about providing better UX. In other words, I'm definitely open for suggestions, but I'd leave it for later.

4. Access. Currently one has to know a link :D Short approach would be to provide a link in the footer for now if this experimental feature is enabled, but I'd rather tend to provide a horizontal navbar on the top, below the current navbar. WDYT?

5. (minor), currently the initialization looks like this:

Because (I think?) we cannot be certain whether RES instance is initialized during mounting the browser. So either we do this, or we pass the `` and take `event_store` from the `event_store_locator`, but then lock the API of the constructor in that QueryObject... I'm not certain what's better, I prefer passing built instances.

---

WDYT?


ruby_event_store-browser/lib/ruby_event_store/browser/event_types_querying/default_query.rb
```diff
@@ -0,0 +1,25 @@
...
+    module EventTypesQuerying
+      class DefaultQuery
+        def initialize(event_store); end
+
+        def call
+          all_event_subclasses(RubyEventStore::Event)
+            .select { |klass| !klass.name.nil? }
+            .sort_by(&:name)
+            .uniq(&:name)
+            .map { |klass| EventType.new(event_type: klass.name, stream_name: "$by_type_#{klass.name}") }
```
swistak35: Also this stream name, should not be hardcoded

PR #1937, opened 2026-05-15
Remove deprecations

tomaszpatrzek:
Remove deprecated code

  This branch removes everything that was warned about in the deprecation-warnings branch. Each commit
  corresponds to one deprecation warning and references the source commit that introduced it.

  What's removed

  Specification
  - Specification#in_batches_of — use in_batches
  - Specification#of_types — use of_type

  RailsEventStore
  - RailsEventStore::* constant aliases for RubyEventStore::* classes
  - *.rails_event_store instrumentation event names — only *.ruby_event_store names remain

  Projection
  - Projection.from_stream, from_all_streams, init, when, run class-level API
  - Multi-scope Projection#call(*scopes) — single scope only
  - Projection.new is now private — use Projection.init

  AggregateRoot
  - apply_* method naming convention — use explicit apply(Event) calls
  - AggregateRoot::Configuration and default_event_store
  - with_strategy and with_default_apply_strategy includes — use AggregateRoot.with(strategy:)

  Dispatchers
  - ImmediateAsyncDispatcher — use ImmediateDispatcher
  - AfterCommitAsyncDispatcher — use AfterCommitDispatcher
  - Dispatcher alias — use SyncScheduler
  - Class-as-subscriber support in SyncScheduler — subscribers must respond to #call

  Mappers
  - EventClassRemapper and events_class_remapping: — use upcasting
  - NullMapper — use Default
  - JSONMapper — use Default
  - serialize/deserialize mapper instrumentation event names — only event_to_record/record_to_event remain

  Repository
  - events: and messages: payload keys in InstrumentedRepository — use records:
  - ensure_supported_any_usage: option in InMemoryRepository (now always enforced)

  Client
  - Passing nil to publish, append, or link now raises ArgumentError immediately

  Gem
  - rails_event_store_active_record shim no longer emits a runtime warn or post-install message


cbillen:
Does this also drop the `require "rails_event_store_active_record"` in `rails_event_store.rb:3`

fidel:
> Does this also drop the `require "rails_event_store_active_record"` in `rails_event_store.rb:3`

@cbillen sorry for the hassle with superfluous deprecation messages. This shouldn't happen. I just finished working on a fix and refactoring the way we handle deprecations at all. People who use this shim indirectly shouldn't be penalized and I was able to achieve that (already on the master branch).

The 2.19.2 is expected to resolve this soon. Following branch needs to be readjusted though.

fidel:
No longer relevant

PR #1939, opened 2026-06-16
Add Gigs to the homepage companies list

brennovich:
Gigs (gigs.com) uses Rails Event Store in production. This adds them to the homepage companies section.
Screenshot 2026-06-16 at 15 53 13


andrzejkrzywda:
Thanks @brennovich!

PR #1946, opened 2026-07-16
Fix event type resolver usage mismatch

mpraglowski:


fidel:
Two things before merge:
1. Publish now routes by `event_type_resolver.call(event.class)` and stops honoring `metadata[:event_type]` — that's a behavior change, not just a fix (tried to explain that in RN AgggregateRoot section)
2. `OnDSL#on` registers under `event_klass.to_s`, but `DefaultApplyStrategy` looks up by `event.event_type`. Equal in practice, but deliberate, or should the lookup use `event.class.to_s` for symmetry?

PR #1947, opened 2026-07-16
Add Docker-based local dev environment (no Ruby install required)

Kotmin:
## Why

Contributing to this repo currently requires a local Ruby toolchain matching the CI matrix (correct Ruby version, native deps for pg/mysql/sqlite, etc.). That's friction for anyone who doesn't already have a Ruby environment set up, or who wants to try a gem in isolation without touching their system Ruby.

## What

Adds an opt-in Docker/Compose based dev environment under `support/docker/`, wired into each gem's `Makefile` alongside the existing `make` targets rather than replacing them:

- `support/docker/Dockerfile` — multi-stage build (`test` / `mutate` / `app` / `console` targets, selected via `GEM_DIR` build arg)
- `support/docker/docker-compose.yml` / `docker-compose.dev.yml` / `docker-compose.test.yml` — app-only, live-reload overlay, and Postgres/MySQL test stack respectively
- New `make docker-test`, `make docker-mutate`, `make docker-dev`, `make docker-dev-watch`, `make docker-console` targets (currently on `ruby_event_store-active_record` and `ruby_event_store-browser`, the two gems with external DB/asset build requirements)
- A CI smoke test (`docker_dev_setup_test.yml`) that builds and boots the dev environment
- `support/ci/generate`'s `support_paths` updated to exclude `support/docker/**`, so changes here don't trigger the full ~57-workflow gem test/mutate/coverage matrix
- A docs page under `railseventstore.org/docs/contributing/`

Scope is intentionally minimal — this only adds a new opt-in path; it doesn't touch existing `make test`/`make mutate` behavior or CI for any gem. Opening as a draft since I'd like feedback on whether this is a direction the project wants before polishing further (e.g. whether more gems should get `docker-test` targets, or whether the smoke-test workflow should be folded into `support/ci/generate` itself).

## Test plan

- [x] `make docker-test` in `ruby_event_store-active_record` against Dockerized Postgres/MySQL — 414 examples, 0 failures, 7 pending
- [x] `make docker-test` in `ruby_event_store-browser` — 87 examples, 0 failures, 4 pending
- [x] `make docker-dev` boots the browser app; root, favicon, and fingerprinted CSS/JS/Stimulus assets all resolve correctly
- [x] Rebased cleanly onto current upstream `master`


fidel:
Thanks for your contribution! 

However, I had a different idea for that, especially due to fact that we already had some building blocks in place.

See: https://github.com/RailsEventStore/rails_event_store/commit/1cfc917f02c978c193a346cf94fd841d786fa79e

PR #1949, opened 2026-07-16
swimline view

pjurewicz:
## Summary

- Adds a "compare streams" view to `ruby_event_store-browser`: from a stream's page you can add other streams and see their events merged into one newest-first timeline table, one column per stream, with infinite scroll. An event linked into several compared streams renders once, as a link in each matching column.
- The data layer lives entirely in the browser gem and composes each page from existing single-stream reads — no new repository/core API. Each compared stream is read in `as_at` order and the per-stream pages are k-way merged by event time, so the merge inputs are sorted by the merge key and the result is correct also for linked events and custom timestamps.
- The timeline can be sorted along either axis of the bi-temporal model: created-at (`as_at`, default) or valid-at (`as_of`), toggled in the header. The whole pipeline — per-stream reads, merge key, cursor comparisons, group draining and the displayed time column — follows the chosen axis.
- Pagination uses a plain timestamp cursor. Pages never split a group of events sharing one timestamp: a page runs to the end of its last timestamp (fetching the rest of the group with `between(t..t)` when the per-stream page got cut exactly there), so the next page asks for strictly older events and nothing is dropped or repeated. Pages can therefore be a little longer than the nominal count — and a mass of identical timestamps (e.g. a backfill with second precision) renders as one long page.
- Fixes the in-memory repository applying `limit` before the time filters, unlike the SQL adapters where `WHERE` precedes `LIMIT` — `limit` + `older_than` is exactly the combination this pagination relies on. A shared lint example pins the ordering for all adapters.

## Test plan

- [x] `ruby_event_store` suite (incl. the new limit-vs-time-filter lint example) — green
- [x] `ruby_event_store-active_record`, `contrib/ruby_event_store-sequel`, `contrib/ruby_event_store-rom` suites — green (no adapter changes; lint example passes everywhere)
- [x] `ruby_event_store-browser` suite incl. compare view, JSON endpoint, UI/Capybara — green
- [x] Mutation testing at 100% for the browser gem diff and the in-memory fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pjurewicz:
it's already there

PR #1955, opened 2026-07-17
Add ruby_event_store-browser-swimlane extension gem

pjurewicz:
## Summary

A browser extension gem (`contrib/ruby_event_store-browser-swimlane`) comparing several streams side by side: their events merged into one newest-first timeline table, one column per stream, with infinite scroll and a created-at/valid-at sort toggle. An event linked into several compared streams renders once, as a link in each matching column.

- **Data layer**: each page is composed from existing single-stream reads — one time-ordered (`as_at`/`as_of`) read per compared stream, k-way merged by event time. Count is the read depth per stream, not the page size: a page emits everything above the *completeness horizon* (the highest chunk end among the streams whose read was cut), so nothing provably complete is thrown away and a page carries between `count` and `count × streams` events. The shared cursor is that horizon timestamp; a chunk cut exactly at it has the rest of that timestamp's group drained with `between(t..t)`, so the next page asks for strictly older events and nothing is dropped or repeated. A mass of identical timestamps renders as one long page.
- **Built entirely on the extension API**: `/compare` and `/compare/more` via `register_routes` (json + `render_partial`), an entry link on every stream page via `stream_links`, its own Stimulus controllers via `scripts` (a separate Stimulus application importing the vendored stimulus module; timezone display synced with the core selector through the shared localStorage key). No core changes.
- CI wired via `support/ci/generate` (test / mutate / coverage workflows).

Relies on #1954 (in-memory repository: limit after time filters) and #1950 (Sequel: time sort on backward reads) — both already merged.

Supersedes the in-repo approach of #1949.

## Test plan

- [x] 36 specs: reader unit coverage (merge, cursors, timestamp groups, linked events, both sort axes) and web coverage through `Browser::App.for(extensions:)` behind `Rack::Lint`
- [x] Mutation testing — 100% (428 mutations)

### DIFF
--- a/.github/workflows/ruby_event_store-browser-swimlane_coverage.yml
+++ b/.github/workflows/ruby_event_store-browser-swimlane_coverage.yml
@@ -1,0 +1,54 @@
+name: ruby_event_store-browser-swimlane_coverage
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types:
+    - script
+  push:
+    branches:
+    - master
+    paths:
+    - contrib/ruby_event_store-browser-swimlane/Gemfile.lock
+    - ".github/workflows/ruby_event_store-browser-swimlane_coverage.yml"
+    - support/**
+    - "!support/bundler/**"
+    - "!support/ci/**"
+  pull_request:
+    paths:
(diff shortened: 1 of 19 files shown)


pjurewicz:
it's already there

PR #1958, opened 2026-07-24
poc/configurable clients

fidel:
- **Extract configurable seams from RailsEventStore::Client**
- **Add transformations seam so subclasses compose the JSON pipeline**
- **Spike: package client variants as composable modules**
- **Spike: add Encoding::Base64 module closing encrypted JSON end-to-end**
- **Tighten Base64 reads: fetch where key is invariant, [] where optional**


mokevnin:
A concrete use case for this seam, in case it helps justify landing it.

We store `T::Enum` (Sorbet) values inside event data. `PreserveTypes` writes them as a bare class name and reads them back through pass-through, so the event is effectively write-only — and the same value crossing ActiveJob comes back as a *new* object, which `T::Enum` doesn't compare equal, so `category == Ecommerce::ProductCategoryType::Program` was silently false in production. The fix is one `register` call per enum class (the class has to be closed over: the deserializer only receives the value, so a single registration for the base type can't work).

But there is nowhere to put that call. `JSONClient`'s pipeline is a default argument value, `Client` exposes no reader for it, and `PreserveTypes#registry` is private. So adding one type means passing `mapper:` and re-declaring all seven built-in registrations (`Symbol`, `Time`, `ActiveSupport::TimeWithZone`, `Date`, `DateTime`, `BigDecimal`, `OpenStruct`) in our own code. That copy is now pinned by a test on our side, but the test cannot notice the gem changing a format or adding a type — a new built-in would silently fall through to pass-through, which is exactly the bug we were fixing.

With the `preserve_types` seam from this PR the whole thing collapses to:

```ruby
class EventStoreClient < RailsEventStore::JSONClient
  ENUMS = [Ecommerce::ProductCategoryType, ...].freeze

  private

  def preserve_types
    ENUMS.reduce(super) do |preserve_types, enum_class|
      preserve_types.register(
        enum_class,
        serializer: ->(v) { v.serialize.to_s },
        deserializer: ->(v) { enum_class.from_serialized(v) },
      )
    end
  end
end
```

Prior art that we're not the only ones doing this: in #1334 the suggested workaround was literally `preserve_types_as_seen_in_json_client.register(OpenStruct, ...)` — "as seen in JSONClient" is the copy this seam removes.

Question rather than a request: could the `transformations` / `preserve_types` seam land on its own, separately from the `Encryption` / `Encoding::Base64` / module-composition part? The seam alone is additive and behaviour-preserving, while the module composition is the part that needs a design decision — splitting them would let the smaller half through without committing to the larger one. Happy to help with either if that's useful.

PR #1960, opened 2026-08-24
Reset column information after including SkipJsonSerialization

nlsrchtr:
`SkipJsonSerialization` overrides `hook_attribute_type`, which ActiveRecord consults while resolving a model's attribute types and then memoises. Models are built lazily, so anything that resolves them before the repository is first used leaves the include without effect and the payload is serialized twice: the JSON column ends up holding a JSON *string*, `data ->> 'key'` returns NULL, and GIN indexes on metadata match nothing.

`rails/test_help` does exactly that — it calls `load_schema` on every ActiveRecord descendant when the host application eager loads, which Rails generates for the test environment as `config.eager_load = ENV["CI"].present?`. An application that builds the repository in an initializer and first uses it inside a test therefore writes double-encoded events.

The fix is one line: reset column information right after the include.

## It is not Postgres-only

The include sits on the path shared by every adapter, and I could reproduce it on **MySQL** as well as PostgreSQL. The added spec therefore covers postgres `json` and `jsonb` and mysql `json`, with adapter-aware SQL — MySQL addresses JSON through a path expression and has no `::` cast.

Verified red/green on each, using the repository's own `Gemfile.database`, matching the CI matrix:

| variant | with the fix | without |
|---|---|---|
| sqlite3 + binary | 305 / 0 | — |
| postgres 14 + binary | 418 / 0 | — |
| postgres 14 + jsonb | 420 / 0 | **420 / 1** |
| postgres 18 + binary | 418 / 0 | — |
| postgres 18 + json | 420 / 0 | **420 / 1** |
| postgres 18 + jsonb | 420 / 0 | **420 / 1** |
| mysql 8.4 + binary | 305 / 0 | 305 / 0 |
| mysql 8.4 + json | 306 / 0 | **306 / 1** |
| mysql 9.7 + binary | 305 / 0 | — |
| mysql 9.7 + json | 306 / 0 | **306 / 1** |

Two things worth reading out of that table. Without the fix there is **exactly one** failure on every JSON variant, and it is the new spec — the other ~420 examples stay green, which is why this could sit unnoticed. And `mysql + binary` passes with and without, so the change is inert where there is no JSON column.

## In production

Running on a staging deployment for two days, on postgres with `jsonb`:

- 261,645 events total, `jsonb_typeof` returns a single group: `object` / `object`
- 3,939 of those written since the deploy, all objects
- 243,570 rows matched by `metadata -> 'site_id' IS NOT NULL`, so the GIN path resolves

## About the spec

It runs in a subprocess. `skip_json_serialization_spec.rb` cannot observe this, because `run_lifecycle` resets column information before each example and hides the memoised attribute types the bug is about. This is the same approach `spec/lazy_load_spec.rb` already takes.

I also passed `stderr: $stderr` through `run_in_subprocess` and had the failing branch print what the column actually holds. Without it the failure a maintainer sees is `expected: 0, got: 1`, which says nothing:

```
data ->> '$.foo' returned nil; the column holds "\"{\\\"foo\\\":\\\"bar\\\"}\""
```

Happy to split the diagnostics out or adjust the spec's shape if you would prefer it differently.


pjurewicz:
@nlsrchtr thanks for the PR! We will look into it very soon

nlsrchtr:
Hi @pjurewicz,

the code and PR is obviously not my brain-child, but AI powered work. The core problem stays the same and I hope this PR helps to enhance the CI suite to cover such hidden problems, which might end up in plain text JSON in JSON columns or related issues.

Let me know what you think.

Regards,

andrzejkrzywda:
Hi @nlsrchtr,

No worries about the AI work here. What's important is that you spotted an important regression here and fixed it. Thank you!

BTW, could you share the specifics of your app when it comes to exposing this bug? It seems we have missed it in our test setup.

pjurewicz:
@nlsrchtr, I'm happy to announce that your change was released in rails_event_store 3.1.0, 3.0.1, and 2.19.3!
Thanks again for your contribution.
Would you be up for a short call (around 10 minutes) to chat about your experience with RES - what works well, what could be better, and what you'd like to see next?

nlsrchtr:
> BTW, could you share the specifics of your app when it comes to exposing this bug? It seems we have missed it in our test setup.

Hi @andrzejkrzywda,

sure I can add some more context on how we stubled upon this:

We were raising the rails_event_store family from 2.17 to 3.0 as part of a Rails 8 upgrade. The suite stayed green through it, and nothing raised.

What exposed it was writing a test on purpose: after a major bump we tend to add a few guards that assert how something works rather than that it works. One of them asserted the stored representation of `data/metadata — jsonb_typeof` is `object, data ->> 'key'` comes back, `Portal::Event#data` is a `Hash`. That is what failed.

The defect is not reliably observable from inside a suite run. In a full run something resets the model's memoized attribute types and it disappears — we measured it both ways. Asserting the stored format from an ordinary test only fails when that file runs alone or with a small subset. We ended up booting a fresh process and asserting what Postgres actually stored, which is the same shape as your own `spec/lazy_load_spec.rb`. That is probably why this could sit in a lot of green suites.

Production was never affected — it does not load rails/test_help, and we confirmed the stored data is clean.

nlsrchtr:
> @nlsrchtr, I'm happy to announce that your change was released in rails_event_store 3.1.0, 3.0.1, and 2.19.3! Thanks again for your contribution. Would you be up for a short call (around 10 minutes) to chat about your experience with RES - what works well, what could be better, and what you'd like to see next?

Hi @pjurewicz,

I'm happy, that I was able to support this very useful project. Surely, we can have a call, but honestly I don't know if my feedback is of any help for you - since I think we are using RES in a pretty normal way. If you stil think it might be worth your time, feel free to contact me directly.

Regards!

nlsrchtr:
> > @nlsrchtr, I'm happy to announce that your change was released in rails_event_store 3.1.0, 3.0.1, and 2.19.3! Thanks again for your contribution. Would you be up for a short call (around 10 minutes) to chat about your experience with RES - what works well, what could be better, and what you'd like to see next?
> 
> Hi @pjurewicz,
> 
> I'm happy, that I was able to support this very useful project. Surely, we can have a call, but honestly I don't know if my feedback is of any help for you - since I think we are using RES in a pretty normal way. If you stil think it might be worth your time, feel free to contact me directly.
> 
> Regards!

Hi @pjurewicz,

the more I think about it, the more topics come to my mind f.e. GDPR and multi tenancy.

Regards,

pjurewicz:
@nlsrchtr, great! Please set up a meeting at any time that works for you:
https://scheduler.zoom.us/piotr-jurewicz/30-mins-with-piotr

PR #1962, opened 2026-08-25
Feature/async bulk dispatcher and scheduler

jdlubrano:
For applications where actions may publish many events in one transaction or there are many subscribers to a particular type of event, you can end up fanning out the events to many async handlers.  If you use a relational database as your ActiveJob backend, those database calls that are quick individually can grow into a performance bottleneck.

Rails 7.1 added the `perform_all_later` API to `ActiveJob` to optimize calls to ActiveJob's backend.  Instead of making one call per job, `perform_all_later` enqueues many jobs in one call.  Under the hood `perform_all_later` leverages Sidekiq's bulk push API or `insert_all` to backends like Solid Queue or GoodJob.  This PR introduces a `RailsEventStore::AfterCommitBulkDispatcher` and `RailsEventStore::ActiveJobBulkScheduler` so that RES consumers may opt in to the same kinds of optimizations for their async handlers.

To opt in, users must configure the dispatcher and scheduler in their application (no behavior changes by default for applications using RES):

```ruby
event_store = RailsEventStore::Client.new(
  message_broker: RubyEventStore::Broker.new(
    dispatcher: RubyEventStore::ComposedDispatcher.new(
      RailsEventStore::AfterCommitBulkDispatcher.new(
        scheduler: RailsEventStore::ActiveJobBulkScheduler.new(serializer: RubyEventStore::Serializers::YAML)
      ),
      RubyEventStore::SyncScheduler.new
    )
  )
)
```

Since the dispatcher is opt-in, I chose not to raise errors (or otherwise validate) that `perform_all_later` is available in the consumer's application.  I can certainly add that sort of thing if the maintainers would like to suggest an approach for doing that 🤔 . 

## Open Questions

* It's entirely possible that publishing many events within a transaction is considered a smell for RES implementations.  Therefore, this optimization may be discouraged conceptually in which case I can close the PR 😃.


rails_event_store/lib/rails_event_store/active_job_bulk_scheduler.rb
```diff
@@ -0,0 +1,24 @@
...
+module RailsEventStore
+  class ActiveJobBulkScheduler < ActiveJobScheduler
+    def bulk_call(subscribers_and_records)
+      jobs = []
+
+      subscribers_and_records.each do |subscriber, record|
+        payload = record.serialize(serializer).to_h.transform_keys(&:to_s)
+
+        if subscriber.instance_of?(ActiveJob::ConfiguredJob)
+          subscriber.perform_later(payload)
```
jdlubrano, line 15: `ConfiguredJob` does not appear to have a public `new` method to use with bulk enqueuing.  Usually configured jobs are bulk enqueued with

```ruby
job = Job.new(...).set(...)
ActiveJob.perform_all_later([job])
```

The `ConfiguredJob` that is provided as a handler, does not know it's inputs, so it seems like we have to enqueue `ConfiguredJobs` one-by-one.

mpraglowski:
Hi Joel,
Thanks for this PR. We've already discussed something similar recently, and this looks like a good fit for what we've been talking about. Give us some time to review it :)
BTW, I'm curious what your use cases are that require a bulk dispatcher.

jdlubrano:
Hi @mpraglowski 👋 .

Thanks so much for giving this consideration.

Regarding our use case, our application allows users to create and edit purchase orders.  Purchase orders have many line items (with attributes like quantity and price).  Our UI for editing the line items is an editable grid so that the user can add or edit many lines at one time.  To "commit" their changes, the user clicks a save button, which saves the entire purchase order and saves each new or edited line.  When lines get committed, we publish events and there are downstream changes from publishing those events handled by subscribers.  We update budgets, we may notify admins for approvals, we also maintain a sort of ledger of the line item changes themselves.  If a user has updated a lot of lines in their purchase order, we see the fanout from enqueuing those subscribers adding up in terms of latency.  We use GoodJob for our ActiveJobs, so each INSERT to the GoodJob DB to enqueue an async subscriber, while only a few milliseconds, can add up for a purchase order with 100+ lines.  We wrapped the interaction in `GoodJob`'s bulk enqueue functionality, which helped quite a bit, so I figured I would try to integrate the same idea generically in Rails Event Store itself 😃 .

mpraglowski:
338593a79073888ac8921b577fe4b3af42479058 is unrelated - please remove

mpraglowski:
Hi @jdlubrano 

I've decided to propose my own PR heavily inspired by your code :)
There are a few problems with this PR that I've tried to address in my PR #1966:

* registry in dispatcher: could be shared between different instances, also could be not released when transaction won't be committed or rolled back by some reason (i.e. dropped connection)
* in my PR I've decided not to change (almost) the dispatcher - all. "plumbing" is in schedulers
* there are 2 different (payloads) after commit schedulers - would be nice to support both of them
* as Rails 7.1 has reached its EOL already I've decided to use `Transaction#after_commit` callback introduced by Rails 7.2 - that simplifies things quite a lot :) But at the same time I still do not force other users who does not use bulk schedulers to upgrade to Rails 7.2 (at least not yet) 

Let me know it this fits your solution (main requirement is Rails 7.2) and if you have any remarks to my PR #1966

jdlubrano:
@mpraglowski #1966 solves my problem and looks great.  I'll close this PR in favor of it 🎉 .

mpraglowski:
Thx @jdlubrano for inspirations. I think the #1966 will be out as RES 3.2 release ... no date yet.
If you have some other RES related topics I would be happy to discuss it - just schedule a meeting here https://scheduler.zoom.us/mpraglowski

PR #1967, opened 2026-08-31
Let AfterCommitDispatcher be told which class owns the transaction

mpraglowski:
`AfterCommitDispatcher` defers scheduling async handlers until the surrounding
transaction commits, and drops them when it rolls back. It looks for that
transaction on `ActiveRecord::Base`. When the application opens its unit of work
on a class with a connection pool of its own, `ActiveRecord::Base` reports no
open transaction, `joinable?` is false, and the dispatcher quietly degrades to
scheduling immediately. Handlers run before the records they describe are
committed, and a rollback cannot take back a job already enqueued, with nothing
in the logs to say so.

The trigger is `connects_to`, not where the events are stored. An abstract class
calling `connects_to` gets its own pool even when it names the very same
database, unless it is the `primary_abstract_class` (then Active Record registers
the pool under `ActiveRecord::Base` and the default already works). An abstract
class without `connects_to` resolves to `ActiveRecord::Base`'s pool, so a default
Rails application needs nothing.

`AfterCommitDispatcher.new` now takes an optional `transaction_owner:`, defaulting
to `ActiveRecord::Base`. It names the class whose transaction wraps the unit of
work:

```ruby
RailsEventStore::AfterCommitDispatcher.new(
  scheduler: RailsEventStore::ActiveJobScheduler.new(serializer: RubyEventStore::Serializers::YAML),
  transaction_owner: ApplicationRecord
)
```

It is unrelated to the repository's `model_factory`. `Client#publish` appends
first and dispatches after, and the repository commits its own transaction on
the event store connection inside `append_to_stream`, so by the time the
dispatcher runs there is nothing open there to join. Pointing `transaction_owner`
at the event store's abstract class reproduces the failure rather than fixing it.

Also in this PR:

- docs: the repository side of keeping events on another database
  (`WithAbstractBaseClass`), including the warning that a second database gives
  up atomicity: a business rollback after `publish` takes the records and leaves
  the events behind
- docs: a section on `connects_to` and `transaction_owner` in Subscribe
- specs: `connects_to` hiding a transaction from `ActiveRecord::Base` on one
  database, and an integration spec publishing through `Client` across two
  connections
- `Migrator#run_migration` accepts `connection:` and migrates through
  `exec_migration`


fidel:
The premies in the description doesn't hold. I'm also afraid that the configuration recommended in the docs can make things even worse. Pushed a follow-up which keeps the option but changes what is it for.

The two transaction never overlap. `Client#publish` appends first and dispatches after:

```ruby
append_records_to_stream(records, stream_name: stream_name, expected_version: expected_version)
enriched_events.zip(records) do |event, record|
  @broker.call(topic || event.event_type, event, record)
end
```

`EventRepository#start_transaction opens on `event_klass` inside `append_to_stream`, so by the time `AfterCommitDispatcher#run` asks for `current_transaction` that transaction has already committed. Nothing is open on the ES connection to join. Dispatcher rather joins application's unit of work, the ten around saving records and publishing.

With the ES on a send db and the business transaction on `AR::Base`, that means master was already correct and the documented `model: EventStoreRecord` introduces the failure:

```ruby
AppRecord.transaction do
  AppRecord.lease_connection.execute("INSERT INTO orders (number) VALUES ('42')")
  event_store.publish(MultiDbOrderPlaced.new(data: { order_id: 42 }))
  raise ::ActiveRecord::Rollback
end

expect(MultiDbAsyncHandler.queued).not_to be_nil                                         # job enqueued
expect(AppRecord.lease_connection.select_value("SELECT COUNT(*) FROM orders")).to eq(0)  # order gone
expect(event_store.read.of_type([MultiDbOrderPlaced]).count).to eq(1)                    # event committed
```

Handler scheduled for an order that doesn't exist. That is the failure this dispatcher meant to prevent.

The real trigger is `connects_to`, not the second DB. I've added specs for this scenario:

- an abstract class without `connects_to` has no connection of its own and resolved `AR::Base` pool, so `ApplicationRecord.transaction` is the same txn object and the default already joins it
- a class calling `connects_to` gets its own pool even when the `db_config` names the very same DB and the txn opened there is invisible to `AR::Base`, which reports `NullTransaction`

A single db app with read replica on `ApplicationRecord`, exactly as the Rails multiple-db guide shows, loses the guarantee. Where the events live has nothing to do with it.

What changed:
- renamed `model:` to `transaction_owner:` since the class to pass is the one whose transaction wraps the unit of work, not the one holding events
- docs now point at `connects_to` and `ApplicationRecord`, the repo page no longer presents the dispatcher as the other half of its own setup
- removed specs that opened the transaction by hand on the ES connection, a situation that won't happen
- added an integration spec publishing through `Client` across two connections and spec pining what `connects_to` does to a txn
- the repo page now warns that a second DB gives up atomicity: the repo commits its onward transaction as soon `publish` returns, so a business rollback afterwards takes the records and leaves the events

btw. the title should probably state "Let AfterCommitDispatcher be told which class owns the transaction"

